PR #96, opened 2016-06-21
changed "how to run" section

Scyks:
I figured out that the described process does now work
like described. So finally i found out how it works.


MasseGuillaume:
thanks !

PR #105, opened 2016-06-22
add support for scala 2.10 and 2.12

Scyks:


data/src/main/scala/ch.epfl.scala.index.data/Main.scala
```diff
@@ -25,7 +25,14 @@ object Main {
 
     if(list) {
       val listPomsStep = new ListPoms
-      listPomsStep.run()
+      // TODO: should be located in a config file
+      val versions = List("2.10", "2.11")
```
MasseGuillaume: 2.12 too !

PR #114, opened 2016-06-24
Cleanup

Scyks:


MasseGuillaume:
Make sense. As discussed, the import could be shorter :)

```
package ch.epfl.scala.index.model            
import ch.epfl.scala.index.model.misc.{GeneralReference, ISO_8601_Date, MavenReference}
import ch.epfl.scala.index.model.release._
```

vs

```
package ch.epfl.scala.index.model
import misc.{GeneralReference, ISO_8601_Date, MavenReference}
import release._
```

Issue #112, opened 2016-06-24
Strategy to determine libraries with "Help wanted"


raulraja:
@MasseGuillaume It turns out that all Github Repos come with built in labels for issues. One of the labels is `help wanted`. We could potentially iterate over the project issues and determine if it's a project wanting help by the presence of these tags in their issue list.

Scyks:
This could be a featured tag (#90, #113) then, that you can search for artifacts with `help wanted`. That's a very cool idea.

MasseGuillaume:
This is a good idea! I'm curious to see how often this is used already.

Scyks:
Regarding the github search it's 782 issues for scala:
https://github.com/search?l=scala&q=label%3A%22help+wanted%22&ref=searchresults&type=Issues&utf8=%E2%9C%93

MasseGuillaume:
* [add `List[Issue]` to the model for the github api](https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/github/GithubModel.scala#L1)
* [add `List[Issue]` to the scaladex model](
https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/misc/GithubInfo.scala#L17)
* [modify the function to convert github api model to scaladex model](https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/github/GithubReader.scala#L24)
* [add a step to download github issues](
https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala#L278)
* (convert contributorsWanted from `Boolean` to `Option[Boolean]` where `None` means it was not defined by the user, `Some(x)` it was set by the user)[https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Project.scala#L39]
* [count `help wanted` issues to update the project](https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala#L139)

https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala#L139

MasseGuillaume:
this is now the scope of @MichaelViveros 's GSOC project.

MasseGuillaume:
This is now live! Check it out.

Issue #48, opened 2016-06-08
follow redirects when downloading github readmes


MasseGuillaume:
https://github.com/scalacenter/scaladex2/blob/master/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala#L73

this should be a stream graph with retries

PR #117, opened 2016-06-25
Refactoring downloader and use play-ws

### LINKED ISSUE #115: Re-add and wire up statistics (UI is there, but GitHub HTTP reqs not)
screen shot 2016-06-25 at 11 09 59 am

Scyks:
rebased on cleanup because of model touching. So please merge #114 first.

implemented play-ws api for downloading content
from websites (Github, Maven, Bintray) to avoid the
problems of redirects because play-ws does that by
default.

Also there is now a more abstract way to download
content to avoid repeating code.

@MasseGuillaume What do you think about that?

will also fix #115 and maybe #48 and prepared #112


build.sbt
```diff
@@ -87,7 +87,8 @@ lazy val data = project
       "me.tongfei"              % "progressbar"                       % "0.4.0",
       "org.apache.maven"        % "maven-model-builder"               % "3.3.9",
       "ch.qos.logback"          % "logback-classic"                   % "1.1.7",
-      "org.jsoup"               % "jsoup"                             % "1.9.2"
+      "org.jsoup"               % "jsoup"                             % "1.9.2",
+      "com.typesafe.play"       % "play-ws_2.11"                      % "2.5.4"
```
MasseGuillaume: ``` "com.typesafe.play"       %% "play-ws"                      % "2.5.4"```

PR #121, opened 2016-06-30
Feature/118 refactor akka http

Scyks:
Since Github fetching is already migrated to play-ws, the POM-fetching should be converted to.


data/src/main/scala/ch.epfl.scala.index.data/bintray/ListPomsNEw.scala
```diff
@@ -0,0 +1,125 @@
...
+package bintray
+
+import java.nio.file.Files
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{FileIO, Flow, Keep, Source}
+import akka.util.ByteString
+import com.github.nscala_time.time.Imports._
+import model._
```
MasseGuillaume: I try to order imports by relevance. Like anything from the project at the top, then protocols, transport, utils, etc

data/src/main/scala/ch.epfl.scala.index.data/bintray/ListPomsNEw.scala
```diff
@@ -0,0 +1,125 @@
...
+   * @param scalaVersion the current scala version
+   * @return
+   */
+  def getNumberOfPages(scalaVersion: String): Future[InternalBintrayPagination] = {
+
+    val request = discover(wsClient, PomListDownload(scalaVersion, 0))
+
+    request.head.flatMap { response =>
+
+      response.status match {
```
MasseGuillaume: would use if here

data/src/main/scala/ch.epfl.scala.index.data/bintray/ListPomsNEw.scala
```diff
@@ -0,0 +1,125 @@
...
+   * them over
+   *
+   * @param scalaVersion the current scala version
+   * @return
+   */
+  def getNumberOfPages(scalaVersion: String): Future[InternalBintrayPagination] = {
+
+    val request = discover(wsClient, PomListDownload(scalaVersion, 0))
+
+    request.head.flatMap { response =>
```
MasseGuillaume: head like http verbs ?

Scyks: Yep, it's a HEAD request - we only need headers there

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
```diff
@@ -1,90 +1,130 @@
...
+
+    verifyChecksum(content, sha1)
+  }
+  /**
+   * get a list of bintraySearch object where no sha1 file exists
+   */
+  private val searchesBySha1: Set[BintraySearch] = {
+
+    BintrayMeta.readQueriedPoms(bintrayCheckpoint)
+      .filter(s => !Files.exists(pomPath(s)) || !verifySHA1FileHash(pomPath(s), s.sha1))
```
Scyks: @MasseGuillaume I implemented the sha1 check like described in a TODO on it. During that i found out that the writing of the poms before was not compatible with the orignal file. This means that 100% of the downloaded files does not match against SHA1 hash. We need to think about downloading all again or keeping it.

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
```diff
@@ -1,90 +1,130 @@
...
+   * @param search the bintray search
+   * @param response the download response
+   */
+  private def processPomDownload(search: BintraySearch, response: WSResponse): Unit = {
+
+    if (200 == response.status) {
+
+      val path = pomPath(search)
+
+      if (Files.exists(path)) {
```
Scyks: @MasseGuillaume: Does it make sense to delete an existing file, when i receive a new one with the same hash?
What about hash collision - could that cause into problems here?

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
```diff
@@ -1,90 +1,130 @@
...
-    }.map{
-      case Right((pom, search)) => {
-        progress.step()
-
-        val printer = new scala.xml.PrettyPrinter(80, 2)
-        val pw = new java.io.PrintWriter(pomPath(search).toFile)
-        pw.println(printer.format(scala.xml.XML.loadString(pom)))
-        pw.close
+
+      if (verifyChecksum(response.body, search.sha1)) {
```
Scyks: @MasseGuillaume: Here i check the Checksum - to make sure we have the right content. As mentioned earlier the writing is now native to make sure SHA1 hash will be the same. Do you see problems?

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
MasseGuillaume: did you tried downloading one manually ?

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
MasseGuillaume: We do have some equivalent contents, like two paths giving the same sha, that's allright we can keep one or the other since they are equivalent :-)

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
MasseGuillaume: It's a good idea :)

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
Scyks: So delete old content before should be okay?

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
Scyks: yes i compared the hashes, they are equal now but not to any old downloaded file

Scyks:
@MasseGuillaume fetching parent poms is now also refactored. Now you can review the whole PR. If there is something to do, let me know.

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
MasseGuillaume: so per discussion it's because we pretty print
https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala#L77-L79

data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
MasseGuillaume: per discussion we ignore hash collision

Issue #129, opened 2016-07-01
Who's analytics code is this?


heathermiller:
https://github.com/scalacenter/scaladex/blob/76509b1bada483ea577f930903a58247e2e10447/template/src/main/twirl/ch.epfl.scala.index.views/main.scala.html#L112

heathermiller:
Resolved :)

PR #130, opened 2016-07-01
Updating analytics

heathermiller:
Can someone tell me who's collecting index.scala-lang.org's analytics data?

In any case, it should all go with the rest of the Scala properties (this PR)


heathermiller:
Question raised in #129 still valid, prease.

PR #135, opened 2016-07-02
Index claimed repos

Scyks:
This is a quick fix for indexing claimed repos.

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/cleanup/GithubRepoExtractor.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/cleanup/GithubRepoExtractor.scala
@@ -20,8 +20,8 @@ class GithubRepoExtractor extends DefaultJsonProtocol {
     cleanupIndexBase.resolve(Paths.get("claims.json")).toFile
   )
   private val claims = source.mkString.parseJson.convertTo[Map[String, Option[String]]]
-  private val matchers = claims.toList.sorted.
-    map{case (k, v) => v.map((k, _))}.flatten.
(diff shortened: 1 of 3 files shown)


data/src/main/scala/ch.epfl.scala.index.data/cleanup/GithubRepoExtractor.scala
```diff
@@ -73,7 +73,8 @@ class GithubRepoExtractor extends DefaultJsonProtocol {
         matches(m, s"$groupId $artifactId")
       }.map(_._2).toList
 
-    (fromPoms ++ fromClaims).map{ 
+    /* use claims first - because project indexing uses only the head github repo */
```
Scyks: @MasseGuillaume I analysed the source code a bit to find out what happens with the list of repositories. The Indexing process will only take the *first* github repo and uses that to show related infos on the page.

The github download will take all github repos and fetch the data which actually makes no sense because only the first is relevant for indexing.

I suggest, that we clean up the process for the `GithubRepoExtractor` to only return 1 repo. Also we need to use the claimed repo first and then the defined POM repo. If we do so, we need to cleanup the downloaded github data and delete unused repos.

Would be nice, if we can have a chat about that.

MasseGuillaume: we could have multiple project reference in one project, this would handle old repos.

Scyks: True, but for now it's not yet used. So we stick to the schema?

PR #143, opened 2016-07-03
Adds an organization page, fixes #139

### LINKED ISSUE #139: Organization page?
Maybe we should (eventually) consider creating an organization page which contains all published projects/artifacts under a specific user?

heathermiller:
Adds an organization page to Scaladex. E.g., http://index.scala-lang.org/twitter

Resolves #139


server/src/main/scala/ch.epfl.scala.index.server/Server.scala
```diff
@@ -156,15 +164,15 @@ object Server {
...
-          parameters('artifact, 'version.?){ (artifact, version) =>
-            val rest = version match {
-              case Some(v) if !v.isEmpty => "/" + v
-              case _ => ""
-            }
-            redirect(s"/$owner/$artifact$rest", StatusCodes.PermanentRedirect)
-          }
-        } ~
+        // path(Segment) { owner =>
+        //   parameters('artifact, 'version.?){ (artifact, version) =>
```
heathermiller: `artifact` was a required parameter here. Given the path, it doesn't make much sense like this.

heathermiller:
Final todo on this:

- [ ] Factor out searchresults into results and base searchresults and organizationpage on it.

PR #149, opened 2016-07-04
restore redirect for artifact and release selection

MasseGuillaume:


Scyks:
Now, as @heathermiller reported in #152 the sorting doesn't work any longer.
Maybe we should fix this by adding a JS to the dropdowns.
 https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/headproject.scala.html#L40

heathermiller:
All good, this PR wasn't related to #152 :)

PR #151, opened 2016-07-04
 fix #150 - pagination handling error

### LINKED ISSUE #150: Not all scala versions where downloaded
Somehow not all Scala versions are downloaded. 2.11.8 is missing by the way.

Scyks:
@MasseGuillaume Seems like that there was a fundamental mistake in the refactoring branch. The Pagination on bintray does not work with pages, it works with a start index. This leads then to a big problem in searching artifacts on bintray because only the first page where downloaded plus a few from second page.

We need to revert the `bintray.json` to yesterday, before starting reindexing, and then do the search again + downloading the missing poms.

I'm Sorry

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
@@ -55,7 +55,7 @@ class DownloadPoms  (implicit val system: ActorSystem, implicit val materializer
   private val searchesBySha1: Set[BintraySearch] = {
 
     BintrayMeta.readQueriedPoms(bintrayCheckpoint)
-      .filter(s => !Files.exists(pomPath(s)) || !verifySHA1FileHash(pomPath(s), s.sha1))
+      .filter(s => !Files.exists(pomPath(s)))// || !verifySHA1FileHash(pomPath(s), s.sha1))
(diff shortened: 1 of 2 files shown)


data/src/main/scala/ch.epfl.scala.index.data/bintray/DownloadPoms.scala
```diff
@@ -55,7 +55,7 @@ class DownloadPoms  (implicit val system: ActorSystem, implicit val materializer
   private val searchesBySha1: Set[BintraySearch] = {
 
     BintrayMeta.readQueriedPoms(bintrayCheckpoint)
-      .filter(s => !Files.exists(pomPath(s)) || !verifySHA1FileHash(pomPath(s), s.sha1))
+      .filter(s => !Files.exists(pomPath(s)))// || !verifySHA1FileHash(pomPath(s), s.sha1))
```
Scyks: @MasseGuillaume Doesn't seem to work, every time i get 53323 files to download, after commenting that out just 23.

Scyks:
Already downloaded the new poms:
https://github.com/scalacenter/scaladex-poms/pull/3

MasseGuillaume:
@Scyks what's the state of this PR? Did you fix the issue or it's a WIP ?

Scyks:
@MasseGuillaume https://github.com/scalacenter/scaladex-poms/pull/3 prooves that it works. It's ready to merge.

PR #163, opened 2016-07-04
Fixes #160 - not rendering some pages

### LINKED ISSUE #160: Speed up scala-library pages
Actually the scala-library page will not render at all. https://index.scala-lang.org/scala/scala-library/2.11.7

We could optimize the speed of the pages a bit.

Scyks:
The scala-library page by the way is so slow that it terminates after
20 seconds.

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/Release.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/Release.scala
@@ -95,24 +95,33 @@ case class Release(
   /**
    * ordered scala dependencies - tests last
    */
-  lazy val orderedDependencies = scalaDependencies.sortBy(_.scope.contains("test"))
+  lazy val orderedDependencies = {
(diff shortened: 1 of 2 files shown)


model/src/main/scala/ch.epfl.scala.index.model/Release.scala
```diff
@@ -95,24 +95,33 @@ case class Release(
...
+    val (a, b) = scalaDependencies.sortBy(_.reference.name).partition(_.scope.contains("test"))
+    b.groupBy(b => b).values.flatten.toList ++ a
+  }
 
   /**
    * ordered java dependencies - tests last
+   * - watch out the performance on /scala/scala-library
    */
-  lazy val orderedJavaDependencies = javaDependencies.sortBy(_.scope.contains("test"))
+  lazy val orderedJavaDependencies = javaDependencies.sortBy(_.scope.contains("test")).sortBy(_.reference.name)
```
MasseGuillaume: sort sort ?

Issue #158, opened 2016-07-04
Version selector on artifact page should default to core (etc)


heathermiller:
Right now I don't know how the default artifact gets selected. However, most often it's the least relevant one (e.g., -streams instead of -core). 

If a "-core" artifact exists we should probably default to it being selected. Else, we should default to whatever is closest to the project name (e.g. if the project is "scalaz" and if there's no "scalaz-core" perhaps there's an artifact called just "scalaz")

Even further – when the edit functionality is finished, it would be worthwhile to make it possible for library authors to select the "default" artifact to be shown when their index page is loaded. Just in case there are people out there who name their artifacts in non-standard ways.

PR #165, opened 2016-07-04
fix parts of #158 - select *-core frist

Scyks:
Change default of selecting the default artifact.
- *-core
- organization = artifact
- first entry


heathermiller:
👏 🍻 🎏 🎉

PR #181, opened 2016-07-13
flat project reference and search relevance

MasseGuillaume:
requires reindexing

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/bintray/ListPoms.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/bintray/ListPoms.scala
@@ -33,7 +33,7 @@ class ListPoms(implicit val system: ActorSystem, implicit val materializer: Acto
   /**
    * The url to search at
    */
-  val binTrayUri = "https://bintray.com/api/v1/search/file"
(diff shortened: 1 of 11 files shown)


data/src/main/scala/ch.epfl.scala.index.data/bintray/ListPoms.scala
```diff
@@ -65,17 +64,11 @@ class ListPoms(implicit val system: ActorSystem, implicit val materializer: Acto
...
-            case (Some(totalPages), Some(limit)) => InternalBintrayPagination(totalPages.toInt, limit.toInt)
-            case (Some(totalPages), None) => InternalBintrayPagination(totalPages.toInt, 50)
-            case _ => InternalBintrayPagination(0, 50)
-          }
+        Future.successful{
+          InternalBintrayPagination(response.header("X-RangeLimit-Total").map(_.toInt).getOrElse(0))
         }
       } else {
-
+        println(response)
```
Scyks: I guess it's not necessary?

MasseGuillaume: ops

PR #183, opened 2016-07-13
add "how to publish page" to scaladex

### LINKED ISSUE #180: How to publish page
Would be nice also to have the publishing procedure also on the website. Maybe people searching for that already.

We could use the text from [README](https://github.com/scalacenter/scaladex/pull/176/commits/c746c3124f8a2cf38faf97aed8c2972965c12bc1)

Scyks:
fix #180

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/Server.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/Server.scala
@@ -228,6 +228,13 @@ object Server {
             }
           }
         } ~
+        path("howtopublish") {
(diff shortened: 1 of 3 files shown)


heathermiller:
Stupid question... We're copy/pasting a lot of HTML from other templates. Would it be possible to make a "page" template in case we'd like to eventually have more informational pages other than the how to publish page?

We could parameterize it with page title and body content (as an HTML string)

Is this an unnecessary worry?

MasseGuillaume:
If it's static I prefer we leave it in the github wiki.

Issue #188, opened 2016-07-15
Adopt scalafmt


jvican:
Hello, given the success of the project and that some parts of the codebase are not very well formatted, I would propose to adopt scalafmt. I think it's a very good tool that works reliable, plus it's developed by one of the SC guys.
  
If you agree on this, I'll make a PR with it.

MasseGuillaume:
@jvican go for it, I hope it does a good job with the akka-http routing dsl :P

jvican:
I've run scalafmt, and I'm getting a weird error that apparently happens because of a scala 2.10 compatibility issue (see [this](http://stackoverflow.com/questions/25089852/what-is-the-reason-for-java-lang-nosuchmethoderror-scala-predef-arrowassoc-upo)).
  
I run scalafmt with sbt in the root project. It crashes when formatting the server:
```
[info] Resolving org.scala-lang#scalap;2.11.8 ...
[info] Done updating.
[info] Updating {file:/data/rw/code/scala/scaladex/}server...
[info] Resolving com.typesafe.play#twirl-api_2.11;1.1.1 ...
[info] Formatting 27 Scala sources {file:/data/rw/code/scala/scaladex/}data ...
[info] Resolving org.scala-lang#scalap;2.11.8 ...
[info] Done updating.
[info] Formatting 6 Scala sources {file:/data/rw/code/scala/scaladex/}server ...
[trace] Stack trace suppressed: run last sbtScaladex/compile:scalafmt for the full output.
[error] (sbtScaladex/compile:scalafmt) java.lang.NoSuchMethodError: scala.Predef$.ArrowAssoc(Ljava/lang/Object;)Ljava/lang/Object;
[error] Total time: 8 s, completed Jul 15, 2016 9:23:28 PM
```
/ping @olafurpg 
  
Are you using scala 2.10 somewhere in the project aside from sbtScaladex?

MasseGuillaume:
only the sbt plugin

olafurpg:
This is a known issue with the sbt scalafmt plugin, see
https://github.com/olafurpg/scalafmt/issues/190

Since it seems many are affected but it, I'll try to take a look at it when
I find time.

jvican:
Thanks @olafurpg, there's no rush, though, we can live without it for now.

MasseGuillaume:
#205

PR #190, opened 2016-07-15
Format sources with scalafmt #188

jvican:


server/src/main/scala/ch.epfl.scala.index.server/Api.scala
```diff
@@ -154,95 +184,118 @@ class Api(github: Github)(implicit val ec: ExecutionContext) {
...
 
-        lastNonPreReleases.headOption.map(r =>
-          (
-            r,
-            sortedByVersion.map(_.reference.version).distinct,
-            lastNonPreReleases.map(r => (r.reference.target, r.maven))
-          )
-        )
+          lastNonPreReleases.headOption.map(
+              r =>
```
jvican: This will be fixed soon by Olaf, so we shouldn't care about it now.

model/src/main/scala/ch.epfl.scala.index.model/release/SemanticVersion.scala
```diff
@@ -1,79 +1,82 @@
...
       }
-      
+
       val tv1 = tupled(v1)
       val tv2 = tupled(v2)
-      
+
       def preCmp(pr1: Option[PreRelease], pr2: Option[PreRelease]): Int = {
         (pr1, pr2) match {
-          case (None, None)                                               => EQ
```
MasseGuillaume: @jvican is it possible to align those ?

jvican: Yes, if you enable vertical alignment. I'm not a huge fan, personally, I can do it if you're so interested in it.

MasseGuillaume: I think it help readability. Can you turn it on ?

jvican: Yes, I'll look into it on Monday.

jvican:
Will take care of this tomorrow.

MasseGuillaume:
#205

PR #201, opened 2016-07-20
Change target elastic directory

jvican:


data/src/main/scala/ch.epfl.scala.index.data/elastic/package.scala
```diff
@@ -48,7 +48,7 @@ package object elastic extends ProjectProtocol {
   val maxResultWindow = 10000 // <=> max amount of projects (June 1st 2016 ~ 2500 projects)
   private val base = build.info.BuildInfo.baseDirectory.toPath
   val esSettings = Settings.settingsBuilder()
-    .put("path.home", base.resolve(".esdata").toString())
```
MasseGuillaume: hum this already works https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#resolve(java.lang.String)

PR #202, opened 2016-07-20
Fix #195: Paginate by default 20 items

### LINKED ISSUE #195: Paginate by default 20 items
By default, we paginate every 10 items. I think it'd be better if we do every 20, since 10 is probably shorter than the user will expect, and nobody likes to click links to navigate through all the results.

jvican:


template/src/main/scala/ch.epfl.scala.index.views/package.scala
```diff
@@ -5,7 +5,7 @@ import model._
 
 package object html {
   // https://www.reddit.com/r/scala/comments/4n73zz/scala_puzzle_gooooooogle_pagination/d41jor5
-  def paginationRender(selected: Int, max: Int, toShow: Int = 10): (Option[Int], List[Int], Option[Int]) = {
+  def paginationRender(selected: Int, max: Int, toShow: Int = 20): (Option[Int], List[Int], Option[Int]) = {
```
MasseGuillaume: oh I tough you wanted to show more results per page.
This will make the pagination rendering go like:

`G o o o o o o o o o o o o o o o o o o o o l e` vs `G o o o o o o o o o o l e`

jvican: Doesn't showing more results per page decrease the amount of links shown? I don't get what you mean.

MasseGuillaume: I think what you want to change is only this: https://github.com/scalacenter/scaladex/blob/master/server/src/main/scala/ch.epfl.scala.index.server/Api.scala#L20

jvican: Yes, it seems so, thanks.

jvican:
I'm resubmitting changes here.

PR #207, opened 2016-07-20
refactor non standard lib

MasseGuillaume:
@Scyks if you have time to take a look ;-)

with https://github.com/scalacenter/scaladex-data/pull/8


MasseGuillaume:
ops went from a small refactor to a lot of refactoring.

MasseGuillaume:
basically it started when I was investigating for: #209

PR #212, opened 2016-07-21
WIP: edit project

MasseGuillaume:
* [x] remove foreground/background in favor of https://github.com/scalacenter/scaladex/issues/214

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/Project.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/Project.scala
@@ -16,8 +16,6 @@ import misc.{GithubInfo, GithubRepo}
   * @param customScalaDocUrl expression to subsitute scaladoc
   * @param documentationLinks user documentation & etc
   * @param logoImageUrl absolute url to a logo (ex: http://spark.apache.org/images/spark-logo-trademark.png)
-  * @param background background css attributes (see https://developer.mozilla.org/en/docs/Web/CSS/background)
-  * @param foregroundColor color css attribute (see https://developer.mozilla.org/en-US/docs/Web/CSS/color)
   * @param liveData the project was updated/created by a user
   * @param id elasticsearch id only used for updating projects
   * @param created date of the first release
@@ -38,9 +36,6 @@ case class Project(
     deprecated: Boolean = false,
     contributorsWanted: Boolean = false,
     artifactDeprecations: Set[String] = Set(),
-    logoImageUrl: Option[String] = None,
-    background: Option[String] = None,
-    foregroundColor: Option[String] = None,
     liveData: Boolean = false,
     id: Option[String] = None,
     created: Option[String] = None,
--- a/model/src/main/scala/ch.epfl.scala.index.model/Project.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/Project.scala
@@ -10,10 +10,9 @@ import misc.{GithubInfo, GithubRepo, Url}
   * @param repository (ex: spark)
   * @param github github information representation
   * @param keywords predefined keywords (ex: database)
-  * @param stackOverflowTags see http://stackoverflow.com/tags (ex: akka)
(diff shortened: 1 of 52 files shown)


jvican:
Are you sure you want to set the column limit to 120? It doesn't really make sense to have a scala formatter then... I think that we should follow the 80-char column limit, which is a standard in a lot of projects (like Scalajs).

MasseGuillaume:
@jvican 
The 80-char column limit exist because of the IBM punch card.

I code on a BenQ RL2460HT monitor setup on it's vertical alignment. A 120 char wide column happens to fit perfectly horizontally.

jvican:
Yes, and it keeps being a standard because people read code in mobile phones and small laptops. That's why it's useful to still follow the convention. Also, there are studies showing that vertical code is more readable than horizontal code (take Haskell vs Scala, for instance).

heathermiller:
I agree with Jorge on following convention. I always format to 80-char column widths in everything that I do, and most people I interact with do the same. If 120 is our limit, there's little need for a formatter.

MasseGuillaume:
@Scyks What is your preference for the column size ?

heathermiller:
To be honest, given that we are so weeks past our deadlines, I think wasting time on this argument is rather moot. Instead, maybe we should focus on the goal line?

Scyks:
Actually I prefer a softlimit of 160 chars. In times of big screens there should not be s Problem. Personally I really don't like 80 chars, it makes code more unreadable for me. But I can live with softlimit of 120.

model/src/main/scala/ch.epfl.scala.index.model/Project.scala
```diff
@@ -24,11 +28,18 @@ case class Project(
...
-    customScalaDocUrl: Option[String] = None,
+    artifacts: List[String] = Nil,
+    releaseCount: Int = 0,
+    customScalaDoc: Option[String] = None,
     documentationLinks: Map[String, String] = Map(),
+    deprecated: Boolean = false,
+    contributorsWanted: Boolean = false,
+    artifactDeprecations: Set[String] = Set(),
     logoImageUrl: Option[Url] = None,
+    background: Option[String] = None,
```
MasseGuillaume: @heathermiller this is actually pretty cool it let project owner customize their page.

server/src/main/scala/ch.epfl.scala.index.server/Api.scala
```diff
@@ -102,21 +100,31 @@ class Api(github: Github)(implicit val ec: ExecutionContext) {
...
                     must(
-                        termQuery("reference.organization", project.organization),
-                        termQuery("reference.repository", project.repository)
+                      references
                     )
                 )
             )
         )
-        .size(5000)
+        .size(500)
```
MasseGuillaume: artifactReference will filter releases so we only have one artifact (ex: cats-core). This is why I can put a lower limit.

server/src/main/scala/ch.epfl.scala.index.server/Server.scala
```diff
@@ -71,47 +71,48 @@ object Server {
       } yield views.html.frontpage(keywords, targets, dependencies, latestProjects, latestReleases, userInfo)
     }
 
+    def canEdit(owner: String, repo: String, userState: Option[UserState]) = {
+      userState.map(s => s.isAdmin || s.repos.contains(GithubRepo(owner, repo))).getOrElse(false)
+    }
+
+    def editPage(owner: String, repo: String, userState: Option[UserState]) = {
+      val user = userState.map(_.user)
+      // if(canEdit(owner, repo, userState))
```
MasseGuillaume: Currently working on it. I will re-enable

MasseGuillaume:
a

@heathermiller @Scyks ^

MasseGuillaume:
2016-07-24-204518_2446x1920_scrot

Issue #219, opened 2016-07-27
Down arrow + enter on dropdown should let you select a library


heathermiller:

MasseGuillaume:
Esc or click away to bail out

PR #233, opened 2016-08-04
#219 Add the ability to close the autocompletion popup with Esc

julienrf:


client/src/main/scala/ch.epfl.scala.index.client/Client.scala
```diff
@@ -88,21 +88,23 @@ trait ClientBase {
   }
 
   def navigate(event: KeyboardEvent): Unit = {
-    if (event.keyCode == 38 && completionSelection.choices.nonEmpty) { // Up
+    if (event.keyCode == 38 /* Up */ && completionSelection.choices.nonEmpty) {
```
MasseGuillaume: https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/ext/Extensions.scala#L102

julienrf: Great, thanks!

Issue #254, opened 2016-08-19
Not possible to see pages 2 and 3 of my projects


Sciss:
So if I log in and click projects, it say I have 53 published projects. Page 1 is fine: https://index.scala-lang.org/search?q=you&you (but note the strange `you&you`; but if I click on page 2 at the end of this page, I get an invalid URL: https://index.scala-lang.org/search?q=youyou&page=2 (note the `youyou`). If I try to manually correct URL to https://index.scala-lang.org/search?q=you&page=2 - I get a selection random projects, none of which is mine. This is with Firefox 45.

MasseGuillaume:
Thanks for reporting the bug ;-)

as a workaround you can use https://index.scala-lang.org/sciss?q=organization:sciss&page=1

MasseGuillaume:
something is wrong here: https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/searchresult.scala.html#L13

MasseGuillaume:
solved in https://github.com/scalacenter/scaladex/pull/269

PR #264, opened 2016-09-30
do not add "you" as it is gonna be appeared in "q" taken from GET

novakov-alexey:
It is probably intermediate fix. 
We already get the "you" from GET request based on the "Projects" button URL, so that no additional "you" is needed in the pagination URL.


jvican:
Doing a clean clone on scaladex reveals that the master is dirty and some tests are failing. I think this changeset looks good, but given the current state of master I won't merge. Can you have a look @MasseGuillaume ?

jvican:
Fixes #254.

MasseGuillaume:
@novakov-alexey I tested locally and the issue was bit more complex:

here is how I solved it: https://github.com/scalacenter/scaladex/pull/269

novakov-alexey:
Yes, indeed. Now, it is more clear. I can try to fix other simple issue.

Issue #81, opened 2016-06-16
Feedback


olafurpg:
Really excited to see this take off 😄 Here are a few things I noticed.
- [x] I would be really nice to see immediately what Scala versions libraries support (e.g., 2.10/2.11/..., Scala.js)
  - targets in [project](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Project.scala#L45) (ex: scala_2.11, scala.js_0.6)
  - [seach result html templating](https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/result.scala.html#L42) 
**moved to #288**
- [x] Dependents should not include libraries from the same repo, for example https://index.scala-lang.org/typelevel/cats-free has https://index.scala-lang.org/typelevel/cats has a dependent
  - [html templating](https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/project/dependents.scala.html)
  - [release model](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Release.scala#L37)
  - same repo == same [Project.Reference](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Release.scala#L215)
- [x] Dependencies includes sbt plugins and test dedendencies. For example for [scalafmt-core](https://index.scala-lang.org/olafurpg/scalafmt-core), it lists 6 dependencies out of which only scalameta and sourcecode are actual dependencies to the published library.
- [x] The CSS is slightly broken in Chrome (fixed in #75)

<img width="390" alt="screen shot 2016-06-16 at 07 37 15" src="https://cloud.githubusercontent.com/assets/1408093/16116374/c8132742-33cc-11e6-95bc-c712bf6734b3.png">
- [X] The recent releases list doesn't contain links to the projects, only their names (fixed in #74)
- [x] The recent releases seem to have duplicates, for example the top three libraries now are
  - kaliber - play-s3 6.0.0
  - kaliber - play-s3 8.0.0
  - kaliber - play-s3 7.0.2
- [X] The version number defaults to the latest release even if it's a RC. I think it would be cool to show the latest stable release as well. (#78)
[...]

PR #265, opened 2016-09-30
Separate internal dependencies from external dependencies

alexdeleon:
This should fix the first point of issue #81 

screen shot 2016-09-30 at 19 12 20

### DIFF
--- a/template/src/main/twirl/ch.epfl.scala.index.views/project/dependencies.scala.html
+++ b/template/src/main/twirl/ch.epfl.scala.index.views/project/dependencies.scala.html
@@ -11,6 +11,12 @@
   }
 </div>
 
+<div class="dependencies box">
+  <h4>@release.internalDependencyCount Internal Dependencies</h4>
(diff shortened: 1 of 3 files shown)


template/src/main/twirl/ch.epfl.scala.index.views/project/dependencies.scala.html
```diff
@@ -11,6 +11,13 @@
   }
 </div>
 
+@if(release.internalDependencies.nonEmpty) {
+<div class="dependencies box">
+  <h4>@release.internalDependencyCount Internal Dependencies</h4>
```
olafurpg: Some day, we should pluralize these messages so that it's
"1 internal dependency"
" 2 internal dependencies"
This PR is fine because it's consistent with the other boxes.

MasseGuillaume:
@alexdeleon thanks! I just deployed your feature on https://index.scala-lang.org.

PR #266, opened 2016-10-01
Fixed failing tests

alexdeleon:
I noticed these failures on the unit test while working in PR #265 . One problem was a compile error and the other was lack of a expected configuration value by the code being tested.


alexdeleon:
BTW @MasseGuillaume why don't we user Travis? I think you guys have a Jenkins in place but Travis will be more visible to people contributing to the project.

MasseGuillaume:
@alexdeleon Travis is limited in terms of memory / cpu usage. We need a lot of memory for the indexing (3GB). I can put Travis back to run the tests/compile.

alexdeleon:
I think Travis would be useful just for compile and unit test. It will help verifying pull requests.

Issue #256, opened 2016-08-24
improve ammonite/coursier install section


MasseGuillaume:
coursier should only show if it's a command line artifact

ammonite should not show up if it's on bintray

MasseGuillaume:
* [install html template](https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/project/install.scala.html)

we could add a "is a command line artifact" checkbox to the edit page of a project:
* [edit html template](https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/project/editproject.scala.html)
* [project model](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Project.scala#L27)
* [project form](https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala#L7)

MasseGuillaume:
https://github.com/scalacenter/scaladex/commit/c2b865ebd10a280d8bf085d80f7a62e89b5778a2

PR #268, opened 2016-10-02
Ammonite should not show up if it's on bintray

alexdeleon:
This resolves the first part of issue #256. Can be tested with project `marpiec/mpjsons` which is on Bintray only

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/Release.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/Release.scala
@@ -204,12 +204,10 @@ case class Release(
   }
 
   /**
-    * True is these release artifact is hosted on Bintray but not Maven central
+    * True is this release artifact is hosted on Maven Central
     */
-  lazy val isHostedOnBintray = resolver.map {
-    case _: BintrayResolver => true
-    case _ => false
-  }
+  lazy val isOnMavenCentral = resolver.isEmpty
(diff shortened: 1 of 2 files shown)


model/src/main/scala/ch.epfl.scala.index.model/Release.scala
```diff
@@ -201,6 +202,14 @@ case class Release(
 
     orderedReverseDependencies.filter(d => d.reference.name == dep.reference.name).map(_.reference.version)
   }
+
+  /**
+    * True is these release artifact is hosted on Bintray but not Maven central
+    */
+  lazy val isHostedOnBintray = resolver.map {
```
MasseGuillaume: it could be `def isOnMavenCentral = resolver.isEmpty`

alexdeleon: Agree, it's better like that

MasseGuillaume:
@alexdeleon hehe really sorry about that. Thanks for staying in the loop. If you need help I'm always in the gitter.im channel: https://gitter.im/scalacenter/scaladex

MasseGuillaume:
Hey @alexdeleon, I really want to merge this, do you need help?

MasseGuillaume:
I'm closing this PR. We can reopen it later ;-).

PR #270, opened 2016-10-04
bintray logo

MasseGuillaume:


heathermiller:
These logos really break the layout. Especially on mobile. We really need to spend some effort trying and keep the UI in tact. As it stands, it will be expensive correcting the multitude of issues that have arisen since the initial UI design. We can't keep quickly/carelessly throwing in elements like this without thinking about how it looks, especially across devices. This + other changes have degraded usability over time.

PR #274, opened 2016-10-08
add specific routes for web-resources; move sass files to css folder.

novakov-alexey:
@MasseGuillaume just used your recommendation to fix it in http routes only without touching sbt.


server/src/main/scala/ch.epfl.scala.index.server/Server.scala
```diff
@@ -383,9 +383,17 @@ object Server {
...
+              getFromResource("lib/" + path)
+            } ~
+            path("assets" / "img" / Remaining) { path ⇒
+              getFromResource("img/" + path)
+            } ~
+            path("assets" / "css" / Remaining) { path ⇒
+              getFromResource("css/" + path)
+            } ~
+            path("assets" / "js" / """\w*[\-\w]*\.js[\.map]*""".r) { path ⇒
+              getFromResource(path)
```
MasseGuillaume: I pushed a commit on top of this one to fix this bug:

it would still allow access to anything in the classpath: for example `/assets/js/application.conf` would get `application.conf` dirrectly

MasseGuillaume:
Thanks! I added a commit to make sure we hit all cases. jQuery was pointing to the correct version (2.3.x)

server/src/main/scala/ch.epfl.scala.index.server/Server.scala
novakov-alexey: Actually, it would only allow files with *.js or *.js.map to get from classpath. But probably simple "get" routes are better.

MasseGuillaume: oh correct! Sorry my eyes focused on `getFromResource(path)` without a prefix.

PR #276, opened 2016-10-12
Return a “no published release” badge rather than a 404 when there is no known release of a project

julienrf:


MasseGuillaume:
rebased in https://github.com/scalacenter/scaladex/pull/277/commits/af1e662818a8fd5ce8bb9538daf3f56a4075cd01

PR #283, opened 2016-10-23
Fix issue with some user avatars

adamwy:
It's definitely a GitHub bug, which I have already reported, but as a workaround for now we can enforce image size to 38 pixels which seems to fix the problem.

Here is a screenshot of what it looked like:
scaladex-avatar


MasseGuillaume:
cool thanks !

PR #286, opened 2016-10-25
Improvements to the user projects button

adamwy:
Makes user projects button more consistent with the rest of UI.

The following screenshot presents comparison between current version and both improvements from this PR:
screenshot from 2016-10-25 11 15 05


MasseGuillaume:
Thanks !

PR #308, opened 2016-12-28
Correct clone URLs in contributors guide

vossad01:
The git links used would fail with `Permission denied (publickey)`.  Using the http links solves this issue.

Also set depth for the index data because they are large and intended as read-only anyway.


MasseGuillaume:
Those are public repo. I think you need to setup your ssh key.

PR #310, opened 2016-12-28
Made page title come before site tite in title bar

vossad01:
This puts the page title before the site title.  This puts the most pertinent information first improving usability.

Pages are now titles (example):
 > akka ! Scaladex

instead of,
 > Scaladex ! akka

When having multiple tabs open, many browsers narrow the tab reducing the number of visible characters.  Prior to this change, a user would have many tabs open all indicating "Scalad..." but would not be able to tell at a glance what project they were for.  With this change it is more likely a user will be able to identify a tab without switching to it.


MasseGuillaume:
T'y:)

PR #311, opened 2016-12-28
Hyperlink Statistics and Footer Logos

vossad01:
As a user, I expect logos to be hyperlinked so that I can easily navigate to get more information.
As a user, I expect statistics to be hyperlinked so I can verify the claim or get more information.

The only statistic not hyperlinked by this is Releases since I was not sure the clear place to go for that information.

The footer logos are "Scala Center" and "JFrog Bintray"

### DIFF
--- a/server/src/main/assets/css/partials/_main.scss
+++ b/server/src/main/assets/css/partials/_main.scss
@@ -370,6 +370,9 @@ main {
                     }
                 }
             }
+            a {
+                color: inherit;
+            }
         }
         .install {
             .nav-tabs > li:not(.active) > a {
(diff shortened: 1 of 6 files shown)


vossad01:
I noticed this is live :+1:
I'm not sure if it is considered an issue or if it is already known (cache is a pain) but thought I should report just in case.

My browser was still using the cached version of the CSS making the statistics orange for hyperlinks making me momentarily wonder if I missed one of my changes.  Flushing the cache caused the new CSS to be downloaded and they returned to their usual grey.

MasseGuillaume:
Hum, it look like we need to tag our assets with some hash to avoid caching.

Issue #309, opened 2016-12-28
"Contributors Wanted" badge does not provide explanation


vossad01:
Ideally, the "Contributors Wanted" badge should be hyperlinked to a project-specific page that directs a user how to contribute.

This may be:

- The Project's Issue tracker
- The Project's CONTRIBUTING.md file
- Some other project-specific resource that gives direction on how to get involved.

In the absence of such direction the badge does not really provide much information and may even be hurtful to projects that don't have it.  Perhaps this view is not shared, but when a project is on GitHub, I assume that it wants contributors unless it indicates otherwise.  The badge on Scaladex suggests the default is the other way around: projects don't want contributors unless they say they do.

heathermiller:
I agree that we should give the library author an option of something to link to when someone clicks on the "contributors wanted" badge. Makes sense 👍 

Though to respond to this point:

> In the absence of such direction the badge does not really provide much information and may even be hurtful to projects that don't have it. Perhaps this view is not shared, but when a project is on GitHub, I assume that it wants contributors unless it indicates otherwise.

Nah, actually this isn't true. There's actually numbers and research done on this (I'd have to dig around for a while to find citations for you), but it's actually very common for companies or individuals to post code on GitHub for the sake of open sourcing it (for others to fork), but with little or no intentions of accepting contributions. In fact, many companies even have internal policies not to accept outside contributions (this is the case for many Microsoft projects that have been open sourced, for example). Yet, on the other hand, there are libraries that are scrambling for help, and there's no way to visually indicate on which side of the spectrum some library is. 

So I would disagree that the "contributors wanted" badge is bad thing. Instead, I think we could invest more effort here helping library authors to communicate their needs to potential contributors via Scaladex... Help wanted brainstorming what we could do to better help this! 😄

MasseGuillaume:
we could link to the following search: 
https://index.scala-lang.org/search?q=contributorsWanted%3Atrue

vossad01:
@MasseGuillaume I think that is best for now as it assigns some more meaning to the badge.

My best ideas on how to improve the help wanted badge towards making it useful are in the OP.

[Up For Grabs [.net]](http://up-for-grabs.net/#/) is the dominant player I know in this space.  It establishes a convention and requires a commitment in grooming issues.  I have looked in the past at submitting more Scala projects to it, but for the handful I looked at, I was not convinced of my ability to correctly identify a tag meeting the appropriate requirements.

[...]

PR #312, opened 2016-12-29
Make 'Contributors Wanted' badge hyperlink to all such projects

vossad01:
Per https://github.com/scalacenter/scaladex/issues/309#issuecomment-269647840, this makes clicking the "Contributors Wanted" badge link to https://index.scala-lang.org/search?q=contributorsWanted:true so users can find all such projects.

Longer term, https://github.com/scalacenter/scaladex/issues/309#issuecomment-269676545 is probably the desired implementation but is also more involved.

More discussion can be found on #309


MasseGuillaume:
awesome :)

PR #322, opened 2017-01-04
Fix type in example

lunaryorn:
"credentials" is a list; we need += to add a value.

### DIFF
--- a/doc/user/publish-scaladex.md
+++ b/doc/user/publish-scaladex.md
@@ -22,7 +22,7 @@ Add the following to your build.sbt file:
 
 ```scala
 scaladexKeywords in Scaladex := Seq("Foo", "Bar", "Baz")
-credentials in Scaladex := Credentials(Path.userHome / ".ivy2" / ".scaladex.credentials")
+credentials in Scaladex += Credentials(Path.userHome / ".ivy2" / ".scaladex.credentials")
 /*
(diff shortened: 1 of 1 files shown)


MasseGuillaume:
Thanks

lunaryorn:
@MasseGuillaume You're welcome, thanks for merging so fast.

By the way, do you happen to know why the document scopes `credentials` in `Scaladex`?  Does that make a difference?

Like most people I already got a global `credentials.sbt` with my Sonatype credentials; adding my Scaladex credentials globally would require me to install the plugin globally to have the Scaladex scope available.  Since I don't want to do that I've added the credentials without the scope restrictions, and things still seem to work well, so I'm left quite confused 😕 :blush:

MasseGuillaume:
Hum, it should be scoped: https://github.com/scalacenter/scaladex/blob/master/sbtScaladex/src/main/scala/org.scala_lang.index.sbt/ScaladexPlugin.scala#L41

I'm not sure why it works.

lunaryorn:
@MasseGuillaume Well, according to `inspect actual scaladex:credentials` it depends (i.e. delegates?) to `*:credentials`, i.e. what I change with `credentials += …` in a global `credentials.sbt`.

That's why it works, but I wonder whether that's intentional, given that the documentation makes the scope explicit.

Issue #314, opened 2016-12-30
Clarification of "Dependents", "Dependencies", and "Artifact Dependents"


vossad01:
Dependents, Dependencies, Artifact Dependents
Perhaps these are the established terms, well understood by those in the know, and there is nothing that can be done about non-intuitive behavior.  However, as is, it is the changing frames of reference make Scaladex and even individual pages within it confusing as I will try to show below using akka-actor as an example.

![image](https://cloud.githubusercontent.com/assets/965439/21557816/4672a2e2-cdf7-11e6-9290-d0238c6fa933.png)

Statistics shows **14 Dependents**
The Dependents box shows **13 Dependents**

* Unless there is an unrelated bug here, _Dependents_ means two different things on the same page.  I don't know what the difference is to know how to remedy this aspect.

* If I click on the "Artifact Dependents" link I am taken to a Scaladex search for "dependencies:akka/akka-actor".  I thought I was looking for "dependents" but now am searching for "dependencies" which is surprising.
After some thought, the user may realize they just changed frames of reference and now they are looking for _things where dependencies contains akka/akka-actor_ rather than _things that depend on akka/akka-actor_.  From a technical perspective, this makes sense and they are clearly equivalent, but from a user perspective it can be confusing.  If "depends-on:akka/akka-actor" would be an options instead of "dependencies:akka/akka-actor" that is the best alternative I have thought of.

* I can understand _Artifact Dependents_ as everything that depends "akka-actor", but then I am left not knowing what _Dependents_ is.  It is not a proper subset of _Artifact Dependents_.  Unless told otherwise, I will assume who have worked with on large projects understand the distinction.

MasseGuillaume:
We use dependent to denote the opposite of dependencies. We could also use reverse dependencies.

> Unless there is an unrelated bug here, Dependents means two different things on the same page. I don't know what the difference is to know how to remedy this aspect.

They should be equal.

> depends-on:akka/akka-actor" would be an options instead of "dependencies:akka/akka-actor"

This could work too. Good idea.

vossad01:
> Statistics shows **14 Dependents**
The Dependents box shows **13 Dependents**

Tracked this part down.  The difference is because the lower filters out duplicates and in the case of akka-actor the following dependencies are in the Seq

> ScalaDependency(Reference(hseeberger,akka-log4j,**akka-log4j, 1.2.2**,Some(ScalaTarget(2.12,None))),Some(compile)),
[...]

PR #324, opened 2017-01-04
Improve consistency of dependent, term and occurences

vossad01:
Pursuant to #314 this does the following:

 * Makes the statistic for number of dependents and the dependents box indicate the same number of dependents
 * Support for **depends-on:** as an alias for **dependencies:** in queries.
 * Usages within Scaladex changed to use **depends-on:**.  No change was made to the behavior of **dependencies:**
 * Miscellaneous changes from 'dependencies' to 'dependents' where the former had been used but conceptually the latter was represents.

This does not address clarifying "depedents" vs "artifact dependents"


MasseGuillaume:
awesome, well done!

PR #332, opened 2017-01-10
Display artifact’s binary targets in the result list

### LINKED ISSUE #288: Display which targets are library supports
splitting from #81

- [ ] I would be really nice to see immediately what Scala versions libraries support (e.g., 2.10/2.11/..., Scala.js) + target logos
  - targets in [project](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Project.scala#L45) (ex: scala_2.11, scala.js_0.6)

- [ ] improve the front page ui with some target logos

julienrf:
Before:
before

After:
after

What do you think?

Fixes #288.

### DIFF
--- a/server/src/main/assets/css/partials/_main.scss
+++ b/server/src/main/assets/css/partials/_main.scss
@@ -228,12 +228,11 @@ main {
         margin: 0;
         border-bottom: 1px solid rgba($gray-dark, 0.12);
         h4,
+        .targets,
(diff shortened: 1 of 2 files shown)


julienrf:
Note that I don’t display supported Scala.js versions. I think it would be good to show both, but I couldn’t find a lean way to do it.

PR #325, opened 2017-01-05
Minimal implementation of artifact pages

vossad01:
This provides artifacts their own page resolving some abstraction issues with displaying artifact information on the project page.

This is not perfect in that there are a lot of additional things that I can think of that someone might iterated on top of this, but I do not believe it regresses and it takes care of all the hard (busy) work of providing the two different types of pages.

**Example Project Page:**
screen shot 2017-01-04 at 16 06 05-fullpage

**Example Artifact Page:**
screen shot 2017-01-04 at 16 11 23-fullpage

### This partially addresses:
 - [Display which targets are library supports, #288](https://github.com/scalacenter/scaladex/issues/288): Provides information without graphical logos; no front-page or search results modifications.

### This obsoletes:
 - #320
 - #317
 - #313

PR #340, opened 2017-01-14
Artifact pages

### LINKED ISSUE #313: Pressing "Back" after changing the "Artifact" does not restore the previous "Artifact"
Steps to reproduce:
 1. Got to https://index.scala-lang.org/akka/akka/akka-actor
 2. Choose "akka-camel" from the artifact selection box.
 3. Note that Statistics, Dependencies, and Dependents change
 4. Press Back
 5. Notice that Statistics, Dependencies, and Dependents change back but the artifact selection box still says "akka-camel."  The page is now in a state of presenting false information.

### LINKED ISSUE #317: Invalid navigation paths presented in project select boxes
Steps to reproduce:

- Got to https://index.scala-lang.org/akka/akka-http/akka-http-core
- Under the Scala version drop down choose "scala 2.11.0"
- Observe you are then redirected to a 404 Page Not Found page

### LINKED ISSUE #326: Display artifacts in search results instead of projects
There is an inconsistency between the perspective of the Search Box and the Search Results.  The search box expects at "$project/$artifact" while the search results show "$organization/$project"

For example if I query for "depends-on:akka/akka-actor" I see the top result is "scala-academy/performance-analysis".
If I then query on "depends-on:scala-academy/performance-analysis" to see what "scala-academy/performance-analysis" depends on I get no results because "scala-academy/performance-analysis" is not an artifact.

Similarly, if I search for "scala-reflect":
"scala/scala" comes up as the first result, but "scala/scala-reflect" is nowhere to be seen because "scala/scala-reflect" is an artifact.

To resolve this the search results should display artifacts and not projects.

[Discussion leading to this conclusion began with this post on gitter.](https://gitter.im/scalacenter/scaladex?at=586e5b48c02c1a3959edeb0c)

MasseGuillaume:
rebased https://github.com/scalacenter/scaladex/pull/325, thanks @vossad01!

fix #313, fix #317 (no more artifact dropdown)
fix #326


MasseGuillaume:
@vossad01 changes are live https://scaladex.scala-lang.org/

I removed the artifact list from the search page. https://scaladex.scala-lang.org/search?q=*

we can revert this from this commit: https://github.com/scalacenter/scaladex/commit/40b4eac7d703b09f268708e62a5b7e30819237d0 if we want to keep it.

vossad01:
Thanks for letting me know :)

I figure it will be decided based on whether anyone's workflow is affected
enough to merit that change (or some alternative for it.)

heathermiller:
I'm happy that folks are working on this issue, but as I've said in other tickets, I think that as-is this is a very confusing change and should likely be reverted until it's 100% worked out. 

Now a user has no idea that they have to manually select an artifact every time they visit a Scaladex page–they didn't have to do this before before–this is a confusing change in UI behavior that confuses returning visitors. It literally took me minutes to figure out that I had to select an artifact on the right side of the page. For the many people that use Scaladex every day, we're collectively wasting peoples' time and confusing them when I believe it's not necessary at all. What has improved so much so as to warrant this new confusion we're going to subject users to?

Further, I have no idea from looking at the project page, which targets a project is available for. I feel silly saying this, but if someone want to know if a specific project has been compiled for Scala.js, the no-brainer solution is to go to the project's page on index.scala-lang.org to see which targets the library has been compiled for. Now that information is nowhere to be seen. We have instead **reduced** the amount of information that Scaladex now displays, which is a no-go. Especially when people are depending on Scaladex day-to-day. 

Again, I think these changes should be reverted until they are 100% worked out.

MasseGuillaume:
reverting changes: https://github.com/scalacenter/scaladex/pull/360

@vossad01 I'm really sorry about that :-(

vossad01:
@heathermiller, I am sorry to have broken your workflow.

Can you please help me to understand the issues you see needing to be “100% worked out?”

heathermiller:
Hi @vossad01, I like where this change is going (an artifact page is in general is a good idea), it's just that the workflow in general for everybody (not just my personal workflow) is suddenly broken, with important info now gone or more difficult to find. 

Simply put, the addition of artifact pages, as currently proposed, adds more steps, and more clicks, to the process of getting some of the most basic information about a library.

Take for example any other package index (cargo, npm, etc). The first thing you tend to see on any other library's page is its version number. The version number is as important as the name of the project. E.g, 

image

For Scaladex, this information is now buried one more click down. A user has to now select a specific artifact (when arguably we should just show a default artifact), without knowing which artifact to choose. Which artifact is "cats"? Or "shapeless"? etc. It requires two or three new actions from a user; they have to now (1) realize they have to click on an artifact (they didn't have to click on one before), (2) they now have to make a decision about which artifact they want (when many times they want whatever the default one is), and (3) they have to then click the artifact to discover what the most current version is.

This entire line of reasoning is the same for targets. Now, asking the simple question, "does library xyz compile for Scala.js" requires a lot of clicking. I'm aware that Guillaume worked on adding a filter panel to the query page, but if I skip the query page and arrive at http://index.scala-lang.org/akka/akka and I want to figure out "does _akka_ compile to Scala.js?" By that I probably want to figure out whether `akka-actor` has a Scala.js target. But with the current changes, I have to make all kinds of decisions and click around just to get this information. 

Before, it was possible to answer these questions without having to click through several artifact pages. 

That said, I'm really not convinced at the moment that adding artifact pages (as currently proposed) is going to make users' lives any easier. Every way I look at it currently, I just see that we're adding extra steps to peoples' workflows without even notifying them that their workflow has changed. They just get this rude awakening where they're suddenly confused because the UI has changed and maybe they're not even sure whether or not this new behavior is a bug. Most UIs that have many users, when they're changed around, have dialog boxes that pop up the first time a returning user visits the new UI and explains what's changed. We don't do that here. 

Lastly, I'm definitely open to changing these things so long as users' lives are indeed improved. I'm just struggling to understand right now how this is more convenient for end users. Every day I look at it, it just seems to me that extra steps were added, and I'm not sure what the overall benefit is to the end user. Maybe you could explain?

vossad01:
I record the following suggestions, trying to restate them constructively:

* Displaying the currently selected version more prominently on the Artifacts page would have been nice
* In addition to being able to specify the version of an artifact, it is important to be able to specify a target for the artifact as well.  Reasoning unspecified.
* The concept of a default artifact is useful.  Manifestation unspecified.
* It would be nice if the Project Page had [aggregated] information about the supported targets of its Libraries.
* It would be cool if someone implemented an in-page tutorial for Scaladex that could tell users how to use it and inform them of changes.
___

>  Maybe you could explain?

For the end-user who already understands how Scaladex was working, intuitively knows the inconsistencies, and doesn’t mind the slow behavior of the triple-select drop-down, there is probably not an immediate benefit.

For all other users, the artifacts page helps make the system more intuitive, accurate, responsive, and less error prone.  Returning users not falling into the first category may even realize they had not accurately understood Scaladex previously.

Scaladex tries to present that a GitHub Project and Library are equivalent.  This allows for nice things, like pulling the README.md from GitHub.  Some package managers like NPM and Bower require one repository per library/component making the equivalence true.

Most other sources, Crates.io/Crates, NuGet, Scaldex, etc., do not have this this restriction and allow for multiple libraries from a single git repository.  Because of this, the Project-Library equivalence does not hold.  It may be applicable for some projects but counter-examples are easy to find on Scaladex.

The hybrid Project-Artifact page does a convincing job at presenting that the Project is the Library which makes it very easy to not see the impedance.  Not knowing that distinction results in users not understanding the the information being presented (ex. What are the various dependencies numbers and why do they change) or actually erroneous interpretations of the data (Project x does not support indicated Scala targets, only Library y of project x does).  If this problem were contained to the Project page, significant UI changes could probably reify the concept of the Artifact on a shared page and eliminate many UX problems.  However, the impedance also been seen with search results, with dependents/dependencies, and potentially future extensibility of Scaladex.   Improving the hybrid page does less in addressing those issues.

MasseGuillaume:
> Scaladex tries to present that a GitHub Project and Library are equivalent. This allows for nice things, like pulling the README.md from GitHub. Some package managers like NPM and Bower require one repository per library/component making the equivalence true.

My vision of the UI is to split the project and the release part with the sidebar:

Project | Release

Currently we have something like:



```
|           Project Head               | Selectors |

|              <README>                |  <stats>  |
|                 +                    |   <sbt>   |
|                 +                    |   <doc>   |
|                 +                    | <contribs>|
|                 +                    | <license> |
|                 +                    |  <deps>   |
|                 +                    |  <deps>   |
|                 +                    |           |
|                 +                    |           |
```

I think we could put everything related to the project on the left, so it's clear that everything on the left is project related and everything on the right is release related. We can change the target selector to a radio group such that the user see all targets available.

```
|           Project Head               | Selectors |

|           <license, stats>           |   <sbt>   |
|             <README>                 |   <doc>   |
|                 +                    |   <deps>  |
|                 +                    |           |
|                 +                    |           |
|                 +                    |           |
|              <contribs>              |           |
```

PR #343, opened 2017-01-14
Update favicon to reflect Scaladex

vossad01:
Previously, the favicon indicated Scala Center. Now it shows the Scaladex logo.

This utilized http://realfavicongenerator.net to provide appropriate images and configuration for many platforms.

With this you can visually distinguish between Scaladex pages and Scala Center pages:
scaladex vs scalacenter

I didn't do anything as far as cache busting with this so it would probably take a long time for previous visitors to see the change.  Perhaps I should have?


MasseGuillaume:
nice :)

PR #345, opened 2017-01-14
Makes the organization page redirect to search

vossad01:
This reduces duplication on the site by making it so there are no longer two places to search for an organization's projects.

This makes it less likely that links to "/organization" will proliferate, leaving that path available (less likely to break external references) if an organization page is ever added.

With this change, navigating to:

> https://scaladex.scala-lang.org/akka

takes the user to:
> https://scaladex.scala-lang.org/search?q=organization:akka

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/routes/SearchPages.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/routes/SearchPages.scala
@@ -44,25 +44,8 @@ class SearchPages(dataRepository: DataRepository, session: GithubUserSession) {
         }
       } ~
         path(Segment) { organization =>
-          optionalSession(refreshable, usingCookies) { userId =>
-            parameters('page.as[Int] ? 1, 'sort.?) { (page, sorting) =>
-              pathEnd {
-                val query = s"organization:$organization"
-                complete(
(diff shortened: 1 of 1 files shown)


MasseGuillaume:
nice simplification!

heathermiller:
Ah, I really don't like this change. 

Showing `index.scala-lang.org/orgname` is convenient, à la GitHub. People realize just by looking at the URL that they can search for all repos by a given organization via the URL without having to write the query by hand. 

I would really rather that this was like before. i.e., you type in `index.scala-lang.org/orgname` and the URL isn't rewritten.

vossad01:
This change was reverted by  #359.

To clarify on the original change this change,

`/orgname` was still allowed provided the user the same content as before functional (Scaladex itself linked to it for that matter).  Should the user try navigating by URL hacking they would also get the expected results.  Though, I would pose that such URL probably means Scaladex is not doing something well and should strive to improve the deficient workflow.

GitHub has a distinction between an Organization page: http://github.com/scalacenter/
And Search: https://github.com/search?utf8=✓&q=user:scalacenter

Currently Scaladex only has search and the original change was specifically to allow for and reduce the impact of adding a GitHub-like organization page in the future.

PR #361, opened 2017-01-17
Unit test routing

vossad01:
Done as a rather long refactor to try reduce the chances of breaking something, this decouples the routing logic from service behavior allowing for the behavior of the Akka HTTP route/path configuration to be tested without needing to initialize any of the rest of Scaladex.

NB.  While I was pulling in akka-http-testkit I also updated to a non-experimental version of Akka HTTP.


MasseGuillaume:
@vossad01 nice! reviewing now.

server/src/test/scala/ch.epfl.scala.index.server.routes/RoutesTests.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.{Route, StandardRoute}
+import ch.epfl.scala.index.server.UserState
+import ch.epfl.scala.index.server.routes.spec2.SpecificationRouteTest
+
+
+object RouteTests extends org.specs2.mutable.Specification with SpecificationRouteTest {
```
MasseGuillaume: I don't think this is a good testing strategy. Let's take PublishApi in the master branch for example

```
class PublishApi(paths: DataPaths, dataRepository: DataRepository, github: Github) // ...
```

We can see here it depends on DataPaths, DataRepository and Github. This is what you want to mock.

Let's first define the specification for PublishApi. It's used for two things currently: receiving new releases from Sonatype and receiving new releases from the sbt plugin.

* You can't publish something that is already there (GET /api/publish part)
* When you publish a release it must contain a POM, with the exception of sbt, it sends sha1 and other non-relevant fies and they must be ignored.
* When you publish a release you must be authenticated with Github
* Github Users need to own the repository in the scm tag of their release, but Sonatype can publish on behalf of any user.
* ...

You can also take a shortcut and only mock the Github Service.

PR #365, opened 2017-01-19
Fix ordering of targets in search results

### LINKED ISSUE #363: supported targets in search result list should be ordered by version
currently ordered by string

currently: 
scala/scala 2.9, 2.8, 2.7, 2.6, 2.12, 2.11, 2.10

expected:
scala/scala 2.6, 2.7, 2.8, 2.9, 2.10, 2.11, 2.12

see https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala#L54

adamwy:
Fixes #363.

ordering-fix

### DIFF
--- a/template/src/main/scala/ch.epfl.scala.index.views/package.scala
+++ b/template/src/main/scala/ch.epfl.scala.index.views/package.scala
@@ -108,7 +108,8 @@ package object html {
     val prefix = "scala_"
     project.targets
       .collect { case t if t.startsWith(prefix) => t.drop(prefix.length) }
-      .to[List].sorted(Ordering.String.reverse)
+      .map(SemanticVersion.apply).flatten
+      .to[List].sorted(SemanticVersion.ordering.reverse)
       .mkString(", ")
   }
 }


template/src/main/scala/ch.epfl.scala.index.views/package.scala
```diff
@@ -108,7 +108,8 @@ package object html {
     val prefix = "scala_"
     project.targets
       .collect { case t if t.startsWith(prefix) => t.drop(prefix.length) }
-      .to[List].sorted(Ordering.String.reverse)
+      .map(SemanticVersion.apply).flatten
```
julienrf: You could use `flatMap` instead of `map` + `flatten`.

adamwy: It looks like it collides with `SemanticVersion.apply` defined for its case class:

`[error] /home/adam/IdeaProjects/scaladex/template/src/main/scala/ch.epfl.scala.index.views/package.scala:111: type mismatch;
[error]  found   : (major: Long, minor: Long, patch: Option[Long], patch2: Option[Long], preRelease: Option[ch.epfl.scala.index.model.PreRelease], metadata: Option[String])ch.epfl.scala.index.model.SemanticVersion <and> (version: String)Option[ch.epfl.scala.index.model.SemanticVersion]
[error]  required: String => scala.collection.GenTraversableOnce[?]
[error]       .flatMap(SemanticVersion.apply)`

MasseGuillaume:
looks good ;-).

template/src/main/scala/ch.epfl.scala.index.views/package.scala
julienrf: Alright, you should actually write it a little bit differently in order to get the implicit conversion triggered, but that doesn’t really matter, let’s merge it as it is now!

MasseGuillaume:
Thanks!

MasseGuillaume:
@adamwy now live: https://scaladex.scala-lang.org/search?q=*&page=1

adamwy:
Awesome, thanks! :smile:

PR #372, opened 2017-01-31
Update scalafmt

lunaryorn:
I noticed that the scalafmt plugin was quite dated, so I took the liberty to update it :blush:


MasseGuillaume:
thanks!

I use this now: https://github.com/scala-native/scala-native/blob/master/bin/scalafmt#L1

this is why I did not realize it ;-).

PR #384, opened 2017-02-08
Use a default referer if not supplied

### LINKED ISSUE #382: Login link / page doesn't work in all cases
It seems the login resource at https://index.scala-lang.org/login only sometimes manages to do the redirect to github. In other cases, it will interpret the URL as a search as shown here:

screenshot from 2017-02-07 09 53 56

I tried with different browsers. With firefox and chrome clicking on the button seems to work mostly but visiting the link directly does never. On my default chromium (which also has lots of plugins which might interfere) clicking the button doesn't work as well.

julienrf:
Fixes #382.

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/routes/Oauth2.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/routes/Oauth2.scala
@@ -20,13 +20,13 @@ class OAuth2(github: Github, session: GithubUserSession) {
   val routes =
     get {
       path("login") {
-        headerValueByType[Referer]() { referer =>
+        optionalHeaderValueByType[Referer]() { referer =>
           redirect(
             Uri("https://github.com/login/oauth/authorize").withQuery(
               Query(
                 "client_id" -> github.clientId,
                 "scope" -> "read:org",
-                "state" -> referer.value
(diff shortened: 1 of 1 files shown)


server/src/main/scala/ch.epfl.scala.index.server/routes/Oauth2.scala
```diff
@@ -20,13 +20,13 @@ class OAuth2(github: Github, session: GithubUserSession) {
...
       path("login") {
-        headerValueByType[Referer]() { referer =>
+        optionalHeaderValueByType[Referer]() { referer =>
           redirect(
             Uri("https://github.com/login/oauth/authorize").withQuery(
               Query(
                 "client_id" -> github.clientId,
                 "scope" -> "read:org",
-                "state" -> referer.value
+                "state" -> referer.map(_.value).getOrElse("/")
```
jrudolph: Will a relative URL work? I wonder if github can then figure out where to redirect back?

jrudolph: Ah I see it's bounced back to the callback which will do the redirect, so I guess it'll work. Thanks for fixing so quickly :)

PR #385, opened 2017-02-08
Add more logs to trace what happens in the publish API

julienrf:


server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/PublishActor.scala
```diff
@@ -24,12 +22,7 @@ class PublishActor(paths: DataPaths,
 
   def receive = {
     case publishData: PublishData => {
-      if (publishData.isPom) {
         sender ! Await.result(publishProcess.writeFiles(publishData), 10.seconds)
-      } else {
-        if (publishData.userState.isSonatype) sender ! ((BadRequest, "Not a POM"))
-        else sender ! ((OK, "ignoring")) // for sbt, ignore SHA1, etc
-      }
```
julienrf: This has been removed because this logic is duplicated [here](https://github.com/scalacenter/scaladex/blob/f0f95d958f9d8627433c47059e970e02a57f6a88/server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/PublishProcess.scala#L52-L87).

PR #386, opened 2017-02-09
Sbt plugins maven style

julienrf:
Supports indexing sbt plugins published on sonatype.


server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/PublishProcess.scala
```diff
@@ -48,47 +48,51 @@ private[api] class PublishProcess(paths: DataPaths, dataRepository: DataReposito
...
+      Future {
+        data.writeTemp()
+      }.flatMap { _ =>
+        getTmpPom(data) match {
+          case List(Success((pom, _, _))) =>
+            getGithubRepo(pom) match {
+              case None => {
+                logger.debug("POM saved without Github information")
                 data.deleteTemp()
-                (Forbidden, s"${data.userState.user.login} cannot publish to ${repo.toString}")
```
MasseGuillaume: ???

MasseGuillaume:
Thanks !

PR #409, opened 2017-04-11
Add optional subject parameter to latest badge

### LINKED ISSUE #408: Support `subject` parameter for the `latest` badge
This would be useful for the multi-module projects like Alpakka. There is no one artifact named Alpakka but many smaller artifacts that are always released under the same version. So a badge like this would be useful to show in the Alpakka readme:

https://index.scala-lang.org/akka/alpakka/akka-stream-alpakka-amqp/latest.svg?subject=alpakka

Here I chose the first of the artifacts under Alpakka project to query the version number, but would like to change the name in the badge to "alpakka".

2m:
Fixes #408


2m:
Is this in production already?

PR #414, opened 2017-05-10
Add contributor count and filter #271

### LINKED ISSUE #271: Show contributors count & filter by contributors
The number of contributors is a good metric for a healthy open source project. We can show contributor count in the search result list and add a search sorting field based on the contributors.

search result list at: https://index.scala-lang.org/search?q=
- [ ] show contributor count
  - [html template](https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/searchresult.scala.html)
  - [project model](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/Project.scala#L30) this is where the Github information is stored
  - [github info](https://github.com/scalacenter/scaladex/blob/master/model/src/main/scala/ch.epfl.scala.index.model/misc/GithubInfo.scala#L26)
- [ ] add a search sorting field
  - [html template](https://github.com/scalacenter/scaladex/blob/master/template/src/main/twirl/ch.epfl.scala.index.views/result.scala.html#L33)
  - [api](https://github.com/scalacenter/scaladex/blob/master/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala#L24)
  - [elasticsearch sort by array length](http://stackoverflow.com/a/27883807/449071)

MichaelViveros:
Started looking into this so I could get familiar with Scaladex before my GSOC project officially starts. I got a solution working but there's a few issues.

This fixes #271 

### Scripting Language Error

I kept getting an error about the default scripting language (Groovy) not being supported when I tried adding scripting to sort the results of a query by the number of contributors for a repo (see `DataRepository.scala`). This was weird  because the Groovy scripting language is built-in to elasticsearch (es) according to the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/modules-scripting.html). I wasn't able to fix it so instead I downloaded es v2.4.1 from their website and pointed my scaladex es instance to the es instance I downloaded by setting `elasticsearch = remote` in the `org.scala_lang.index.data` config. This fixed the error. I left the config to `elasticsearch = local` in my PR to keep it consistent with how it was.

**So I'm not sure if my scaladex es instance wasn't setup properly or if there's actually something missing from the scaladex repo that needs to be added to support Groovy scripting.**

The es instance I downloaded has a `modules` directory which contained the module for the Groovy language (`lang-groovy`). I couldn't find any such directory in the scaladex repo and I noticed that `.esdata` is the configured home directory for es so this is where the `modules` folder would be but `.esdata` is in the `.gitignore` so it's not tracked under source control. But even after setting up scaladex and running the server, `.esdata` didn't contain a `modules` folder.

Additionally, I used the nodes info API to confirm that the es instance I downloaded has the module for Groovy by doing `curl  http://localhost:9200/_nodes/plugins` and seeing the resulting list of modules. When I did the same thing for my scaladex es instance, the resulting list of modules was empty.

### Security

I found 2 ways to run the Groovy script to sort the results of a query by the number of contributors for a repo:

1. Put the code for the script in a file and reference this file when doing the sorting - [docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/modules-scripting.html#modules-scripting)

This doesn't have any security issues but requires putting the script file in the `.esdata/config/scripts` directory which was being ignored by source control. This is what I implemented in my PR.

2. Use dynamic inline scripting by passing the code for the script as text when doing the sorting

This would involve removing `.esdata/config/scripts/contributors_size.groovy`, updating `DataRepository.scala` to use an inline script instead of a script file and setting the `script.engine.groovy.inline.search` es config to `true`.

New `DataRepository.scala`:
```scala
...
val sortQuery = ...
      case Some("contributors") =>
        scriptSort("""if (_source.containsKey('github')) _source['github'].contributors.size() else 0""") typed "number" order SortOrder.DESC
```

New `ch.epfl.scala.index.data/elastic/package.scala`:
```scala
...
val esSettings =
        Settings.settingsBuilder().put("path.home", base.resolve(".esdata").toString())
            .put("script.engine.groovy.inline.search", true) 
      ElasticClient.local(esSettings.build)
```

Allowing dynamic inline script has some security [issues](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/modules-scripting.html#enable-dynamic-scripting). Es has sandboxed and non-sandboxed scripting languages. I couldn't get the code working for the sandboxed languages so I used a non-sandboxed language, Groovy. The problem with allowing users to run inline scripts with a non-sandboxed language is that users will have the same access to your box as the user that es is running as. The docs recommend doing things like not running es as root user and having a proxy between es and your users, I'm not sure if scaladex is already doing these or not.

Additionally, the `search` part of the `script.engine.groovy.inline.search` config means it only allows search operations (Search api, Percolator api and Suggester api) to execute inline Groovy scripts. So you can't execute inline Groovy scripts when doing `update` operations for example. This may or may not be secure enough for scaladex.

**So which option is best to run the Groovy script?**

### DIFF
--- a/.esdata/config/scripts/contributors_size.groovy
+++ b/.esdata/config/scripts/contributors_size.groovy
@@ -1,0 +1,1 @@
+if (_source.containsKey('github')) _source['github'].contributors.size() else 0
--- a/.esdata/config/scripts/contributors_size.groovy
+++ b/.esdata/config/scripts/contributors_size.groovy
@@ -1,1 +1,0 @@
-if (_source.containsKey('github')) _source['github'].contributors.size() else 0
--- a/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
@@ -12,7 +12,6 @@ import ElasticDsl._
 import ch.epfl.scala.index.data.DataPaths
 import org.elasticsearch.common.lucene.search.function.FieldValueFactorFunction.Modifier
 import org.elasticsearch.search.sort.SortOrder
-import org.elasticsearch.script.ScriptService.ScriptType
 import org.slf4j.LoggerFactory
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -36,7 +35,7 @@ class DataRepository(github: Github, paths: DataPaths)(private implicit val ec: 
       case Some("dependentCount") =>
         fieldSort("dependentCount") missing "0" order SortOrder.DESC mode MultiMode.Avg
       case Some("contributors") =>
-        scriptSort(script("contributors_size") scriptType ScriptType.FILE) typed "number" order SortOrder.DESC
+        fieldSort("github.contributorCount") missing "0" order SortOrder.DESC mode MultiMode.Avg
       case Some("relevant") => scoreSort
       case Some("created") => fieldSort("created") order SortOrder.DESC
       case Some("updated") => fieldSort("updated") order SortOrder.DESC
--- a/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
@@ -12,6 +12,7 @@ import ElasticDsl._
 import ch.epfl.scala.index.data.DataPaths
 import org.elasticsearch.common.lucene.search.function.FieldValueFactorFunction.Modifier
 import org.elasticsearch.search.sort.SortOrder
+import org.elasticsearch.script.ScriptService.ScriptType
 import org.slf4j.LoggerFactory
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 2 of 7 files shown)


MasseGuillaume:
@MichaelViveros thanks, I will take a look this afternoon (in +/- 5h).

.esdata/config/scripts/contributors_size.groovy
```diff
@@ -0,0 +1 @@
+if (_source.containsKey('github')) _source['github'].contributors.size() else 0
```
MasseGuillaume: you should not precompile the script. It complexifies the deployment. This would work only with a local instance of Elasticsearch.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -34,6 +35,8 @@ class DataRepository(github: Github, paths: DataPaths)(private implicit val ec:
         fieldSort("github.forks") missing "0" order SortOrder.DESC mode MultiMode.Avg
       case Some("dependentCount") =>
         fieldSort("dependentCount") missing "0" order SortOrder.DESC mode MultiMode.Avg
+      case Some("contributors") =>
+        scriptSort(script("contributors_size") scriptType ScriptType.FILE) typed "number" order SortOrder.DESC
```
MasseGuillaume: I think you can put the script above verbatim.

MasseGuillaume: Actually it would better to precompute this value. It's less trouble to querry elasticsearch after.

for example: https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala#L299

MichaelViveros: So you mean getting rid of the scripting, adding `contributorSize` as a `Project` field like `releaseCount` and then using `fieldSort` in DataRepository.scala?

MasseGuillaume: exactly

MasseGuillaume:
:+1:

MasseGuillaume:
Well done sir!

PR #427, opened 2017-06-21
Fix Topics for 2.12, #380

### LINKED ISSUE #380: Merge GitHub topics with Scaladex keywords

MichaelViveros:
This fixes #380 

Fixed `SeedElasticSearch.scala` to make "github" field an object field (not a nested field)

Fixed downloading topics (`GithubDownload.scala` and `PlayWsDownloader.scala`) to avoid resource error:
`[AsyncHttpClient-3-1] ERROR p.s.a.i.n.u.c.D.rejectedExecution - Failed to submit a listener notification task. Event loop shut down?`

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
@@ -75,36 +75,38 @@ trait PlayWsDownloader {
       downloadUrl: (AhcWSClient, T) => WSRequest,
       process: (T, WSResponse) => R,
       parallelism: Int,
-      graphqlQuery: (T) => JsObject = null,
-      graphqlProcess: (T, WSResponse, AhcWSClient) => R = null
+      graphqlQuery: Option[(T) => JsObject] = None
   ): Seq[R] = {
 
     val client = wsClient
     val progress = new ProgressBar(message, toDownload.size)
 
+    def processItem(item: T) = {
+      val request = downloadUrl(client, item)
+      val response = graphqlQuery match {
+        case Some(q) => request.post(q(item))
+        case None => request.get
+      }
+      response.transform(
+        data => {
+          if (toDownload.size > 1) {
+            progress.step()
+          }
+          process(item, data)
+        },
+        e => {
+          println(
+            s"error on downloading content from ${request.url}: ${e.getMessage}")
+
+          e
+        }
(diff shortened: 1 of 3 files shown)


data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
```diff
@@ -75,36 +75,38 @@ trait PlayWsDownloader {
       downloadUrl: (AhcWSClient, T) => WSRequest,
       process: (T, WSResponse) => R,
       parallelism: Int,
-      graphqlQuery: (T) => JsObject = null,
-      graphqlProcess: (T, WSResponse, AhcWSClient) => R = null
+      graphqlQuery: Option[(T) => JsObject] = None
```
MasseGuillaume: +1000 :)

data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
```diff
@@ -113,7 +115,11 @@ trait PlayWsDownloader {
     }
     try {
       val response =
-        Await.result(processDownloads.runWith(Sink.seq), Duration.Inf)
+        if (toDownload.size == 1) {
```
MasseGuillaume: what was the issue here ?

MichaelViveros: Initially, I thought the issue was with the old code. After getting a 403 response (abuse rate limit error) and waiting for 60 seconds, the code would bottleneck and instead of each thread resuming at the same time only one thread would resume at a time while the others would keep waiting for a minute which resulted in the whole github download step taking longer. The new code I added got rid of the bottleneck so all the threads would resume at more or less the same time and it takes 3 - 7 minutes less time than the old code.

I also thought the issue may have been something github was doing. Perhaps after triggering a 403 response they would keep sending 403 responses back for awhile which would make the code look like there was a bottleneck. But when I hit the github API in another shell while the download step was in the bottleneck, I didn't get any 403 responses so this probably is not the reason why the bottleneck is occurring.

I ran the old code and new code again today and verified the bottleneck still happens for the old code.

PR #432, opened 2017-06-27
Contributing Info

MichaelViveros:
@heathermiller @MasseGuillaume @julienrf This is my WIP PR (that Heather recommended I add) to track my progress on the new feature I'm working on to add contributing info for a project in scaladex like beginner-friendly issues, contributing guide and chatroom. I will also use this PR to get help with any coding questions I have.

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
@@ -14,6 +14,7 @@ import play.api.libs.json._
 
 import scala.concurrent.{Await, Future}
 import scala.concurrent.duration.Duration
+import scala.util.{Try, Success, Failure}
 
 import org.slf4j.LoggerFactory
 
@@ -67,7 +68,6 @@ trait PlayWsDownloader {
     * @param toDownload the set of downloadable elements
     * @param downloadUrl a function to get the WsRequest for the current element
     * @param process a function to process the response in succes case
-    * @param graphqlQuery query sent to Github's GraphQL API
     * @tparam T Input type
     * @tparam R output type
     */
@@ -76,64 +76,151 @@ trait PlayWsDownloader {
       toDownload: Set[T],
       downloadUrl: (AhcWSClient, T) => WSRequest,
       process: (T, WSResponse) => R,
-      parallelism: Int,
-      graphqlQuery: Option[(T) => JsObject] = None
+      parallelism: Int
   ): Seq[R] = {
 
     val client = wsClient
     val progress = ProgressBar(message, toDownload.size, log)
 
-    def processItem(item: T) = {
-      val request = downloadUrl(client, item)
-      val response = graphqlQuery match {
-        case Some(q) => request.post(q(item))
-        case None    => request.get
-      }
-      response.transform(
-        data => {
-          if (toDownload.size > 1) {
-            progress.step()
-          }
-          process(item, data)
-        },
-        e => {
-          log.warn(
-            s"error on downloading content from ${request.url}: ${e.getMessage}")
-
-          e
-        }
-      )
-    }
-
     def processDownloads = {
 
       Source(toDownload).mapAsyncUnordered(parallelism) { item =>
-        processItem(item)
+        val request = downloadUrl(client, item)
+        val response = request.get
+
+        response.transform(
+          data => {
+            if (toDownload.size > 1) {
+              progress.step()
+            }
+            process(item, data)
+          },
+          e => {
+            log.warn(
+              s"error on downloading content from ${request.url}: ${e.getMessage}")
+
+            e
+          }
+        )
       }
     }
 
     if (toDownload.size > 1) {
       progress.start()
     }
-    try {
-      val response =
-        if (toDownload.size == 1) {
-          Seq(Await.result(processItem(toDownload.head), Duration.Inf))
-        } else {
-          Await.result(processDownloads.runWith(Sink.seq), Duration.Inf)
-        }
-
-      if (toDownload.size > 1) {
-        progress.stop()
-      }
-      client.close()
-      response
-    } catch {
-      case e: Exception =>
-        log.warn("failed to download", e)
-        client.close()
-        Seq()
+    val response = Await.result(processDownloads.runWith(Sink.seq), Duration.Inf)
+    if (toDownload.size > 1) {
+      progress.stop()
     }
+    client.close()
+
+    response
+  }
+  /**
+    * Actual download of bunch of documents from the Github's REST API. Will loop through all and display a status bar in the console output.
+    *
+    * @param message the message for the loader info
+    * @param toDownload the set of downloadable elements
+    * @param downloadUrl a function to get the WsRequest for the current element
+    * @param process a function to process the response in succes case
+    * @tparam T Input type
+    * @tparam R output type
+    */
+  def downloadGithub[T, R](
+      message: String,
+      toDownload: Set[T],
+      downloadUrl: (AhcWSClient, T) => WSRequest,
+      process: (T, WSResponse) => Try[R],
+      parallelism: Int
+  ): Seq[R] = {
+
+    def processItem(client: AhcWSClient, item: T, progress: ProgressBar) = {
+      val request = downloadUrl(client, item)
+      val response = request.get
+
+      response.flatMap { data =>
+        if (toDownload.size > 1) {
+          progress.step()
+        }
+        Future.fromTry(process(item, data))
+      }
+
+    }
+
+    processDownloads(message, toDownload, processItem, parallelism)
   }
 
+  /**
+    * Actual download of bunch of documents from Github's GraphQL API. Will loop through all and display a status bar in the console output.
+    *
+    * @param message the message for the loader info
+    * @param toDownload the set of downloadable elements
+    * @param downloadUrl a function to get the WsRequest for the current element
+    * @param query query sent to Github's GraphQL API
+    * @param process a function to process the response in succes case
+    * @tparam T Input type
+    * @tparam R output type
+    */
+  def downloadGraphql[T, R](
+     message: String,
+     toDownload: Set[T],
+     downloadUrl: AhcWSClient => WSRequest,
+     query: T => JsObject,
+     process: (T, WSResponse) => Try[R],
+     parallelism: Int
+  ): Seq[R] = {
+
+    def processItem(client: AhcWSClient, item: T, progress: ProgressBar) = {
+      val request = downloadUrl(client)
+      val response = request.post(query(item))
+      response.flatMap { data =>
+        if (toDownload.size > 1) {
+          progress.step()
+        }
+        Future.fromTry(process(item, data))
+      }
+    }
+
+    processDownloads(message, toDownload, processItem, parallelism)
+  }
+
+  private def processDownloads[T, R](
+      message: String,
+      toDownload: Set[T],
+      processItem: (AhcWSClient, T, ProgressBar) => Future[R],
+      parallelism: Int
+  ):Seq[R] = {
+
+    def processItems(client: AhcWSClient, progress: ProgressBar) = {
+
+      Source(toDownload).mapAsyncUnordered(parallelism) { item =>
+        processItem(client, item, progress)
+      }
+    }
+
+    val client = wsClient
+    val progress = ProgressBar(message, toDownload.size, log)
+
+    if (toDownload.size > 1) {
+      progress.start()
+    }
+
+    val result = Await.ready(processItems(client, progress).runWith(Sink.seq), Duration.Inf)
+
+    if (toDownload.size > 1) {
+      progress.stop()
+    }
+    // pause for 1s before closing client so other threads that were trying to download
+    // don't get interrupted and throw p.s.a.i.n.u.c.D.rejectedExecution if download stopped due to error
+    Thread.sleep(1000.toLong)
+    client.close()
+
+    result.value.map(_ match {
+      case Success(value) => value
+      case Failure(e) => {
+        log.warn(s"ERROR - $e")
+        Seq()
+      }
+    }).getOrElse(Seq())
+  }
 }
--- a/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
@@ -6,6 +6,7 @@ import download.PlayWsDownloader
 import cleanup.GithubRepoExtractor
 import maven.PomsReader
 import model.misc.GithubRepo
+import model.Project
 
 import org.json4s._
 import org.joda.time.DateTime
@@ -98,6 +99,19 @@ class GithubDownload(paths: DataPaths,
   }
 
   /**
+    * Apply github authentication strategy and set response header to html
+    *
+    * @param request the current request
+    * @return
+    */
+  def applyCommunityProfileHeaders(request: WSRequest): WSRequest = {
+
+    applyBasicHeaders(
+      request.addHttpHeaders(
+        "Accept" -> "application/vnd.github.black-panther-preview+json"))
+  }
+
+  /**
     * Save the json response to directory
     *
     * @param filePath the file path to save the file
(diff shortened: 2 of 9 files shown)


data/src/main/scala/ch.epfl.scala.index.data/github/GithubModel.scala
```diff
@@ -78,13 +84,24 @@ case class GraphqlRepositoryTopic(
     url: String = ""
 )
 
+case class GraphqlIssue(
+    number: Int = 0,
+    title: String = "",
+    bodyText: String = "",
+    url: String = ""
+)
```
julienrf: Why not use `Option[String]` if these fields can be empty?

julienrf: Also, why is it useful to have a default value for the issue number?

data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
```diff
@@ -15,9 +20,25 @@ case class ProjectForm(
...
-    documentationLinks: List[(String, String)]
+    documentationLinks: List[(String, String)],
+    beginnerIssuesLabel: Option[String],
+    chatroom: Option[Url]
 ) {
-  def update(project: Project): Project = {
+  def update(project: Project, paths: DataPaths): Project = {
+    val oldBeginnerIssueLabel = project.github.map(_.beginnerIssuesLabel).getOrElse(None)
+    val beginnerIssues = if (beginnerIssuesLabel != oldBeginnerIssueLabel) {
+      implicit val system = ActorSystem()
```
julienrf: You should probably reuse an existing actor system instead of creating your own.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
```diff
@@ -387,16 +469,51 @@ class GithubDownload(paths: DataPaths,
...
+
+  /**
     * process all downloads
     */
   def run(): Unit = {
 
     download[GithubRepo, Unit]("Downloading Topics",
                                githubRepos,
-                               githubGraphqlUrl,
+                               null,
```
julienrf: I’m worried by all these nulls, can we use `Option` instead?

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
```diff
@@ -246,49 +257,109 @@ class GithubDownload(paths: DataPaths,
...
+      project: Project,
+      response: WSResponse): Unit = {
+
+    val errorOccurred = checkGraphqlError[Project](project, response, isRetry, issuesQuery, processIssuesResponse)
+
+    if (!errorOccurred) {
+      saveJson(githubRepoIssuesPath(paths, project.githubRepo), project.githubRepo, response.body)
+    }
+
+    ()
```
julienrf: This pattern looks weird. Is it possible to design things such that if there is an error you have no result?

MichaelViveros: Not sure what you mean by "no result". Do you not like `checkGraphqlError` returning `true` if there's an error (would prefer it to be an option)? Or do you not like how `checkGraphqlError` handles the error (throws an exception in once case, retries the download in another case)?

julienrf: Actually it’s not really important, just forget about this one :)

MichaelViveros: Alright. But if there's a better way to handle errors in Scala, I'd like to know (still pretty new to Scala)

julienrf: Idiomatic ways to handle errors are to `Option`, `Either` or `Try`.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubModel.scala
MichaelViveros: [Done](https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-8ef04958e1f190b2b57d7258efa1f307)

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
MichaelViveros: [Done](https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-e03c541cf1bd7ec0322a9a6571160beb), updated it to use `Try`

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
MichaelViveros: [Done] (https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-9b0b9f250047a450506bda2259c11537), added 2 download methods, 1 for normal downloads and 1 for graphql downloads

MichaelViveros:
@julienrf @MasseGuillaume I need some help. I want to hit the Github API to get beginner-friendly issues once a maintainer sets the label they use to mark beginner-friendly issues (Ex. low-hanging-fruit) in the edit project page. Previously, scaladex was only hitting the Github API (`GithubDownload`) when populating the index in the data module. But now the server module needs to be able to hit the Github API. The server module will also need to hit the Github API for the live update feature I looked into earlier which updates a project's github info and topics whenever it's page is loaded.

Initially, I was able to hit the Github API in the server module by creating an `ActorSystem` which was used by a `GithubDownload` object in `ProjectForm.update` so that beginner issues could be fetched if the beginner issues label was set in the edit project page. Julien [noticed](https://github.com/scalacenter/scaladex/pull/432/commits/80698c5d83c68f41e37f00227f00c3027a7e940d#r124469656) that I was creating a new `ActorSystem` but I should be using an existing one so I refactored the code to pass in a `GithubDownload` object to the `update` method. This `GithubDownload` object gets defined when the actor system gets created in either [`Main.scala`](https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-0264debba95bfef61ce4f8b684f20530) in the data module or [`Server.scala`](https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-b781b584db67cf2b83817b3b52cf58f5) in the server module so it re-uses those actor systems. This is a bit messy though because it introduced some boiler plate code to pass this `GithubDownload` object all the way down whenever there's a call to `ProjectForm.update`.

When the server module hits the Github API, it will do it on a per-project basis (Ex. when a project's beginner issues label is set, when a project page is loaded). It won't be hitting it for a bunch of projects at one time like in the data module. Using `akka` and `ActorSystems` to just make one request to the Github API at a time seems like overkill. It might be better to make a new `GithubDownload` class which doesn't use actors and can then be instantiated directly whenever it's needed in the server module (won't need an `ActorSystem` anymore).

So what's the best way to be able to hit the Github API from the server module as well as the data module?

data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
MichaelViveros: [Done](https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-9b0b9f250047a450506bda2259c11537), passed in `GithubDownload` to calls to `update`. Also see my [comment](https://github.com/scalacenter/scaladex/pull/432#issuecomment-311820363) on the PR

julienrf:
> So what's the best way to be able to hit the Github API from the server module as well as the data module?

What you did looks perfect, I don’t see any problem in passing the `GithubDownload` all the way down.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
```diff
@@ -246,49 +257,98 @@ class GithubDownload(paths: DataPaths,
...
+    * @param response the response
+    * @return
+    */
   private def processTopicsResponse(isRetry: Boolean)(
       repo: GithubRepo,
-      response: WSResponse): Unit = {
+      response: WSResponse): Try[Unit] = Try {
+
+      checkGraphqlError[GithubRepo](repo, response, isRetry, topicQuery, processTopicsResponse)
+      saveJson(githubRepoTopicsPath(paths, repo), repo, response.body)
```
julienrf: Did you test what happens in case of error?

I’m afraid I introduce more confusion :(

Now, `checkGraphqlError` doesn’t throw an exception anymore, but it _returns_ the exception in case of failure. If you ignore the result of `checkGraphqlError` then you omit checking whether there was an error or not.

You should have a look at the `map` method of `Try`.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubReader.scala
```diff
@@ -99,8 +103,50 @@ object GithubReader {
     val repoTopicsPath = githubRepoTopicsPath(paths, github)
     val graphqlResult = read[GraphqlResult](
       Files.readAllLines(repoTopicsPath).toArray.mkString(""))
-    graphqlResult.data.repository.repositoryTopics.nodes.map { node =>
-      node.topic.name
+    graphqlResult.data.get.repository.get.repositoryTopics.get.nodes.get.map { node =>
+      node.topic.get.name.get
```
julienrf: Did you check what happens if those fields are empty?

In general, calling `.get` on `Option`s doesn’t help because it tends to make you forget about handling all the states of your application. Try to implement something without any `.get` call.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
MichaelViveros: Haven't tested it yet since I'm also refactoring how the code handles errors from the GraphQL API as part of another issue.

Are you sure `checkGraphqlError` doesn't throw an exception anymore? I didn't change that method. I just changed the methods that call it (`processIssuesResponse` and `processTopicsResponse`) to return a `Try[Unit]`. So then when `checkGraphqlError` throws its error, it gets returned as a `Failure` by the process methods which gets [handled](https://github.com/scalacenter/scaladex/pull/432/commits/0d0e08ce6dbd290c746345aa8c56493a424dcd08#diff-9b0b9f250047a450506bda2259c11537R129) in `downloadGraphql`

data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
```diff
@@ -107,10 +94,69 @@ trait PlayWsDownloader {
...
+    */
+  @throws(classOf[Exception])
+  def downloadGraphql[T, R](
+      message: String,
+      toDownload: Set[T],
+      downloadUrl: AhcWSClient => WSRequest,
+      query: T => JsObject,
+      process: (T, WSResponse) => Try[R],
+      parallelism: Int
+  ): Seq[R] = {
```
MasseGuillaume: This Should return Try[Seq[R]]

data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
```diff
@@ -107,10 +94,69 @@ trait PlayWsDownloader {
...
+      val request = downloadUrl(client)
+      val response = request.post(query(item))
+      response.transform(
+        data => {
+          if (toDownload.size > 1) {
+            progressBar.step()
+          }
+          process(item, data) match {
+            case Success(result) => result
+            case Failure(ex) => throw ex
```
MasseGuillaume: this defeat the purpose of using a Try.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
julienrf: Indeed I’m wrong, the `checkGraphqlError` still throws an exception. You may want to make it return a `Try[Unit]` instead.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubReader.scala
MichaelViveros: Will do. Would replacing `get` with `flatMap` be better?

Ex. `graphqlResult.data.flatMap(_.repository.flatMap(_.repositoryTopics.flatMap(...`

julienrf: Exactly!

MasseGuillaume: or with a for comp.

data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
MichaelViveros: True. The whole reason I'm throwing the exception in the first place is so that when a thread [hits](https://github.com/scalacenter/scaladex/pull/432/files#diff-e03c541cf1bd7ec0322a9a6571160bebR325) the API rate limit, it will throw the error which will get caught by the parent thread in `PlayWsDownloader` and will then stop all the other threads from executing (since when the rate limit has been reached, those other threads will get errors when they hit the API so no point in them executing anymore).

Is there a better way to do this @julienrf? Maybe sending messages between actors ... I tried looking into that but couldn't figure out anything. Or we could pause all the threads until the API rate limit resets (maximum time of 1 hour).

julienrf: You should propagate `Try` values instead of throwing exceptions.

MichaelViveros:
@MasseGuillaume @julienrf 

Got rid of `.get` and added for comp in [`GithubReader`](https://github.com/scalacenter/scaladex/pull/432/commits/c4eafc31e053839daaec84fc85bab938f4c8de7a#diff-101331e07214cf2d6fb80a630b7f5926)

Got rid of code that threw exception and added code to propogate `Try` values but it got messy.
* Had to unwrap the `Try` value that gets returned for each item in [`processDownloads`](https://github.com/scalacenter/scaladex/pull/432/commits/c4eafc31e053839daaec84fc85bab938f4c8de7a#diff-9b0b9f250047a450506bda2259c11537R149) so that I could return a failed `Future` if the code that processed the response of the download returned a `Failure`. With the old code, a successful `Future` would be returned if the code that processed the response returned a `Failure` and the parent thread wouldn't stop all the other threads from processing downloads.
* Had to get rid of call to [`Await.result`](https://github.com/scalacenter/scaladex/pull/432/commits/c4eafc31e053839daaec84fc85bab938f4c8de7a#diff-9b0b9f250047a450506bda2259c11537L167) since it would still throw an exception if the result of one of the `Futures` from `processDownloads` failed. Replaced it with call to `Await.ready` which won't throw the exception but now `downloadGraphql` returns a `Unit` instead of a `Seq[R]` or `Seq[Try[R]]`.

Is there any cleaner way to propagate `Try` values so that the code will stop processing all downloads if processing one of them returned a `Failure`?

julienrf:
Hi @MichaelViveros,

Thanks for your efforts :) Dealing with errors is a hard thing.

> Is there any cleaner way to propagate Try values so that the code will stop processing all downloads if processing one of them returned a Failure?

That sounds like `flatMap` :)

For instance, consider your following code:

~~~ scala
val response = request.post(query(item))
response.transformWith {
  case Success(data) => {
    if (toDownload.size > 1) {
      progress.step()
    }
    process(item, data) match {
      case Success(result) => Future.successful(result)
      case Failure(e) => Future.failed(e)
    }
  }
  case Failure(e) => Future.failed(e)
}
~~~

You can take advantage of `flatMap` to rewrite it as follows:

~~~ scala
val response = request.post(query(item))
response.flatMap { data =>
  if (toDownload.size > 1) {
    progress.step()
  }
  process(item, data)
}
~~~

Eventually, you want to convert your `Try` into a `Future`, like you did. I suggest you to implement a helper function like so:

~~~ scala
def tryToFuture[A](tryA: Try[A]): Future[A] = …
~~~

Does that help?

MichaelViveros:
Definitely. Added `flatMap` and call to `Future.fromTry` in latest [commit](https://github.com/scalacenter/scaladex/pull/432/commits/dba2762e1ad6d35a8bd24e1e2239c8771783b27f).

And thanks for taking the time to give me all this feedback @julienrf, I really appreciate it and I'm learning lots.

MasseGuillaume:
@MichaelViveros is this good to go? I will do the rebase.

MichaelViveros:
Not yet, most of the back-end is done but there's still some minor issues. And none of the front-end of showing the contributing info in the UI is done yet.

MasseGuillaume:
@MichaelViveros okay. Can you get the backend part done first? A smaller PR is easier to merge and easier to review.

MichaelViveros:
Will do.

MichaelViveros:
@julienrf @MasseGuillaume Just rebased and squashed my commits. This PR is now up to date with master. The only changes since you guys last checked it are that I refactored `GithubDownload`/`PlayWsDownloader` to handle errors from the REST API in the same way it handled them from the GraphQL API (passing around a `Try` object) and I removed the code that allowed maintainers to set a project's `beginnerIssuesLabel` and `chatroom` link in the edit project page since I wanted this PR to just be for back-end changes (my next PR will be for front-end).

This PR is good to go once you guys have checked it.

MasseGuillaume:
Awesome, I will take a look today.

MasseGuillaume:
Well done it improved a lot with Julien's feedback. I added my review here: https://github.com/scalacenter/scaladex/pull/448/commits/5141b6e702efc22049ec6135a74d03cb9cab39b3

If I made any mistake feel free to open another PR.

Closing this PR and merging this one https://github.com/scalacenter/scaladex/pull/448 with your authorship.

PR #465, opened 2017-07-18
Change the browser-tab title

jducoeur:
Currently, the title in the browser tab for the Scaladex homepage says "Home", which is unhelpful, especially when folks are bookmarking.  Changing this to "Scaladex".


MasseGuillaume:
Ops forgot to merge. Thanks!

PR #467, opened 2017-07-24
Contributing Info UI

MichaelViveros:
@heathermiller @MasseGuillaume @julienrf 

New features:
* Front Page - added contributing info section, displays contributing info panels for randomly selected projects that have contributing info (beginner issues, chatroom, contributing guide)
* Contributing Search - added search param to search page that filters projects that have contributing info displays contributing info in the search results, http://localhost:8080/search?q=*&contributingSearch=true
* Edit Project Page - added fields for beginner issues label, chatroom link, contributing guide link. When the beginner issues label is set, issues with that label are fetched and stored for the project as beginner issues. Alternatively, you can set the beginner issues label in `projects.json` and when the data is re-indexed, it will fetch and store issues with that label for the project
* Project Page - added contributing info panel on the right which only shows for projects that have contributing info

TODO:
* let maintainer set which specific issues are displayed for their project after they set beginner issues label in the edit project page

I tried pushing this code to the dev server but got some memory errors when trying to run it, I'll let you know when those are resolved so you can see things for yourself.

Front Page:

frontpage

Search Page:

searchpage

Project Page:

projectpage

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
@@ -2,8 +2,9 @@ package ch.epfl.scala.index
 package data
 package project
 
+import data.github.{GithubDownload, GithubReader}
 import model.Project
-import model.misc.GithubInfo
+import model.misc.{GithubInfo, Url, GithubIssue}
 
 case class ProjectForm(
     // project
@@ -18,9 +19,18 @@ case class ProjectForm(
     // documentation
     customScalaDoc: Option[String],
     documentationLinks: List[(String, String)],
-    primaryTopic: Option[String] = None
+    primaryTopic: Option[String] = None,
+    beginnerIssuesLabel: Option[String],
+    beginnerIssues: List[GithubIssue] = List(),
+    selectedBeginnerIssues: List[GithubIssue] = List(),
+    chatroom: Option[Url],
+    contributingGuide: Option[Url],
+    codeOfConduct: Option[Url]
 ) {
-  def update(project: Project): Project = {
+  def update(project: Project,
+             paths: DataPaths,
+             githubDownload: GithubDownload,
+             fromStored: Boolean = false): Project = {
 
     val githubWithKeywords =
       if (project.github.isEmpty) {
@@ -30,6 +40,21 @@ case class ProjectForm(
           github => github.copy(topics = github.topics ++ keywords)
         )
       }
+
+    val oldBeginnerIssueLabel = project.github.flatMap(_.beginnerIssuesLabel)
+    val getBeginnerIssues = fromStored && beginnerIssues.isEmpty && beginnerIssuesLabel.isDefined
+    val newBeginnerIssues =
+      if (getBeginnerIssues) {
+        githubDownload.runBeginnerIssues(project.githubRepo,
+                                         beginnerIssuesLabel.getOrElse(""))
+        GithubReader
+          .beginnerIssues(paths, project.githubRepo)
+          .getOrElse(List())
+      } else beginnerIssues
+
+    val newChatroom = chatroom.filterNot(_.target == "")
+    val newContributingGuide = contributingGuide.filterNot(_.target == "")
+    val newCodeOfConduct = codeOfConduct.filterNot(_.target == "")
 
     project.copy(
       contributorsWanted = contributorsWanted,
@@ -42,7 +67,29 @@ case class ProjectForm(
       artifactDeprecations = artifactDeprecations,
       cliArtifacts = cliArtifacts,
       hasCli = cliArtifacts.nonEmpty,
-      github = githubWithKeywords,
+      github = githubWithKeywords.map(
+        github =>
+          github.copy(
+            beginnerIssuesLabel = beginnerIssuesLabel.filterNot(_ == ""),
+            beginnerIssues = newBeginnerIssues,
+            selectedBeginnerIssues = selectedBeginnerIssues,
+            // default to project's chatroom/contributingGuide/codeOfConduct
+            // if updating from stored project and stored project didn't override
+            // that value
+            chatroom =
+              if (fromStored && !newChatroom.isDefined)
+                project.github.flatMap(_.chatroom)
+              else newChatroom,
+            contributingGuide =
+              if (fromStored && !newContributingGuide.isDefined)
+                project.github.flatMap(_.contributingGuide)
+              else newContributingGuide,
+            codeOfConduct =
+              if (fromStored && !newCodeOfConduct.isDefined)
+                project.github.flatMap(_.codeOfConduct)
+              else newCodeOfConduct
+        )
+      ),
       customScalaDoc = customScalaDoc.filterNot(_ == ""),
       documentationLinks = documentationLinks.filterNot {
         case (label, link) => label == "" || link == ""
@@ -67,7 +114,13 @@ object ProjectForm {
       cliArtifacts,
       customScalaDoc,
       documentationLinks,
-      primaryTopic
+      primaryTopic,
+      github.flatMap(_.beginnerIssuesLabel),
+      github.map(_.beginnerIssues).getOrElse(List()),
+      github.map(_.selectedBeginnerIssues).getOrElse(List()),
+      github.flatMap(_.chatroom),
+      github.flatMap(_.contributingGuide),
+      github.flatMap(_.codeOfConduct)
     )
   }
 }
--- a/server/src/main/assets/css/partials/_main.scss
+++ b/server/src/main/assets/css/partials/_main.scss
@@ -172,6 +172,42 @@ main {
             @include box-shadow(0 1px 15px rgba(0, 0, 0, 0.15));
         }
     }
+
+    .contributing-projects {
+        @extend .recent-projects;
+        h2 {
+            margin-bottom: 0;
+        }
+        h4 {
+            margin-bottom: ($line-height-computed * 1.5);
+        }
+        .content-contributing-project {
+            @extend .content-project;
+            height: 350px;
+            h4 {
+                margin: 0;
+                display: inline-block;
+                padding-left: 15px;
+                width: 90%;
+            }
+            img {
+                width: 28px;
+                height: auto;
+                float: left;
+            }
+            a {
+                display: block;
+                padding-bottom: 10px;
+                color: $gray;
+                &:hover,
+                &:focus {
+                    color: $brand-primary;
+                    text-decoration: underline;
+                }
+            }
+            td { padding-left: 20px; }
+        }
+    }
 }
 
 // Result page
@@ -297,6 +333,14 @@ main {
                 color: $brand-primary;
             }
         }
+    }
+    .contributing-result {
+        a {
+            display: block;
+            padding-bottom: 10px;
+        }
+        td { padding-left: 80px; }
+        p { padding-bottom: 20px; }
     }
 }
 
@@ -444,6 +488,19 @@ main {
                 position: relative;
                 margin-left: 5px;
             }
+        }
+        .project-contributing-info {
+            a {
+                display: block;
+                padding-bottom: 10px;
+                color: $gray;
+                &:hover,
+                &:focus {
+                    color: $brand-primary;
+                    text-decoration: underline;
+                }
+            }
+            td { padding-left: 30px; }
         }
     }
     .collaborators {
@@ -601,3 +658,5 @@ html, body {
     }
   }
 }
+
+.white { color: #fff; }
--- a/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
@@ -6,6 +6,7 @@ import model.misc._
 
 import data.DataPaths
 import data.project.ProjectForm
+import data.github.GithubDownload
 
 import release._
 import misc.Pagination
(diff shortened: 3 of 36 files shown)


MasseGuillaume:
Awesome!

Some suggestions:

* I would also add a button to the filter to jump between contributor search / normal search.
* It would also be nice to add a call to action (big button) above the fold (without scrolling) on the front page.

data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
```diff
@@ -30,6 +37,18 @@ case class ProjectForm(
         )
       }
 
+    val oldBeginnerIssueLabel = project.github.flatMap(_.beginnerIssuesLabel)
```
MasseGuillaume: This should be in the Github Step.

server/src/main/assets/css/partials/_main.scss
```diff
@@ -172,6 +172,42 @@ main {
             @include box-shadow(0 1px 15px rgba(0, 0, 0, 0.15));
         }
     }
+
+    .contributing-projects {
```
MasseGuillaume: Can you add some screenshot with a smaller window?

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -24,14 +25,17 @@ import org.slf4j.LoggerFactory
...
  * @param github  Github client
  * @param paths   Paths to the files storing the index
  */
-class DataRepository(github: Github, paths: DataPaths)(
-    private implicit val ec: ExecutionContext
-) {
+class DataRepository(
+    github: Github,
+    paths: DataPaths,
+    githubDownload: GithubDownload
```
MasseGuillaume: ibid as the previous comment. Github downloads should happen when indexing.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -478,6 +500,20 @@ class DataRepository(github: Github, paths: DataPaths)(
       .map(_.totalHits)
   }
 
+  def contributingProjects() = {
+    esClient
+      .execute {
+        search(indexName / projectsCollection)
+          .query(
+            functionScoreQuery(contributingQuery)
+              .scorers(randomScore(scala.util.Random.nextInt(10000)))
```
MasseGuillaume: would be good to double check if this means `a + b + c + random` where a,b, c

```
existsQuery("github.beginnerIssues")
existsQuery("github.contributingGuide")
existsQuery("github.chatroom")
```

are a, b, c values between 0 and 1? This would make 1 + 1 + 1 + large random value.

template/src/main/twirl/ch.epfl.scala.index.views/contributingInfo.scala.html
```diff
@@ -0,0 +1,36 @@
...
+      }
+    }
+}
+@for(github <- project.github) {
+  @for(issue <- github.beginnerIssues.take(3)) {
+    <a href="@issue.url.target" target="_blank">@issue</a>
+  }
+
+   @for(contributingGuide <- github.contributingGuide; chatroom <- github.chatroom) {
+        <table>
```
MasseGuillaume: try to avoid tables, they are harder to style.

MichaelViveros:
@MasseGuillaume Thanks

> I would also add a button to the filter to jump between contributor search / normal search

Where should this button go? To the left of the "Sorting" label? Alternatively, I could add a link to the normal search in the description for "Contributing Search" on the contributing search page (like how I had link to the Contributing Search on the front page).

> It would also be nice to add a call to action (big button) above the fold (without scrolling) on the front page

What do you mean by "above the fold"? Where the Scaladex and Github Login buttons are? And this button would jump to the "Looking for a project to contribute to?" section of the front page?

heathermiller:
This is **really great**!

On the first pass, @MichaelViveros and I will have to go through the template changes together. For now, let's not worry so much about that. (I will help). 

But right now in review, let's focus on changes to everything else other than the views. Generally everything looks good to me. I have questions about a few refactorings...

template/src/main/twirl/ch.epfl.scala.index.views/contributingInfo.scala.html
heathermiller: @MichaelViveros and I will deal with that together 🙂

heathermiller:
> Where should this button go?

Why not have small text beneath the "All the projects below..." text that says "go to normal search" or something?

heathermiller:
> It would also be nice to add a call to action (big button) above the fold (without scrolling) on the front page

Good idea. Let's think about how we can do this so it's not too distracting.

MasseGuillaume:
> Where should this button go?

I would say on the left because it's a filter. Maybe above the filter label?

> What do you mean by "above the fold"? Where the Scaladex and Github Login buttons are? And this button would jump to the "Looking for a project to contribute to?" section of the front page?

Above the fold is the part of the screen when you load a page that you don't have to scroll. On my screen it's 900px. It could be on top of the keyword list for example.

data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
MichaelViveros: I put it there so that when the user sets the beginner issues label in the edit project page, this code will be run so that it can grab the beginner issues right away. 

If this code was in the Guthub step, the beginner issues wouldn't get downloaded until the next time scaladex is re-indexed so the user would have to wait until then for their project to show up in the front page panel, contributing search results, ...

The next thing I want to add is to get the beginner issues in the client when the user sets the beginner issues label in the edit project page (similar to how you added updating the readme in the client), so that the user can then select which specific issues will be displayed for their project. This will get rid of the above code completely.

MichaelViveros:
> Why not have small text beneath the "All the projects below..." text that says "go to normal search" or something?

@heathermiller Will do
 
> It could be on top of the keyword list for example

@MasseGuillaume Got it, thanks for the clarification

data/src/main/scala/ch.epfl.scala.index.data/project/ProjectForm.scala
heathermiller: Yeah, this is better here rather than in indexing.

MasseGuillaume:
Feedback from our meeting:

* [x] Post screenshots on contributors.scala-lang about the upcoming contributing feature or Scaladex (https://contributors.scala-lang.org/t/feedback-request-connecting-contributors-with-projects-in-scaladex/988)
* [ ] Manually add more projects to the contributing
* [ ] Boost the issues fields when searching on the contributor's view
* [ ] Pinned issues could come from a GitHub label
* [ ] The UI could redundantly fetch contributor data like it does for the Readme
* [ ] Make sure to set scaladex on a cron job
* [ ] We could create a banner to announce new features
* [ ] Add a button on top of the filters to go back and forth from the contributor view to the search view

MichaelViveros:
@MasseGuillaume 

> Make sure to set scaladex on a cron job

Could you elaborate on this? What will get run (`github` step, `elastic` step, ...), how often, ...

MasseGuillaume:
> Could you elaborate on this? What will get run (github step, elastic step, ...), how often, ...

via the cron command. I need to ask the sysadmin for this

daily

it would run ./index.sh

MichaelViveros:
@MasseGuillaume @heathermiller For the link to all of the beginner-friendly issues, which is better?

1. "Beginner-friendly Issues" link above the list of issues

screen shot 2017-07-26 at 3 44 16 pm

2. "More Issues ..." link below the list of issues

screen shot 2017-07-26 at 3 44 09 pm

heathermiller:
@MichaelViveros Option #2 is better, assuming that the list is populated with at least 3 issues.

MichaelViveros:
> Add a button on top of the filters to go back and forth from the contributor view to the search view

@heathermiller @MasseGuillaume In addition to a button, I was thinking of adding a filter (called Have Contributing Info) so that you could see how many/which projects in the current search have contributing info (Ex. 5 in the screenshot below). 

screen shot 2017-07-27 at 10 26 51 am

So there would be a Contributing Search button above the filters that takes the user to the Contributing Search page (where results will show contributing info) and there is also a Have Contributing Info filter (where results are the same as the normal search and show project description, targets, scala versions). 

Does this sound like a good idea or is just the Contributing Search button good enough (leave out the Have Contributing Info filter)?

heathermiller:
Sure, this is a good idea. I'd not put it at the top though. Probably better at the bottom of the list of possible filters. 

Also, maybe call it something else? E.g., "Contributors wanted" or something? What about also taking into consideration the "contributors wanted" badge as well? Is there some way unify these things?

MichaelViveros:
I think I'll leave out the Have Contributing Info filter actually (the Contributing Search button should be good enough, there's already a lot of filters like Gui mentioned).



> What about also taking into consideration the "contributors wanted" badge as well? Is there some way unify these things?

@heathermiller Yeah, I think having the contributing info/panel and Contributors Wanted badge separate at the moment is confusing. 

Perhaps the `Contributors Wanted` badge could be replaced by something like a `Contributor-Friendly` badge which instead of getting set by a maintainer in the edit project page it would get automatically set if the project has contributing info (beginner-friendly issues, contributing guide and chatroom)? I think it makes the badge more meaningful - with the Contributors Wanted badge all the maintainer had to do was set Contributors Wanted to true in the edit project page but with the Contributor-Friendly badge the maintainer has to make sure their project has beginner-friendly issues, contributing guide, chatroom. But this might make projects that don't have those 3 things look "unfriendly" which we don't want to do. Perhaps there's a better name for the badge than Contributing-Friendly.

MichaelViveros:
Got some of the minor tasks done (updated [checkboxes](https://github.com/scalacenter/scaladex/pull/467#issuecomment-318019464)).

Here's a screenshot of the Contributing Search page with the button to switch between the normal search and contributing search, a link to more beginner friendly issues and link to a project's Code of Conduct (COC). The normal search page has a similar button to go to the Contributing Search page.

screen shot 2017-07-27 at 8 32 04 pm

@MasseGuillaume I fixed #473 for the Contributing Search so that selecting a filter and then typing a search term will narrow down the results. I fixed it by adding a [minimum score](https://github.com/scalacenter/scaladex/pull/467/commits/4ee37a6dc26d2fc97c8b66d8a036acaca253d592#diff-c5de88d14364dfaadbdecdc462d6c7d1R270) to the function score query that is generated when searching elastic search. If you select a filter, the generated function score query contains a must clause for the selected filter along with should clauses for the boosted queries that determine the function score. By default, the minimum score for a function score query is 0 so if you select a filter and then enter something in the query string, the results just have to pass the must clause (for the filter) and can score 0 for the should clause (the query string) resulting in results matching the filter but not the query string. Specifying a minimum score makes it so that the results must pass the must clause and some of the should clause so they get correctly filtered. 

Let me know what you think of this solution.

Ex. Select filter for Js Target, 4 results:

screen shot 2017-07-27 at 8 37 47 pm

Then enter "quill" as query string, 1 result:

screen shot 2017-07-27 at 8 37 51 pm

MasseGuillaume:
Looking Good! Is this PR ready to merge. Anything more we should add?

MichaelViveros:
My bad about not running scalafmt.

It's missing the features below but they're minor and I can make another PR for them.

The first one about adding more projects with contributing info definitely needs to be done before the the server is re-indexed. If you just want to merge this PR (and not deploy it), then go ahead. If you want to deploy it and re-index, I can add some more projects over the next hour.

* [ ] Manually add more projects to the contributing
* [ ] Boost the issues fields when searching on the contributor's view
* [ ] Pinned issues could come from a GitHub label or the user can select them in the edit project page UI (going with the 2nd option for now)
* [ ] The UI could redundantly fetch contributor data like it does for the Readme
* [ ] We could create a banner to announce new features

MichaelViveros:
Submitted a PR to scaladex-index for adding more repos - https://github.com/scalacenter/scaladex-index/pull/1

This should be good to go once that's resolved

MasseGuillaume:
This should not fail: https://travis-ci.org/scalacenter/scaladex/builds/258938217#L864

You only need the GitHub credential to run the GitHub task.

MichaelViveros:
Don't think that error is due to my PR. Yes, my PR added a dependency on `GithubDownload` in the `elastic` step but running `data/run github` gives the same error when checked out on `master`. I think it's a problem with using `run`, maybe something needs to be added to `build.sbt` to get the credential when using `run` like when using [`reStart`](https://github.com/scalacenter/scaladex/blob/master/build.sbt#L43). I normally use `reStart` instead of `run` and it works fine.

> You only need the GitHub credential to run the GitHub task.

No, the `elastic` step needs it because it [gets beginner issues](https://github.com/scalacenter/scaladex/pull/467/files#diff-960fe3280a436e7e9e3f10c043584657R505) when loading projects by seeing if the project has a stored value for `beginnerIssuesLabel` in `projects.json`. This is the only way I can populate some initial projects with contributing info (only other way is to set `beginnerIssuesLabel` in the edit project page but you need to be a maintainer to have access to that).

MichaelViveros:
New feature banner:

screen shot 2017-07-31 at 3 02 57 pm

MichaelViveros:
@heathermiller Action items from call:
* [ ] Fix banner to match scala-lang's banner
* [ ] Fix UI for front page panels
* [ ] Have client fetch most recent beginner issues on page load for a project and remove any selected issues that were closed
* [ ] Get PR for scaladex-index merged - https://github.com/scalacenter/scaladex-index/pull/1
* [ ] Change "CoC" to "Code of Conduct" in panel
* [ ] Briefly look into keeping the random projects on the front page the same per user session

MasseGuillaume:
Rebased and merged to master. Create another PR for further improvements.

PR #470, opened 2017-07-26
Fix rate limit errors with Github API #431

### LINKED ISSUE #431: Github API Errors
@heathermiller @MasseGuillaume Hitting the GraphQL API for getting a project's topics causes some errors that need to be fixed in order to allow live updates to work (refreshing a project's repo info and topics each time the project's page is loaded).

Known Errors:

1. Rate Limit Expired - expected error when [5000 limit](https://developer.github.com/v4/guides/resource-limitations/#rate-limit) is reached after running `data/reStart github` 3 - 4 times in a row
Response: 200, `X-RateLimit-Remaining` header set to 0
[Code:](https://github.com/scalacenter/scaladex/blob/e0eb3ebbe4fe7e0b4c2405a2ccf9c324f633aa0b/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala#L259) throw error to stop downloading topics

2. Abuse Detection Mechanism - unexpected error that happens sometimes when running `data/reStart github`
Response: 403, message body contains something about triggering an abuse detection mechanism
[Code:](https://github.com/scalacenter/scaladex/blob/e0eb3ebbe4fe7e0b4c2405a2ccf9c324f633aa0b/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala#L274) retry request after waiting for awhile

Not sure why 2 occurs. I verified that when the error occurs, the API keys still have API calls remaining so it's different from 1. Think it might be due to the calls being too close together (frequency as opposed to the volume of calls). 

I reached out to github support (Francis) about a month ago and sent multiple follow up emails but they just told me they were still looking into it, maybe someone else will have better luck contacting them.

To check how many API calls a token has remaining:
`curl -H "Authorization: bearer 5e2ddeed0f9c6169d868121330599b8353ab0b55" -X POST -d '{"query": "query { rateLimit { limit cost remaining resetAt } }"}' https://api.github.com/graphql`

One simple solution might be to add more github API keys for scaladex.

MichaelViveros:
@heathermiller @MasseGuillaume

This fixes #431 by updating the code that downloads repo info from Github to avoid getting errors about hitting the Github API rate limit and abuse rate limit (see #431 for more info about those errors).

I created a personal Scaladex Github [account](https://github.com/scaladex-personal) and a machine Scaladex [account](https://github.com/scaladex-machine) and generated 1 Github API token for each account which adheres to Github's terms of service of using at most 2 tokens when hitting their API (1 token from a personal account and 1 token from a machine account). I will commit these tokens to the scaladex-credentials repo once this PR is deployed.

This PR is similar to the solution described in #431:
* Instead of splitting up the step that downloads the Github info into 2 steps (as mentioned in my previous solution), the code uses one step like it did before and if it hits the rate limit, it [pauses](https://github.com/scalacenter/scaladex/pull/470/files#diff-e03c541cf1bd7ec0322a9a6571160bebR379) until the rate limit resets. The progress bar keeps printing the same thing while the thread is paused which is annoying but I couldn't find a way around this (there's an open [issue](https://github.com/ctongfei/progressbar/issues/17) with `ctongfei/progressbar` to be able to pause it)
* The download step hits the Github API [serially](https://github.com/scalacenter/scaladex/pull/470/files#diff-9b0b9f250047a450506bda2259c11537R199)

The whole `data/reStart github` step takes around 1.5 hours now (~50 mins downloading, ~40 mins pausing)

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
@@ -135,8 +135,7 @@ trait PlayWsDownloader {
       message: String,
       toDownload: Set[T],
       downloadUrl: (AhcWSClient, T) => WSRequest,
-      process: (T, WSResponse) => Try[R],
-      parallelism: Int
+      process: (T, WSResponse) => Try[R]
   ): Seq[R] = {
 
     def processItem(client: AhcWSClient, item: T, progress: ProgressBar) = {
@@ -152,7 +151,7 @@ trait PlayWsDownloader {
 
     }
 
-    processDownloads(message, toDownload, processItem, parallelism)
+    processDownloads(message, toDownload, processItem)
   }
 
   /**
@@ -171,8 +170,7 @@ trait PlayWsDownloader {
       toDownload: Set[T],
       downloadUrl: AhcWSClient => WSRequest,
       query: T => JsObject,
-      process: (T, WSResponse) => Try[R],
-      parallelism: Int
+      process: (T, WSResponse) => Try[R]
   ): Seq[R] = {
 
     def processItem(client: AhcWSClient, item: T, progress: ProgressBar) = {
@@ -186,18 +184,19 @@ trait PlayWsDownloader {
       }
     }
 
-    processDownloads(message, toDownload, processItem, parallelism)
+    processDownloads(message, toDownload, processItem)
   }
 
   private def processDownloads[T, R](
       message: String,
       toDownload: Set[T],
-      processItem: (AhcWSClient, T, ProgressBar) => Future[R],
-      parallelism: Int
+      processItem: (AhcWSClient, T, ProgressBar) => Future[R]
   ): Seq[R] = {
 
     def processItems(client: AhcWSClient, progress: ProgressBar) = {
-
+      // use minimal concurrency to avoid abuse rate limit error which is triggered
+      // by making too many calls in a short period of time, see https://github.com/scalacenter/scaladex/issues/431
+      val parallelism = 1
       Source(toDownload).mapAsyncUnordered(parallelism) { item =>
         processItem(client, item, progress)
       }
--- a/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
@@ -21,6 +21,9 @@ import play.api.libs.ws.ahc.AhcWSClient
 import play.api.libs.json._
 
 import scala.util._
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
 import java.nio.charset.StandardCharsets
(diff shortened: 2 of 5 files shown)


MasseGuillaume:
Can you run scalafmt ?
 https://travis-ci.org/scalacenter/scaladex/builds/257859657#L517

data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
```diff
@@ -186,18 +184,19 @@ trait PlayWsDownloader {
...
-      parallelism: Int
+      processItem: (AhcWSClient, T, ProgressBar) => Future[R]
   ): Seq[R] = {
 
     def processItems(client: AhcWSClient, progress: ProgressBar) = {
-
+      // use minimal concurrency to avoid abuse rate limit error which is triggered
+      // by making too many calls in a short period of time, see https://github.com/scalacenter/scaladex/issues/431
+      val parallelism = 1
       Source(toDownload).mapAsyncUnordered(parallelism) { item =>
```
MasseGuillaume: you can use named parameters here: `mapAsyncUnordered(parallelism = 1)`

Can we set it to 4? We make all the effort to have parallel code but set it sequentiail :S. There is probably a sweet spot, What kind of rate limit stas Github give us?

MasseGuillaume:
Can you create a PR against https://github.com/scalacenter/scaladex-credentials to add the tokens and passwords?

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
```diff
@@ -21,12 +21,17 @@ import play.api.libs.ws.ahc.AhcWSClient
...
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path}
 
 import com.typesafe.config.ConfigFactory
 import org.slf4j.LoggerFactory
 
+import scala.concurrent.ExecutionContext.Implicits.global
```
MasseGuillaume: you should no import the global execution context.
import system.dispatcher at line 38.

more info here: http://doc.akka.io/docs/akka/current/scala/dispatchers.html

data/src/main/scala/ch.epfl.scala.index.data/download/PlayWsDownloader.scala
MichaelViveros: [Done](https://github.com/scalacenter/scaladex/pull/470/files#diff-9b0b9f250047a450506bda2259c11537R203)

And Github didn't give any stats about the rate limit and abuse rate limits are triggered since if they shared that info people could use it to work around them. I tested with higher concurrency and got the abuse rate limit errors so I think `parallelism = 4` is good enough

MasseGuillaume:
It looks like this travis failed test is unrelated with this PR: https://travis-ci.org/scalacenter/scaladex/builds/258947737#L2094

Issue #461, opened 2017-07-12
Sorting buttons broken on search results page


heathermiller:
Steps to reproduce:

1. Click a topic on the front page. 
2. On the search results page, select the button to order results by stars.
3. Observe that the original query gets lost. You now see all projects ordered by star count rather than the ones in the topic that you originally selected.

PR #482, opened 2017-08-10
Fix sorting for topics, #461

MichaelViveros:
@heathermiller Fix for #461


MasseGuillaume:
awesome.

if you want to autoclose the ticket with GitHub you need to put a commit message like:

`foo bar buzz, fix #461`

PR #483, opened 2017-08-10
Suggest topics, fix #436

### LINKED ISSUE #436: Suggest using topics
If a user searches “foo” and “foo” is known to be a project topic we should add a message “did you mean topics=foo?”

For instance, when one searches for “[json](https://index.scala-lang.org/search?q=json)” we would add the message “did you mean [topics=json](https://index.scala-lang.org/search?topics=json)”.

MichaelViveros:
+scalafmt

@heathermiller @julienrf How does this look after entering a topic like "json" as a query string in the search box and pressing enter?

screen shot 2017-08-10 at 1 33 31 pm

The "topics=json" link goes to:

screen shot 2017-08-10 at 1 33 37 pm


julienrf:
This is great :)

heathermiller:
This is great @MichaelViveros! Thank you!

PR #493, opened 2017-08-30
Add parallelism when hitting github api

MichaelViveros:
I tried increasing parallelism [earlier](https://github.com/scalacenter/scaladex/pull/470#discussion_r130227859) in #470 for code that hits the GraphQL API from 1 to 4 (so that it runs a bit faster but still avoids the rate limit errors), but the changes didn't go through for some reason so I added them in this PR.

Also fixed a line I forgot to change in #486 so that code that downloads poms from non-github APIs runs with max parallelism.


MasseGuillaume:
Awesome, thank!

PR #498, opened 2017-09-22
Updated sbt-bintray link

laughedelic:
Also added a link to publishing docs


MasseGuillaume:
Merci!

laughedelic:
You already have FAQ in the wiki. How about moving these docs to wiki as well? I think it makes little fixes like this one much easier for users/contributors.

MasseGuillaume:
Yep this would be useful too https://github.com/scalacenter/scaladex/wiki/FAQ

The advantage of the README is that it's the first thing you look at. If the information is not there most of the user would give up at this point.

laughedelic:
I agree about readme (which should mention wiki). I meant moving docs in `doc/user/` to wiki.

PR #499, opened 2017-10-04
Fixes the copy/pastable line for libraries with crossSuffix.

mpollmeier:
See e.g. ammonite (https://index.scala-lang.org/lihaoyi/ammonite/ammonite/1.0.2?target=_2.12.3):
Is currently:     `libraryDependencies += "com.lihaoyi" % "ammonite" % "1.0.2 cross CrossVersion.full"`
Correct would be: `libraryDependencies += "com.lihaoyi" % "ammonite" % "1.0.2" cross CrossVersion.full`

N.b. I wasn't able to compile the project (before or after this change) because I don't have `sass` installed.

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/Release.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/Release.scala
@@ -62,7 +62,7 @@ case class Release(
 
       List(
         Some(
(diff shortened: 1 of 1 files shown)


MasseGuillaume:
Thanks, I'm adding tests in a followup commit

PR #504, opened 2017-10-14
Fix #500 Add a keyboard shortcut to jump to search bar

### LINKED ISSUE #500: Add a keyboard shortcut to jump to search bar

piotrkwiecinski:
I've added ```/``` as shortcut for jump to search bar. 
The same key is used in Jira, Github issues and gmail for search.

### DIFF
--- a/client/src/main/scala/ch.epfl.scala.index.client/Client.scala
+++ b/client/src/main/scala/ch.epfl.scala.index.client/Client.scala
@@ -140,6 +140,18 @@ object Client {
     }
   }
 
+  private def jumpToSearchInput(event: KeyboardEvent): Unit = {
+    if (event.keyCode == 191) {
+      getSearchBox.foreach { searchBox =>
+        val input = searchBox.getInput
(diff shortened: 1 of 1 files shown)


client/src/main/scala/ch.epfl.scala.index.client/Client.scala
```diff
@@ -140,6 +140,18 @@ object Client {
     }
   }
 
+  private def jumpToSearchInput(event: KeyboardEvent): Unit = {
+    if (event.keyCode == 191) {
```
piotrkwiecinski: I haven't found forward slash on the list of codes in KeyCode object.

MasseGuillaume:
Cool good idea!

yurikhan:
@piotrkwiecinski Would you object to a different shortcut key, one that does not conflict with Firefox’s Quick Find? Maybe `Ctrl+/`, or `s` as Github does? See #512 for a user pain description.

PR #505, opened 2017-10-14
Fix #356 Fix page header doesn't work well when logged in

### LINKED ISSUE #356: Responsive layout for page heading does not work well when logged in
When not logged in, the site is maybe a little eager to get you to log-in but otherwise fine:
image

However, when you do log in things don't layout quite right in responsive mode:
image

piotrkwiecinski:
* Log in/out buttons and projects button moved to separate div
* Reworked margins for logo and buttons


MasseGuillaume:
Awesome thanks!

Issue #510, opened 2017-12-04
optimize large projects


MasseGuillaume:
scalatest takes 11s to load:

https://index.scala-lang.org/scalatest/scalatest

https://index.scala-lang.org/api/project?organization=scalatest&repository=scalatest&target=JVM&scalaVersion=2.12.4

MasseGuillaume:
The api is here: https://github.com/scalacenter/scaladex/blob/aa5e09a05a2847d70f9787997ba439842c99046a/server/src/main/scala/ch.epfl.scala.index.server/routes/api/SearchApi.scala#L161-L215

The slow part:
https://github.com/scalacenter/scaladex/blob/aa5e09a05a2847d70f9787997ba439842c99046a/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala#L397-L411

We need to fetch all releases because the logic in DefaultRelease is not easily indexable.

adpi2:
I fixed this issue a few time ago: https://github.com/scalacenter/scaladex/pull/624

PR #511, opened 2017-12-04
halve search time, partial fix for #510

OlegYch:
use jawn for parsing, todo - don't fetch/parse deps if they are not needed and/or remove json4s


MasseGuillaume:
Thanks!

MasseGuillaume:
https://index.scala-lang.org/api/project?organization=scalatest&repository=scalatest&target=JVM&scalaVersion=2.12.4

2x improvement, from 10s to 5s

PR #513, opened 2017-12-16
Peaceful resolution of terrorist attack on Firefox find bar :)

### LINKED ISSUE #512: Please do not hijack the ‘/’ key
In Firefox, the `/` key brings up the find bar (like `Ctrl+F` but an easier key). Scaladex overrides this key to focus the search input field. If the user was in the middle of reading a long package description and wanted to search from that point, he/she will be taken to the top of page, losing his/her reading position.

This is probably the relevant part of the code: [Client.scala#L143-L153](https://github.com/scalacenter/scaladex/blob/master/client/src/main/scala/ch.epfl.scala.index.client/Client.scala#L143-L153)

piotrkwiecinski:
Fix #512 Change search key to "s" to prevent find bar hijack in Firefox

### DIFF
--- a/client/src/main/scala/ch.epfl.scala.index.client/Client.scala
+++ b/client/src/main/scala/ch.epfl.scala.index.client/Client.scala
@@ -141,7 +141,7 @@ object Client {
   }
 
   private def jumpToSearchInput(event: KeyboardEvent): Unit = {
-    if (event.keyCode == 191) {
+    if (event.keyCode == KeyCode.S) {
       getSearchBox.foreach { searchBox =>
         val input = searchBox.getInput
         if (event.srcElement != input) {
--- a/data/src/main/scala/ch.epfl.scala.index.data/elastic/package.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/elastic/package.scala
@@ -37,7 +37,8 @@ trait ProjectProtocol {
     Try(f).transform(s => Success(Right(s)), f => Success(Left(f))).get
   }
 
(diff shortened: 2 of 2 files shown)


piotrkwiecinski:
FYI @yurikhan

MasseGuillaume:
:+1: haha nice.

client/src/main/scala/ch.epfl.scala.index.client/Client.scala
```diff
@@ -141,7 +141,7 @@ object Client {
   }
 
   private def jumpToSearchInput(event: KeyboardEvent): Unit = {
-    if (event.keyCode == 191) {
+    if (event.keyCode == KeyCode.S) {
```
yurikhan: Hm, I realized one thing that I had noticed earlier but now it’s going to become important. Using a character key as a shortcut requires carefully checking several conditions.

* Modifier keys: `!(event.ctrlKey || event.altKey || event.shiftKey || event.metaKey)`. Otherwise it may inadvertently hijack `Ctrl+S` and/or the Hi**s**tory menu.
* That the currently focused element is not one that would want the normal character. I believe a reasonably general check for that is (in Javascript talk) `!(event.target instanceof HTMLInputElement || event.target instanceof HTMLTextAreaElement || event.target.isContentEditable)`, although in specific cases you might get away with just testing that none of the known input fields are in focus.

piotrkwiecinski: You're correct. I'll give it another go tomorrow.

Issue #390, opened 2017-02-15
Spacing weird for scala (jvm) and scala.js versions on search results page


heathermiller:
<img width="382" alt="screen shot 2017-02-15 at 10 29 40 pm" src="https://cloud.githubusercontent.com/assets/687163/22996222/5497320a-f3ce-11e6-83e4-427b16178b0a.png">

The spacing is broken. Seems like there's a tab indenting the first line of version #s, but none for the second line.

PR #516, opened 2018-01-28
Fix spacing for scala (jvm) and scala.js versions on search results page #390 

wkomendera:
Added small changes in paddings and margins.


MasseGuillaume:
Thanks!

Issue #518, opened 2018-02-05
Can't type the letter 's' in search bar


SeZuo:
Copy-paste works, but typing letter s does not.

Every other key seems to work.

Tried using Dvorak and US layout on ch and jp keyboard.

My s-key works by the way, and it's not the same key on the dvorak and us layout.

MasseGuillaume:
Hum it looks like the culprit is https://github.com/scalacenter/scaladex/blob/master/client/src/main/scala/ch.epfl.scala.index.client/Client.scala#L149

if the searchbox is focused it should not prevent the default (typing s in the input box).

wkomendera:
@MasseGuillaume, can I work on this ?

MasseGuillaume:
@wkomendera all yours!

MasseGuillaume:
fixed in https://github.com/scalacenter/scaladex/pull/520, thanks @wkomendera 

deployed

PR #520, opened 2018-02-07
#518 changed checking of the source of the key events.

wkomendera:
Typing the 's' in searchbox didn't work only on my FF.

I think I've got it.  I checked the solution on FF, IE, and Chrome on windows.


MasseGuillaume:
Awesome! Thanks

Issue #515, opened 2017-12-23
README on GitHub project page shows a badge with an older version


plokhotnyuk:
I released v0.2 yesterday here:
https://github.com/plokhotnyuk/jsoniter-scala

If I try get image by following link directly: https://index.scala-lang.org/plokhotnyuk/jsoniter-scala/macros/latest.svg?color=blue then it displays latest version right by redirection to https://img.shields.io/badge/macros-0.2-blue.svg

But home page of the project on GitHub where README.md displays this badge rendering an image for the older version v0.1.

Is it caching or redirect handling problem on GitHub side or I should use https://img.shields.io/badge/macros-0.2-blue.svg in README.md directly?

MasseGuillaume:
It looks like HTTP code 307 might be cached. I think the solution is to add [`cachingProhibited`](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/caching-directives/cachingProhibited.html) to the Bagde api https://github.com/scalacenter/scaladex/blob/master/server/src/main/scala/ch.epfl.scala.index.server/routes/Badges.scala#L57

PR #524, opened 2018-02-14
#523 Do not allow caching of badges

### LINKED ISSUE #523: badges 404
```
Martynas Mickevičius @2m 07:12
@MasseGuillaume there seems to be something broken with scala index badges.
For example, https://index.scala-lang.org/akka/akka-stream-contrib/latest.svg returns a badge only on every second request.
And https://index.scala-lang.org/akka/akka-stream-contrib/akka-stream-contrib/latest.svg does not return a badge anymore.
I noticed that from the many broken badges in various readme's across the github.
Like here: https://github.com/propensive/kaleidoscope
I see that the badge code has not been changed lately. So it might have something to do with the deployment?
```

2m:
`cachingProhibited` filters out requests and leaves only those that explicitly request for no caching. It was an incorrect solution for #515.

Instead we should add a `no-cache` header to the response and inform the client that badges should not be cached.

Fixes #523

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -136,7 +136,6 @@ lazy val server = project
       "ch.megard" %% "akka-http-cors" % "0.2.1",
       "com.softwaremill.akka-http-session" %% "core" % "0.5.3",
       "com.typesafe.akka" %% "akka-http" % akkaHttpVersion,
-      "com.typesafe.akka" %% "akka-http-caching" % akkaHttpVersion,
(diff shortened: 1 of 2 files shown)


MasseGuillaume:
Thanks, this is correctly explained here: https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/caching-directives/cachingProhibited.html

Issue #522, opened 2018-02-13
broken image link on utest page


SethTisue:
at e.g. https://index.scala-lang.org/lihaoyi/utest/utest-runner/0.2.4?target=_2.10

there is a broken image link pointing to https://index.scala-lang.org/lihaoyi/utest/utest-runner/docs/Splash.png which is 404.

MasseGuillaume:
This normally fix urls.
https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/github/GithubReadme.scala#L7

I have to investigate.

SethTisue:
it seems this got fixed at some point

PR #528, opened 2018-03-03
#522 Fixing broken links in downloaded README on client side

wkomendera:
I've made some similar changes in the Client based on the https://github.com/scalacenter/scaladex/blob/master/data/src/main/scala/ch.epfl.scala.index.data/github/GithubReadme.scala#L7


MasseGuillaume:
Thanks :-)

PR #531, opened 2018-03-08
Increase the limitation of max page (fix #316)

### LINKED ISSUE #316: An admin of a repository should have the permission to edit settings for the repository on Scaladex
I am one of the admin of https://github.com/ThoughtWorksInc/DeepLearning.scala, I expect that I can access https://index.scala-lang.org/edit/thoughtworksinc/deeplearning.scala , however, the page always shows a `Forbidden` for me .

Atry:


MasseGuillaume:
The server will timeout if this value is too large.

Atry:
So what's your suggestion to resolve #316 ?

MasseGuillaume:
@Atry 

Create one actor responsible to fetchOrgRepos asynchronously. It would have a state `Map[User, Set[Repo]]`.

When the user login, send `actor ! FetchRepos(user)` to initialize fetching pages. This process should be non-blocking and update the actor state `actor ! AddRepos(user, repos)` (for each pages). When the user wants to edit, send `actor ? HasPermission(repo)`. Make sure the ask timeout < request timeout.

Since fetching repos can take some time. It would make sense to add an intermediate state: `fetchingRepos: Set[User]` and signal it to the actor when everything is done. If a user wants to edit a repo, but we are still fetching (eg. `fetchingRepos.contains(user)`), `HasPermission(repo)` should reply `Fetching`. The ui should notify the user we are still processing it's repos and that he has to try again later.

Atry:
Sounds very complicated

PR #532, opened 2018-03-13
Opensearch

wkomendera:
This will add "Scaladex search" to the browser, which then can be used either with TAB key (like in chrome) or just in search box (like in FF)
image
 

Chrome is using that search with TAB key already  somehow
image


MasseGuillaume:
Wow thanks a lot!

julienrf:
Awesome!

PR #562, opened 2018-11-18
Fix login issue.

rodrigogdea:
This PR should fix the login issue.

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/Github.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/Github.scala
@@ -1,26 +1,20 @@
 package ch.epfl.scala.index
 package server
 
-import model.misc._
-import data.github._
-
-import akka.http.scaladsl._
-import akka.http.scaladsl.model._
-import HttpMethods.POST
-import headers._
-import Uri._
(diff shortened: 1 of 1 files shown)


server/src/main/scala/ch.epfl.scala.index.server/Github.scala
```diff
@@ -122,15 +116,17 @@ class Github(implicit system: ActorSystem, materializer: ActorMaterializer)
...
-              val res = acc ++ convert(repos.nodes)
-              if (repos.pageInfo.hasNextPage || n < 5) {
-                loop(cursorStart = Some(repos.pageInfo.endCursor),
-                     acc = res,
-                     n + 1)
-              } else {
-                Future.successful(res)
+              val json: JValue = data \ "data" \ "viewer" \ "repositories"
+              json match {
+                case JNothing => Future.successful(acc)
```
julienrf: In which case do we get a `JNothing`?

rodrigogdea:
Thanks @julienrf ! Let me know when I can login.

Regards,
Rodrigo.

julienrf:
I’ve deployed your changes, thanks again!

Issue #543, opened 2018-05-29
Add "mill" tab for quick ivy dependency copy/paste


Baccata:
Mill's syntax to set ivy dependencies in a project is very similar to Ammonite's, and the two are from the same author, so maybe they ought to be in the same tab. 

But here's ammonite syntax for magic imports : 

```scala 
import $ivy.`org.http4s::http4s-core:0.18.12`
```

Here's mill's for ivy dependencies : 

```scala
ivy"org.http4s::http4s-core:0.18.12"
```

MasseGuillaume:
Hi @Baccata if you can create a pull request for this, it would be really appreciated. Here is the relevant code: 

https://github.com/scalacenter/scaladex/blob/892b67c214b04e57fc535ff534f6b337ac6bcffa/model/src/main/scala/ch.epfl.scala.index.model/Release.scala#L47-L131

https://github.com/scalacenter/scaladex/blob/892b67c214b04e57fc535ff534f6b337ac6bcffa/template/src/main/twirl/ch.epfl.scala.index.views/project/install.scala.html#L1-L45

Baccata:
Cool, thanks for pointing me in to the right direction, I'll send PR as soon as I find the time

rusucosmin:
PR #565 opened.

PR #565, opened 2018-12-15
Added Mill's dependency string

rusucosmin:


julienrf:
Hey, can you add a screenshot showing the result of your changes?

rusucosmin:
Sure, here it is:

screen shot 2018-12-15 at 11 02 30 am

julienrf:
I think we should also show how to add a resolver, when needed, like we do here for sbt:

~~~ scala
libraryDependencies += "de.heikoseeberger" %% "akka-http-circe" % "1.21.1"
resolvers += Resolver.bintrayRepo("hseeberger", "maven")
~~~

model/src/main/scala/ch.epfl.scala.index.model/Release.scala
```diff
@@ -130,6 +130,22 @@ case class Release(
...
+   */
+  def millInstall = {
+    def addResolver(r: Resolver) =
+      r.url map (url => s"""MavenRepository("${url}")""")
+    val artifactOperator = if (isNonStandardLib) ":" else "::"
+    List(
+      Some(
+        s"""ivy"${maven.groupId}$artifactOperator${reference.artifact}:${reference.version}""""
+      ),
+      resolver flatMap addResolver
```
julienrf: I’m not sure we should generate the “Mill” tab at all, if the `addResolver` method returns `None`.

rusucosmin: Why? In the [issue](**https://github.com/scalacenter/scaladex/issues/543**) @Baccata requested only the dependency. I can add an if guard in that case, but I think that it is more convenient if you can just copy paste the dependency even without a resolver.

julienrf:
Thanks! I’m merging it. We can improve it later, if needed.

PR #566, opened 2018-12-15
Fix the pagination and reuse BintrayClient.remainingPages

gmarti:


data/src/main/scala/ch.epfl.scala.index.data/bintray/BintrayClient.scala
```diff
@@ -99,12 +99,19 @@ class BintrayClient(paths: DataPaths) {
...
+        startPos <- response
+          .header("X-RangeLimit-StartPos")
+          .flatMap(s => Try(s.toInt).toOption)
+        endPos <- response
           .header("X-RangeLimit-EndPos")
           .flatMap(s => Try(s.toInt).toOption)
-        if count < total
-        remainingPages = (total - 1) / count
-      } yield Seq.tabulate(remainingPages)(page => (page + 1) * count)
+        if endPos < (total - 1)
```
gmarti: last Element of 100 items is 99

data/src/main/scala/ch.epfl.scala.index.data/bintray/BintrayClient.scala
```diff
@@ -99,12 +99,19 @@ class BintrayClient(paths: DataPaths) {
...
+        nextPos = endPos + 1
+        perPage = nextPos - startPos
+        remainingPageCount = Math
```
gmarti: total = 51
nextPos = 50
We need to get the ceiled integer division

julienrf:
It seems that there is a formatting issue: https://travis-ci.org/scalacenter/scaladex/builds/468427416#L518

Would you have time to fix it?

gmarti:
Yes, I formatted test manually because it was ugly, removed the commit
Do you want to rename all page variable to something more meaningful (since it's more a startIndex) ?

Issue #389, opened 2017-02-15
Improve search relevance


heathermiller:
I can't find spark anymore... I have to search for Spark and then fiddle with the filters. Spark should really turn up on the first page when searching for "spark"

julienrf:
The score is highly dependent on keywords, I think that's the best way to avoid false positives, even though it requires manual editing of projects. 

The other results for the "spark" query are not irrelevant. It is super hard to design a magic scoring function that always put your expected result first... 

I added the "spark" keyword to the apache/spark project, in order to boost its score.

kzys:
Can we just use the number of stars as the default sort order?
https://index.scala-lang.org/search?q=spark&page=1&sort=stars

julienrf:
@kzys We tried that but then some projects that are not completely related to a given search query but have a lot of stars (e.g. playframework) appear at the top of the results. That does not work well either…

MasseGuillaume:
@heathermiller let's use this issue: https://github.com/scalacenter/scaladex/issues/303 and close this one as duplicate ?

cvogt:
this looks off
![image](https://cloud.githubusercontent.com/assets/274947/23582608/1df76388-00fc-11e7-95eb-4ee28f295f13.png)

cvogt:
scalafmt doesn't show anywhere in this list eventhough I typed the exact name

cvogt:
when I type enter it does show on position 8 of that list.

maybe project that have the exact search term contained in there name should show first?

olafurpg:
Scalafmt does show up when searching for scalameta

<img width="860" alt="screen shot 2017-03-29 at 14 18 07" src="https://cloud.githubusercontent.com/assets/1408093/24454070/a31c7aac-148a-11e7-9a3a-ec29d55c4175.png">

I think the scalameta repo should be the top result, however.

ShaneDelmore:
Can we give a heavy weight to matches in the name?  If I search for shapeless I should get shapeless, even if a library named ShapetyShapeShape mentions Shapeless 42 times in the read even and has more stars than Travis Brown has StackOverflow points. Is this controversial?

MasseGuillaume:
@ShaneDelmore The counter example is a library named json. Is it more relevant than circe ?

ShaneDelmore:
Yes. If I search for a library named json I would like it to show up before Circe. If I list the category json by popularity then I would expect Circe to be near the top. Has the counter example proved a larger problem than the current results ranking?

[...]

PR #571, opened 2019-05-11
Fix search relevance

eed3si9n:
Please see https://github.com/scalacenter/scaladex/issues/389#issuecomment-490584056

From looking at GitHub's search result, which is very good, the signal we should be boosting more is the description. The tag seems to mix in more random things, so it's ok as a tie breaker, but it's not very reliable.

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
@@ -91,8 +91,13 @@ class DataRepository(
               current = clamp(page),
               totalPages = Math.ceil(r.totalHits / params.total.toDouble).toInt,
               total = r.totalHits
-            ),
-            r.to[Project].toList.map(hideId)
+            ), {
+              // println(r.hits.toList.take(100) map { h =>
+              //   val source = h.sourceAsMap
+              //   s"""${h.score}: ${source("organization")}/${source("repository")}"""
+              // })
+              r.to[Project].toList.map(hideId)
+            }
         )
       )
   }
@@ -253,21 +258,17 @@ class DataRepository(
         )
         .should(
           List(
-            termQuery("repository", escaped).boost(1000),
-            fuzzyQuery("repository", escaped),
-            termQuery("artifacts", escaped).boost(20),
-            fuzzyQuery("artifacts", escaped),
-            termQuery("organization", escaped).boost(20),
+            termQuery("repository", escaped).boost(40),
+            fuzzyQuery("repository", escaped).boost(4),
+            termQuery("github.description", escaped).boost(20),
+            termQuery("organization", escaped),
             fuzzyQuery("organization", escaped),
-            termQuery("primaryTopic", escaped).boost(10),
-            termQuery("github.topics", escaped).boost(2),
-            fuzzyQuery("github.topics", escaped),
             termQuery("primaryTopic", escaped).boost(
-              if (queryIsATopic) 1000
+              if (queryIsATopic) 5
               else 2
             ),
             termQuery("github.topics", escaped).boost(
-              if (queryIsATopic) 100
+              if (queryIsATopic) 5
               else 2
             ),
             nestedQuery("github.beginnerIssues",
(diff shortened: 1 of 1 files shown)


olafurpg:
Thank you for this contribution! Me and @jvican agree that the search relevance is currently not great and we agree with your intuition that the following changes should improve the results

- boost github description
- remove dependency boost
- deboost the repository name
- boost stars (which should be OK with removed dependency boost)

How did you test this change locally?

We tested this change on the same index that's running in production (via `sbt deployDevServer`) and the search results for "akka" seem to be less relevant compared to the current master.

Before (current master):


Screenshot 2019-05-14 at 18 50 29

Screenshot 2019-05-14 at 18 44 45


After (this PR)
Screenshot 2019-05-14 at 18 44 46

Screenshot 2019-05-14 at 18 50 25


When following the instructions in CONTRIBUTING.md, the local index only contained results for ~130 projects instead of ~7000 project on the production server. I'm not sure how to build the full index locally.

eed3si9n:
The design spec is the parameters that I cooked up in  https://github.com/scalacenter/scaladex/issues/389#issuecomment-490584056 using Google spreadsheet. I am MUCH less confident if I was able to implement to the spec correctly since I actually don't understand when boost takes place etc.

> How did you test this change locally?

I followed the instruction on CONTRIBUTING.md, and thus had the limited set of index myself.

> We tested this change on the same index that's running in production (via sbt deployDevServer) and the search results for "akka" seem to be less relevant compared to the current master.

I think the challenge with GitHub stars is that I don't want to use GitHub stars during the searching, but I do want to use it during the sorting. In terms of the scoring summing makes sense, but maybe the summing also end up picking up completely unrelated projects like scala/scala.

eed3si9n:
Maybe I should just sort the list using Scala instead of trying to figure out Elasticsearch..

olafurpg:
> Maybe I should just sort the list using Scala instead of trying to figure out Elasticsearch.

This sounds like a good idea 👍

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -253,21 +258,17 @@ class DataRepository(
         )
         .should(
           List(
-            termQuery("repository", escaped).boost(1000),
-            fuzzyQuery("repository", escaped),
-            termQuery("artifacts", escaped).boost(20),
-            fuzzyQuery("artifacts", escaped),
-            termQuery("organization", escaped).boost(20),
+            termQuery("repository", escaped).boost(40),
+            fuzzyQuery("repository", escaped).boost(4),
```
eed3si9n: According to the documentation boost only applies to term queries - https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-boost.html

eed3si9n: I am not sure if we should include fuzzy query at all - https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-fuzzy-query.html

> The fuzzy query uses similarity based on Levenshtein edit distance.

adpi2: After testing elasticsearch manually, I can confirm that boost applies to fuzzy queries. We are using version 5.4.2  of elasticsearch. It may not be the case anymore on more recent versions, according to the documentation.

adpi2:
My opinion is that our search engine must include some fuzziness to recover from typos. But I am not sure that the combination of term queries and fuzzy queries are appropriate. I would rather prefer trying match queries, especially for `github.description` that contains full text.

Also, I have the impression that the search result is polluted by irrelevant project that have a lot of github stars (eg `scala/scala` when we search for `akka`). I think I have spotted the problematic query which is the `stringQ`query value. This query is searching for the fuzzy `akka` everywhere in the document (in the description, in the dependencies...). That explains why the number of results is very big for the most dependent on projects (eg `scalatest`).

In the next few hours, I will try to come up to a better solution and I will let you know.

adpi2:
merged with additional commits in #597

PR #582, opened 2020-02-04
Create copy bouton for release install text

joriscode:
Proposal of a copy button for the different release install (Sbt, Amm, ...)

### DIFF
--- a/client/src/main/scala/ch.epfl.scala.index.client/CopyToClipboard.scala
+++ b/client/src/main/scala/ch.epfl.scala.index.client/CopyToClipboard.scala
@@ -1,0 +1,81 @@
+package ch.epfl.scala.index.client
+
+import org.scalajs.dom
+import org.scalajs.dom.document
+import org.scalajs.dom.raw.{Event, HTMLButtonElement, NodeList}
+
+object CopyToClipboard {
+  private val errorPrefix = "[CopyToClipboard] -"
+
+  /**
+   * Add listeners on the copy buttons
+   * @param triggerClass the HTML tag class identifying the buttons to listen
+   */
+  def addCopyListenersOnClass(triggerClass: String): Unit = {
+    val triggers: NodeList = document.getElementsByClassName(triggerClass)
+
+    for (i <- 0 until triggers.length) {
+      val trigger = triggers.item(i)
+      trigger.addEventListener(
+        `type` = "click",
+        listener = listener(targetId = "data-clipboard-target",
+                            displayMsg = Some("Copied!")),
+        useCapture = true
+      )
+    }
+  }
+
+  /**
+   * Listen a copy button
+   * @param targetId identifies the tag to be copied
+   * @param displayMsg Change the button text for 4s by the provided text
+   * @return
+   */
+  private def listener(targetId: String,
+                       displayMsg: Option[String]): Event => Unit = { event =>
+    event.target match {
+      case button: HTMLButtonElement =>
+        val targetAttribute = button.getAttribute(targetId)
+        val target = document.getElementById(targetAttribute)
+
+        if (target != null) {
+          copyToClipboard(target.textContent)
+
+          displayMsg match {
+            case Some(tempMsg) =>
+              val previousContent = button.textContent
+              button.textContent = tempMsg
+              dom.window
+                .setInterval(() => button.textContent = previousContent, 4000)
+
+            case None => ()
+          }
+
+        } else {
+          throw new Exception(
+            s"$errorPrefix no target is associated to the given target attribute"
+          )
+        }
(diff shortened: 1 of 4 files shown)


joriscode:
It seems `bin/scalafmt` does not pass on `master`. Fixing it here https://github.com/scalacenter/scaladex/pull/583

template/src/main/twirl/ch.epfl.scala.index.views/project/install.scala.html
```diff
@@ -26,27 +26,46 @@
   </ul>
   <div class="tab-content">
     <div role="tabpanel" class="tab-pane active" id="sbt">
-      <pre>@release.sbtInstall</pre>
+      <pre id="copy-sbt" value="$@release.sbtInstall">@release.sbtInstall</pre>
```
joriscode: We could factorize it `@installtab(id: String, releaseInstall: String, active: Boolean, linkRef: String, linkText: String)`. Dunno if this any better though

julienrf:
So, now I confirm that I can’t make sbt-sass work on my machine either…

Do you mind posting a screenshot of the changes?

Also, would it be possible to have a Scala.js implementation instead of a JavaScript implementation? The existing client implementation is already written in Scala.js, I prefer if we can keep everything in one place.

joriscode:
> Also, would it be possible to have a Scala.js implementation instead of a JavaScript implementation? The existing client implementation is already written in Scala.js, I prefer if we can keep everything in one place.

I'll look at it later on this week

> So, now I confirm that I can’t make sbt-sass work on my machine either…

Do you have the Ruby gem version of Sass? `gem install sass`, at least version `3.4.0`

> Do you mind posting a screenshot of the changes?

scaladex-copy-button

julienrf:
Hey @joriscode! Sorry, did I just discourage you from finalizing the PR? Please let me know if you don’t have time to do the changes, I’ll give a try!

joriscode:
@julienrf Not at all! I implemented a simple solution in ScalaJs. Let me know what you think before I rebase and remove the previous implementation.
scaladex-copy-btn-2

client/src/main/scala/ch.epfl.scala.index.client/CopyToClipboard.scala
```diff
@@ -0,0 +1,84 @@
...
+        val targetAttribute = button.getAttribute(targetId).stripPrefix("#")
+        val target = document.getElementById(targetAttribute)
+
+        if (target != null) {
+          copyToClipboard(target.textContent)
+
+          displayMsg match {
+            case Some(tempMsg) =>
+              val previousContent = button.innerHTML
+              button.innerHTML = tempMsg
```
julienrf: What do you think of using `textContent` instead of `innerHTML`? `textContent` is a safer default, IMHO, since it prevents DOM injection.

template/src/main/twirl/ch.epfl.scala.index.views/project/install.scala.html
```diff
@@ -26,27 +26,33 @@
   </ul>
   <div class="tab-content">
     <div role="tabpanel" class="tab-pane active" id="sbt">
-      <pre>@release.sbtInstall</pre>
+      <pre id="copy-sbt">@release.sbtInstall</pre>
+      <button class="btn-copy" data-clipboard-target="#copy-sbt">Copy</button>
```
julienrf: I suggest removing the leading `#` from the value of the `data-clipboard-target` attribute.

template/src/main/twirl/ch.epfl.scala.index.views/main.scala.html
```diff
@@ -163,5 +163,8 @@ <h5>Follow us:</h5></li>
 
       Raven.config('https://d9ae900df7ef420b86063800d541d712@@sentry.io/185020').install()
     </script>
+    <script type="text/javascript">
+        addCopyListenersOnClass("btn-copy");
```
julienrf: What about calling this from the `Client.main` entry point instead of having two entry points?

joriscode:
@julienrf I am going to clean the commit history and remove unecessary changes

joriscode:
@julienrf I updated both comments and history

julienrf:
We will need some time before we can publish your feature because of the issue with sass. I’ve tried a few things to make sass work, without success. How did you fix the issue on your side?

joriscode:
@julienrf The issue was coming from Sass installed via Homebrew, with the Ruby gem version it worked correctly. I'm using `Ruby Sass 3.7.4`

julienrf:
Thanks for your patience @joriscode! Your contribution have been deployed (thanks to @adpi2!)

PR #584, opened 2020-02-12
Take path of temporary directory as a configuration parameter

julienrf:
Also remove intermediate directories during clean up


server/src/main/scala/ch.epfl.scala.index.server/Server.scala
```diff
@@ -35,13 +35,14 @@ import org.apache.lucene.queryparser.classic.ParseException
 
 object Server {
   private val log = LoggerFactory.getLogger(getClass)
+  val config = ConfigFactory.load().getConfig("org.scala_lang.index.server")
```
sjrd: Consider making `config` private.

julienrf: No, I use it from another place.

PR #593, opened 2020-03-17
Always use '\n' to split lines when parsing JSON files.

sjrd:
Previously, a platform-dependent line separator was used, which was incorrect on Windows and caused parsing to fail.

Credit for figuring this out goes to @adpi2 ;)

Together with #592, this allows the setup instructions to work on Windows and the server to run.


sjrd:
That's right. Because the files we manipulate are always `\n`-separated. Never platform-new-line-separated.

PR #595, opened 2020-03-18
Upgrade to sbt 1.3.8.

sjrd:
With minimal changes in the sbt plugins to use newer versions that support sbt 1.x.


.travis.yml
```diff
@@ -15,7 +15,6 @@ cache:
   - $HOME/.coursier/
 script:
   - bin/scalafmt --test
-  - export SBT_OPTS="-Xmx1G"
```
sjrd: This was failing the build. The default on Travis is already 2G anyway.

adpi2:
There is this new warning showing up when running most of sbt tasks.

```
[warn] Multiple resolvers having different access mechanism configured with same name 'typesafe-ivy-releases'. To avoid conflict, Remove duplicate project resolvers (`resolvers`) or rename publishing resolver (`publishTo`).
```

It seems to come from sbt 1.3.x. I don't think we should worry about it.

Apart from that, everything work just fine.

PR #597, opened 2020-03-19
Search relevance

adpi2:


sjrd:
@adpi2 Merging between PRs is usually not a good idea. Consider *rebasing* this branch on top of the latest `master`. A rule of thumb is that a Merge commit should never be visible in the list of commits of a PR.

adpi2:
## Description

### Reducing the number of results
This is mostly the result of not searching in the dependencies. I have compared the number of results for those queries:
| query  | master | PR |
|---|---|---|
| `scalatest` | 2862  | 167 |
| `shapeless` | 439 | 206 |
| `scalaz` | 7028 | 218 |
| `akka` | 1114 | 741 |

### Balancing the term matching and popularity scoring
I try to give equal importance to term matching and popularity. A project needs both to appear in the first page. That's why I changed the scoring combining function from `add` to `multiply`. 
I used a logarithmic scale for smoothing the popularity scoring, so that the more popular projects don't appear first when the match scoring is not good enough.
When searching for `json` the 5 top results are now, in order:
- `spray/spray-json`
- `circe/circe`
- `nestorpersist/json`
- `json4s/json4s`
- `playframework/play-json`

### Boosting the github description and github readme
Some popular projects don't have explicit name and they don't use github tags. For instance, the term  `testing` only appears in the description of `scalatest`, the term `database` only appear in the readme of `slick`. I have chosen those boost values:
|field|boost|
|-|-|
|repository|16|
|primaryTopic|8|
|github.description|4|
|organization|4|
|github.topics|4|
|artifacts|2|
|github.readme|1|

### Using full text queries
When entering several terms, the user might be interseted in results that combine some of those terms, the more the better. Exact match queries return only a few results, mostly unpopular. Full text queries return much more result and sort them by relevance and popularity.

## Review
You may want to try those patterns:
- looking for a project: "scalatest", "akka", "spark", "doobie", "zio"
- searching a topic: "json", "web", "testing", "compiler", "distributed"
- searching a combination of terms: "kafka cassandra akka", "spark machine learning" 
- looking for a organization or a library author: "scalacenter", "typelevel", "lihaoyi"
- looking for an artifact inside a project: "scalctic"

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -89,8 +90,13 @@ class DataRepository(
               current = clamp(page),
               totalPages = Math.ceil(r.totalHits / params.total.toDouble).toInt,
               total = r.totalHits
-            ),
-            r.to[Project].toList.map(hideId)
+            ), {
+              // println(r.hits.toList.take(100) map { h =>
+              //   val source = h.sourceAsMap
+              //   s"""${h.score}: ${source("organization")}/${source("repository")}"""
+              // })
```
sjrd: These debug statements probably shouldn't be there.

adpi2: indeed

sjrd:
Overall I like the results.

I observed one regression compared to last time I tried: [searching for `http`](https://index-dev.scala-lang.org/search?q=http) now returns finagle and finatra before http4s. That seems weird.

adpi2:
Nice catch @sjrd.

Their is a bad interaction between the analyzer and the urls (in the readme file) because all urls are tokenized [htpp, www, ..., com]. That's why `http` is now found much more, in project readme that are not much related to http. It would be great if we could exclude urls from analysis.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -230,14 +231,40 @@ class DataRepository(
...
+      else None
+    }
 
-    val queryIsATopic = cachedTopics.contains(params.queryString)
+    val searchQuery =
+      if (escaped == "*") None
+      else
+        Some {
+          boolQuery().should(
+            if (escaped == "*") matchAllQuery()
```
julienrf: This branch seems unreachable: we are already in the else branch of line 241.

PR #602, opened 2020-03-26
Fix autocomplete

adpi2:


data/src/main/scala/ch.epfl.scala.index.data/elastic/SeedElasticSearch.scala
```diff
@@ -22,30 +22,34 @@ class SeedElasticSearch(paths: DataPaths, githubDownload: GithubDownload)(
...
+      Await.result(
+        esClient.execute(deleteIndex(indexName)),
+        Duration.Inf
+      )
     }
 
     val projectFields = List(
-      keywordField("organization")
-        .normalizer("lowercase")
+      textField("organization")
```
julienrf: That seems counterintuitive to me. How come is it better as a text field rather than a keyword?

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -1,667 +1,615 @@
...
+  }
 
-    Await.result(
-      esClient
-        .execute {
-          search(indexName / projectsCollection)
-            .aggregations(
-              termsAggregation(aggregationName).field("github.topics")
-            )
+  private def sortQuery(sorting: Option[String]): SortDefinition =
```
julienrf: Should we use an enumeration rather than a `String`?

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
```diff
@@ -1,667 +1,615 @@
...
+
+  /**
+   * Treats the query inputted by a user as a lucene query
+   *
+   * @param queryString the query inputted by user
+   * @return the elastic query definition
+   */
+  private def luceneQuery(queryString: String): QueryDefinition = {
+    stringQuery(
+      replaceFields(queryString)
```
julienrf: Isn’t this vulnerable to code injection attacks?

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
adpi2: It would be nice because easier to read and less errorprone. Yet the sorting argument come from an http parameter and it could contain any string. There must be a mapping from string to something somewhere. This code was already there and I dont think it is a priority to change it now.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
adpi2: Par exemple, si je clique sur `typelevel` depuis la page de cats
lucene-query

data/src/main/scala/ch.epfl.scala.index.data/elastic/SeedElasticSearch.scala
adpi2: I agree it is counter-intuitive for this one but there is an explanation
As soon as we want to search with text that is inputted by a user, it is better to use a `textField`. If I type `typelevel cats` I won't match the `typelevel` keyword but I would match the `typelevel` text field.
`keywordField` should be used for exact term matching and aggregations only.

adpi2: Another example: if I type `scala js` I would like to match the `scala-js` organization and the `scala-js` repo.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
julienrf: Indeed, sorry I didn’t realize it was already here before.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
julienrf: OK, thanks for the clarification.

server/src/main/scala/ch.epfl.scala.index.server/DataRepository.scala
adpi2: Any comment is appreciated

PR #612, opened 2020-04-01
Add user doc on improving visibility

adpi2:
This is just a markdown file in the `doc/user/` folder, for the moment. In the end, I think, this and all the user documentation should appear on the website under a help section. Something like https://pypi.org/help/

### DIFF
--- a/doc/user/improve-visibility.md
+++ b/doc/user/improve-visibility.md
@@ -1,0 +1,19 @@
+# How to improve the visibility of your project
+
+As a library author or maintainer, the visibility of your project on Scaladex is under your control.
+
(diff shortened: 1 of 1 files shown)


doc/user/improve-visibility.md
```diff
@@ -0,0 +1,19 @@
+# How to improve the visibility of your project
+
+As a library author or maintainer, the visibility of your project on Scaladex is under your control.
+
+As soon as a release is pushed to Maven Central (Sonatype), Scaladex downloads the latest content and metadata from the Github repository. It then uses those metadata to index the project.
```
julienrf, line 5: We also download pom.xml from Maven Central (this is not directly related to SEO, but the way it is currently phrased can bring some confusion I think).

Issue #526, opened 2018-02-20
needn't include old Scala milestones and release candidates


SethTisue:
continuing to list 2.12.0-M3, 2.12.0-RC1, and so on is just noise:

<img width="188" alt="screen shot 2018-02-20 at 10 43 42 am" src="https://user-images.githubusercontent.com/161079/36442570-11b052ac-162b-11e8-9cf2-ca33535b7a8f.png">

as for Scala 2.13, I would suggest that only the latest milestone (2.13.0-M3) is necessary to list, but keeping M1 and M2 around wouldn't be terrible.

PR #614, opened 2020-04-02
Dont show invalid sbt and scalajs versions

adpi2:
This PR is deployed to https://index-dev.scala-lang.org

## First commit

- minor version is optional for parsing scalajs version (e.g. sjs1_2.12)
- drop out invalid scalajs (< 0.6) and sbt versions from home page and filters

## Second commit

- refactor ScalaTarget as an ADT of cases ScalaJvm, ScalaJs, ScalaNative or SbtPlugin

## Third commit

- drop out invalid scalajs and sbt versions from result list


template/src/main/twirl/ch.epfl.scala.index.views/search/resultList.scala.html
```diff
@@ -53,7 +54,7 @@ <h4>@project.reference.organization/@project.reference.repository</h4>
               @if(project.scalaNativeVersion.nonEmpty){
                 <div>
                   Scala-Native:
-                  @for(v <- sortVersions(project.scalaNativeVersion)) {
+                  @for(v <- sortVersions(project.scalaNativeVersion, _ => true)) {
```
mlachkar, line 57: OPT : I would have created ScalaTarger.isValidScalaNativerVersion

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
```diff
@@ -24,20 +24,21 @@ case class OtherPreRelease(o: String) extends PreRelease {
...
+  def isSemantic: Boolean = {
     patch.isDefined &&
-    !patch2.isDefined &&
-    preRelease.map(_.isSemantic).getOrElse(true)
+    patch2.isEmpty &&
+    preRelease.forall(_.isSemantic)
   }
 
   override def toString: String = {
+    val minorPart = minor.map(m => s".$m").getOrElse("")
```
sjrd, line 41:
```suggestion
    val minorPart = minor.map("." + _).getOrElse("")
```
That would preserve the symmetry with `patchPart` and `patch2Part`.

Alternatively, also amend `patchPart` and `patch2Part` to use a string interpolation like you're doing here.

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
```diff
@@ -68,20 +68,40 @@ case class SemanticVersion(
...
+    Ordering.by {
+      case None                       => (3, None, None)
+      case Some(ReleaseCandidate(rc)) => (2, Some(rc), None)
+      case Some(Milestone(m))         => (1, Some(m), None)
+      case Some(OtherPreRelease(pr))  => (0, None, Some(pr))
+    }
 
-  implicit val ordering: Ordering[SemanticVersion] = Ordering.by {
-    x => (x.major, x.minor, x.patch, x.patch2, x.preRelease)
+  implicit def ordering: Ordering[SemanticVersion] = Ordering.by { x =>
```
sjrd, line 79: Why use a `def` here? That doesn't seem like an improvement to me.

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
```diff
@@ -68,20 +68,40 @@ case class SemanticVersion(
...
-    case ReleaseCandidate(rc) => (Some(rc), None, None)
-    case Milestone(m) => (None, Some(m), None)
-    case OtherPreRelease(pr) => (None, None, Some(pr))
-  }
+  private implicit val preReleaseOrdering: Ordering[Option[PreRelease]] =
+    Ordering.by {
+      case None                       => (3, None, None)
+      case Some(ReleaseCandidate(rc)) => (2, Some(rc), None)
+      case Some(Milestone(m))         => (1, Some(m), None)
+      case Some(OtherPreRelease(pr))  => (0, None, Some(pr))
```
sjrd, line 76: Wouldn't the following be enough?
```suggestion
      case None                       => (3, None)
      case Some(ReleaseCandidate(rc)) => (2, Some(rc))
      case Some(Milestone(m))         => (1, Some(m))
      case Some(OtherPreRelease(pr))  => (0, Some(pr))
```

adpi2: It is deployed to http://index-dev.scala-lang.org

adpi2:
It is deployed to http://index-dev.scala-lang.org

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
adpi2: No, its not enough because `pr` is a String whereas `rc` and `m` are Long

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
adpi2: Indeed

template/src/main/twirl/ch.epfl.scala.index.views/search/resultList.scala.html
adpi2: Ok

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
adpi2: Ok

model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
sjrd: Oh, OK.

sjrd:
> It is deployed to http://index-dev.scala-lang.org

Thanks!

Looking at https://index-dev.scala-lang.org/search?q=*&scalaJsVersions=1, I suspect there might be something wrong. There are definitely a lot more than 9 libraries with support for Scala.js 1.x. `cats-core`, and `cats` in general, should be there, for example.

Also, could we *display* Scala.js 1 as `1.x` rather than just `1`. I don't know why, but just `1` feels unnatural. We could also display 0.6 as `0.6.x` for symmetry, though for some reason seeing `0.6` doesn't bother nearly as much as `1`.

adpi2:
There is only 9 scala.js 1.x libraries because I have reindexed an old snapshot dating 2017. Reindexing the entire database will be much longer but I will try that in a few moments.

The scala.js version comes from the name of the artifact (e.g. `cats-core_sjs0.6_2.12` or `cats-core_sjs1_2.12`). It may be confusing not to stick with this naming convention and show `1.x` instead of `1`, don't you think?

sjrd:
> There is only 9 scala.js 1.x libraries because I have reindexed an old snapshot dating 2017. Reindexing the entire database will be much longer but I will try that in a few moments.

Hum, then that means that it is still indexing libraries for milestones/RCs of 1.0.0. We'd like those not to be indexed anymore, as it is not useful, and it is actually downright misleading to say that they support 1.x, because support for 1.0.0-Mx or 1.0.0-RCx does *not* guarantee support for 1.0.0+.

> The scala.js version comes from the name of the artifact (e.g. `cats-core_sjs0.6_2.12` or `cats-core_sjs1_2.12`). It may be confusing not to stick with this naming convention and show `1.x` instead of `1`, don't you think?

When we talk about Scala.js 1.x, we always talk about 1.x. Nobody writes Scala.js 1. It doesn't really matter what the encoding of artifacts is. What matters for display is how people talk about the version number. So IMO 1 is more confusing than 1.x.

adpi2:
@sjrd you are right about the milestones and RCs of scala.js 1.0.0 being indexed.

I need clarification on this: what do we do with future milestones and RCs of scala.js 1.1? Do we include them in 1.x or do we separate them in a 1.1 version?

This issue is not specific to scala.js: see #526 for scala jvm. I'll treat them both here. And what about scala native? Should we display all versions from 0.1 to 0.4?

adpi2:
Also, I am afraid that the reindexing of the entire database does not work. It has been stuck since yesterday 9pm. I have to work on that too before merging.

sjrd:
> I need clarification on this: what do we do with future milestones and RCs of scala.js 1.1? Do we include them in 1.x or do we separate them in a 1.1 version?
> 
> This issue is not specific to scala.js: see #526 for scala jvm. I'll treat them both here. And what about scala native? Should we display all versions from 0.1 to 0.4?

The general rule for me would be: only index a milestone or RC if it is *the very latest* available version. That means that:

* For Scala/JVM: index no milestone nor RC, since the latest stable release is 2.13.1, and there are no M or RC that are > 2.13.1
* For Scala.js: same reasoning; index no milestone nor RC
* For Scala Native: index 0.4.0-M2, since it is the latest available version (do not index 0.4.0-M1, assuming that even exists, since 0.4.0-M2 is more recent)

This general rule seems to work well for all the cases I can think of.

In addition to that, for stable releases, only index the following versions IMO:

* Scala/JVM >= 2.10
* Scala.js >= 0.6
* Scala Native >= 0.3

adpi2:
I am not clear about the meaning of "binary" in "binary version". Does it mean "binary compatibility"?

As far as I know, the rules of "binary compatibility" are not the same in Scala JVM and Scala.js, since 2.12 and 2.13 are not compatible but scala.js 1.0 and 1.1 should be compatible, right? It's even more complicated than that because sjs 0.5 is not compatible with sjs 0.6 and sjs 1.0.0-M2 is not compatible with 1.0.0.

sjrd:
When you parse the artifact names, all you ever get is the *binary version*. When you see `_2.13`, the binary version is 2.13. When you see `_2.13.0-RC1`, the binary version is 2.13.0-RC1. It's the same for Scala.js: when you see `_sjs0.6`, the binary version is 0.6. `_sjs1` means 1. `_sjs1.0-RC2` means 1.0-RC2.

So in Scaladex all you ever see is the binary version, since you *cannot* determine the full version (you can't tell from `_2.13` whether it was built with 2.13.0 or 2.13.1). So as far as Scaladex is concerned, *binary version = version*.

And as soon as two binary versions are different, they create entirely different and incompatible ecosystems. So Scaladex should never try to be "smart" and collate together "similar" versions. No, if Scaladex sees two different (even if "similar") versions, they really *are* different, and must *never* be lumped together in the same bag.

adpi2:
Alright then.

I like your point of view. Scaladex used to truncate binary version `2.12.3` to `2.12` but it makes things complicated. Your proposal is more predictable, I think. Yet, we should not display those weird `2.12.3` binary versions that do not conform to the norm. Agreed ?

sjrd:
The only reason you'd see a binary version like `2.12.3` is if the artifact is a compiler plugin. Otherwise the binary version would already be `2.12`. `2.12.3` is not a "valid" binary version for non-compiler-plugin libraries.

How to deal with compiler plugins, I am not sure. My hunch is that you can just not index them. They are often accompanied by a genuine library anyway or by an sbt plugin, so that would be their entry point in Scaladex. For compiler plugins that are really just compiler plugins (and I can't name one off the top of my head), they would not be indexed, but I think that's fine for now. We can reconsider later if we need to find a solution for them.

So concretely, you can ignore "invalid" binary versions. Valid binary versions are:
* JVM: `x.y`, `x.y.z-Mw`, `x.y.z-RCw`
* JS: `0.y`, `x`, `x.y-Mw`, `x.y-RCw`
* Native: `0.y`, `x.y.z-Mw`, `x.y.z-RCw`
* sbt: `x.y`

adpi2:
One more precision: so far every artifact is indexed even if it's target type is considered obsolete or invalid. I will keep it that way for the moment.

My goal is to discard invalid versions from the front page, the result list and the dropdown list of a project.

sjrd:
Yes, I'm sure ;) It is initially computed [here](https://github.com/scala-js/scala-js/blob/4eac249d0e4a922f5fcdb7c8e7c48c95260438c0/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala#L37-L52), which, for a pre-release version, delegates to `binaryEmitted`, which is defined [here](https://github.com/scala-js/scala-js/blob/4eac249d0e4a922f5fcdb7c8e7c48c95260438c0/ir/src/main/scala/org/scalajs/ir/ScalaJSVersions.scala#L21).

adpi2:
Here it comes ==> https://index-dev.scala-lang.org/

model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
```diff
@@ -80,6 +88,47 @@ object ScalaJvm {
...
+
+  def isValid(version: BinaryVersion): Boolean = {
+    version == `0.3` || version >= `0.4.0-M2`
+  }
+}
+
+object SbtPlugin {
+  val `0.13`: BinaryVersion = MinorBinary(0, 13)
+  val `1.0`: BinaryVersion = MinorBinary(1, 0)
+  val `1.x`: BinaryVersion = MajorBinary(1)
```
sjrd, line 125: I don't think that sbt uses `1.x` or `1`. They (mistakenly) always use the binary version `1.0`, even though it also applies to all 1.1.0 and following.

adpi2: That's right. I will remove this.
 Thank you for reviewing.

PR #615, opened 2020-04-15
Show and filter dotty versions

adpi2:
You can see the result here https://index-dev.scala-lang.org

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
@@ -1,6 +1,10 @@
 package ch.epfl.scala.index.model
 package release
 
+import ch.epfl.scala.index.model.release
+import fastparse.NoWhitespace._
+import fastparse._
+
 object ScalaTargetType {
   implicit val ordering: Ordering[ScalaTargetType] = Ordering.by {
     case Jvm    => 5
@@ -27,7 +31,7 @@ it's dependency on scala-library for example.
 
  */
 sealed trait ScalaTarget extends Ordered[ScalaTarget] {
-  def scalaVersion: BinaryVersion
+  def languageVersion: LanguageVersion
   def render: String
   def encode: String
   def targetType: ScalaTargetType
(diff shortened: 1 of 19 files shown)


model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
```diff
@@ -0,0 +1,98 @@
+package ch.epfl.scala.index.model
+package release
+
+import fastparse.NoWhitespace._
+import fastparse._
+
+sealed trait BackendVersion {
```
julienrf, line 7: It would be nice to have some scaladocs explaining what this sealed trait models, and what `family` and `isValid` are.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
```diff
@@ -0,0 +1,98 @@
...
+      `0.19`.version,
+      `0.20`.version,
+      `0.21`.version,
+      `0.22`.version,
+      `0.23`.version
+    )
+
+  def isValid(version: BinaryVersion): Boolean =
+    latestVersions.contains(version) ||
+      (version > `0.23`.version && version < MinorBinary(0, 30))
```
julienrf, line 71: Why not just this?

```suggestion
  def isValid(version: BinaryVersion): Boolean = version > `0.16`.version
```

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
julienrf: Also, I’m not sure `BackendVersion` is a good name for that. I think “binary version” or “scala version” would be better.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: The reasons are:
- we want to distiguinsh between scala version and dotty version: `2.12` is not a valid dotty version
- previous milestones and RC are not relevant: `0.22-RC1` is not a valid dotty version
but a future `0.24-RC1` is a valid dotty version
- some sbt plugin have bad version: `_0.31_2.12` instead of `_2.12_0.31` and we don't want to interpret this as a plugin for sbt 2.12 and dotty 0.31

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: Agreed for `ScalaVersion` and for the scaladocs

adpi2: Ok for the scaladoc

adpi2: I dont think that `BinaryVersion` is a better name because there already is a `BinaryVersion` ADT that is used for describing the scala-js version and the scala-native version

adpi2: And the problem with `ScalaVersion` is to find a name of the two subtypes that are currently named `ScalaBackend` and `DottyBackend`. Is 0.23 a ScalaVersion? My answer would be not yet because dotty is not Scala 3 yet. But still dotty can be used as a backend for the javascript and scala-native compilers. That is why I choosed the name `BackendVersion`. A scala js target is made of a `BackendVersion` and a javascript `BinaryVersion`. It makes sense to me.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
julienrf: I see, thanks for the explanation!

Does that mean that we will have to update scaladex each time a new Dotty RC is released?

adpi2: That's a good question. As of now I have seen no libraries issued with a Dotty RC version. But we will have to update Scaladex when Dotty version 0.30 will be released. And this is also true for scala 3.0 scala.js 2.x and scala-native 0.4 (not sure for this one) and sbt 2.x.

Hopefully we will make the ScalaTarget parser more robust and then we will make it forward compatible.

julienrf: Maybe we can assume for now that Dotty libraries will always target non-RC Dotty releases? So maybe we can write a more general test that checks that the major version is 0 and that the patch version is a number (it doesn’t contain “RC...”). Wdyt?

adpi2: Yes it would be better. But we still want to discard the versions older than 0.17 and the badly formated sbt versions.

julienrf: For badly formatted sbt versions, I believe we could make a more robust parser: it should only accept one binary suffix, not two.

adpi2: I don't really know where do those sbt target versions come frome but they have two parts, a scala version and a sbt version. Most of the time we have _<scala version>_<sbt version> but sometimes they are inverted.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
sjrd: *Backend* is definitely the wrong word. If you want to reason along your reasoning, it should be *frontend* version. Scala 2/Dotty are frontends, not backends. Backends would be JVM/JS/Native. But even there for JS itself we have a front-end and a back-end. It's all very relative. But in any case the distinction scalac/dotty is the very first "end" of the language, so in any naming scheme it would be a frontend, not a backend.

But I think the best name for this---independently of the philosophical debate whether Dotty is "Scala" or not---is `LanguageVersion`. This definitely represents a version of the *language*, whether that language is called "Scala" or "Dotty". So I would recommend `LanguageVersion` with the subtypes `ScalaVersion` and `DottyVersion`.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def Parser[_: P]: P[ScalaBackend] = {
+    BinaryVersion.Parser
+      .filter(isValid)
+      .map(ScalaBackend(_))
+  }
+}
+
+object DottyBackend {
+  val `0.23`: DottyBackend = DottyBackend(MinorBinary(0, 23))
```
sjrd, line 63: Why singling out any Dotty version? A new one comes up every 6 weeks. You don't want to have to modify this codebase every 6 weeks.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def isValid(version: BinaryVersion): Boolean =
+    version match {
+      case MinorBinary(major, minor) if major == 0 && minor < 30 => true
```
sjrd, line 67:
```suggestion
      case MinorBinary(major, minor) if major == 0 => true
```
All 0.x versions are Dotty version. There's no artificial cap at 30.

model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
```diff
@@ -37,86 +41,92 @@ sealed trait ScalaTarget extends Ordered[ScalaTarget] {
...
+  }
 
-  def isValid(version: BinaryVersion): Boolean =
-    stableBinaryVersions.contains(version)
+  def Parser[_: P]: P[ScalaTarget] =
+    ("_" ~ BackendVersion.Parser).map(ScalaJvm.apply)
 }
 
-object ScalaJs {
+object ScalaJs extends Parsers {
```
sjrd, line 100: Why is this necessary? `ScalaNative` below doesn't have it.

adpi2: it is not :)

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
julienrf: The thing that I don’t understand is why is a pattern such as `_x.y_α.β` successfully parsed in the first place as a valid Scala version?

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: You are totally right. I have inverted the meaning of Backend and Frontend in my head. `LangugageVersion` as you suggest is a better name.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: This is just a constant value that is used in the tests. But we dont really need it.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: Right but I definetely dont want to interpret this sbt plugin version _0.31_2.12 version as a plugin of sbt  2.12 compiled with Dotty 0.31. Would a `isValid` sbt version that discard sbt `2.12` be better?

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: Its not a Scala version its a Sbt target version. Nevertheless I've moved the logic of discarding bad sbt plugin target to SbtPlugin which is indeed better.

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
adpi2: Yes I think it is actually

model/src/main/scala/ch.epfl.scala.index.model/release/BackendVersion.scala
sjrd: If it is only relevant in the tests, you should declare it in the tests ;)

PR #624, opened 2020-06-11
Index dependencies separately from the releases

### LINKED ISSUE #619: Slowness when loading scalatest page
When querying a project page all the project's releases are loaded. In the case of `scalatest/scalatest` this is very slow because of the number of releases (2820) and the number of dependents (287650).

This is counterproductive because only the dependents of the default release are kept. I suggest to remove the `reverseDependencies` field in the `Release` class and to compute the number of dependents of the default release by querying elasticsearch on the `scalaDependencies` field.

adpi2:
Fix #619 

The main goal of this PR is to improve the performance of loading a project page.

Loading a project page used to load all the dependencies and all the dependents of the project releases, which is too much for some project. Ex: scalatest has dozens of thousands of dependents.

In this PR the dependencies are stored separately from the releases, so that we can load the releases first, select the default release and then load its dependencies and dependents only.


adpi2:
The first two commits are pure refactoring:
- Creating a `search` module and move the `DataRepository` class to this module
- Encapsulate the elasticsearch client inside the `DataRepository`: all access to elasticsearch goes through the `DataRepository`

The third commit implements the desired changes:
- Create a distinct mapping for `Release` and `ScalaDependency` called respectively `ReleaseDocument` and `DependencyDocument`.
- Insert and retrieve the dependencies when necessary.

The last commit lower the memory consumption of the `SeedElasticsearch` job that creates and fills the elasticsearch index, so that it is faster and easier to deploy.

adpi2:
The changes have been deployed to https://index-dev.scala-lang.org/

To compare the performance before and after the change I suggest the scalatest_2.12_3.0.5 page which has more than 1500 dependents:
- [before](https://index.scala-lang.org/scalatest/scalatest/scalatest/3.0.5?target=_2.12)
- [after](https://index-dev.scala-lang.org/scalatest/scalatest/scalatest/3.0.5?target=_2.12)

server/src/main/scala/ch.epfl.scala.index.server/routes/ProjectPages.scala
```diff
@@ -293,15 +314,12 @@ class ProjectPages(
...
-                          form
-                        )
-                      ) { _ =>
-                        Thread.sleep(1000) // oh yeah
-                        redirect(Uri(s"/$organization/$repository"), SeeOther)
-                    }
+                        onSuccess {
+                          updateProject(Project.Reference(organization, repository), form)
+                        } { _ =>
+                          Thread.sleep(1000) // oh yeah
```
mlachkar: I know it was already there, but I don't understand why it's necessary, since onSuccess will wait to resolve the future first.

adpi2: I am not sure either. I think elasticsearch update are not striclty atomic. We give it some time to properly index the project, so that when the user is redirected the changes are likely to be effective.

data/src/main/scala/ch.epfl.scala.index.data/maven/PomsReader.scala
```diff
@@ -27,27 +29,30 @@ object PomsReader {
...
+            if (acc.contains(sha)) acc
+            else acc + (sha -> pom)
+        }
+        .values
     }
 
-    val ivysDescriptors = SbtPluginsData(paths.ivysData).load()
-
-    centralPoms ::: bintrayPoms ::: usersPoms ::: ivysDescriptors
+    Await.result(allPoms, Duration.Inf)
```
mlachkar, line 55: OPt Why doing the await here and not higher?

adpi2: I wanted to maintain the method signature. But now that I read the code again, I think that `ReadPoms.iterator` being a future is nonsense. I will change that and move the Await out of it.

server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/IndexingActor.scala
```diff
@@ -82,10 +80,10 @@ class IndexingActor(
 
       val converter = new ProjectConvert(paths, githubDownload)
 
-      val (newProject, newReleases) = converter(
+      val (newProject, newReleases, dependencies) = converter(
```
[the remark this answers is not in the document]
adpi2: Totally agree on this.

PR #628, opened 2020-07-07
Fix project update when indexing new artifact

### LINKED ISSUE #626: Missing artifacts in search api
Eg https://index.scala-lang.org/api/search?q=zio&target=JVM&scalaVersion=2.13
returns 
```
{
  "organization" : "zio",
  "repository" : "zio",
  "logo" : "https://avatars2.githubusercontent.com/u/49655448?v=4",
  "artifacts" : [ "zio-test" ]
}
```

### LINKED ISSUE #627: Missing supported scala versions for some projects
As example, jsoniter-scala is released for Scala 2.11.x, 2.12.x from the first release but [the Scala Index have stopped showing that](https://index.scala-lang.org/search?q=jsoniter-scala):

image

adpi2:
Fixes #626 
Fixes #627

The bug is:
- Scaladex receives a publish request of a maven artifact
- it creates the release object corresponding to this artifact
- it looks into an in-memory cache for other releases of the same project
- it builds the project object based on the cached and new releases
- it inserts the new release and update the project in elasticsearch
- it caches the new release

The problem is that the cache is empty after start-up, so the project is updated with a lot of missing metadata (list of artifacts, supported versions...)

This PR fixes the bug by:
- loading the owner project and all its releases from elasticsearch
- building the updated project based on the loaded releases

There is no cache anymore.

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala
@@ -24,11 +24,11 @@ class ProjectConvert(paths: DataPaths, githubDownload: GithubDownload)
 
   /**
    * @param pomsRepoSha poms and associated meta information reference
-   * @param cachedReleases use previous released cached to workaround elasticsearch consistency (write, read)
+   * @param indexedReleases use previous indexed releases to update the project consistently
    */
   def convertAll(
       pomsRepoSha: Iterable[(ReleaseModel, LocalRepository, String)],
-      cachedReleases: Map[Project.Reference, Set[Release]]
+      indexedReleases: Map[Project.Reference, Seq[Release]]
   ): Iterator[(Project, Seq[Release], Seq[ScalaDependency])] = {
 
     val githubRepoExtractor = new GithubRepoExtractor(paths)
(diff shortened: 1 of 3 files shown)


data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala
```diff
@@ -28,7 +28,7 @@ class ProjectConvert(paths: DataPaths, githubDownload: GithubDownload)
    */
   def convertAll(
       pomsRepoSha: Iterable[(ReleaseModel, LocalRepository, String)],
-      cachedReleases: Map[Project.Reference, Set[Release]]
+      cachedReleases: Map[Project.Reference, Seq[Release]]
```
julienrf, line 31: If I try to re-publish to scaladex a release that was already indexed, are going to count it twice?

julienrf:
```suggestion
      indexedReleases: Map[Project.Reference, Seq[Release]]
```

Or, `oldReleases`, but you already use this identifier below.

adpi2: > If I try to re-publish to scaladex a release that was already indexed, are we going to count it twice?
No because there is filter in the `IndexingActor`

indexedReleases is a better name indeed :)

adpi2: I plan to split this method in two because it has two distinct usage:
- converting a newly published artifact to a project and a release
- converting all the artifacts that are stored in the filesystem and group them by project

The current `Map[Project.Reference, Seq[Release]]` signature has no sense in any of the two usages.

PR #642, opened 2020-11-10
Add gihub actions and update scalafmt

tgodzik:


tgodzik:
Ok, it seems to have run correctly on https://github.com/tgodzik/scaladex/actions/runs/356184937

I think it will not work locally until we merge it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
+on:
+  push:
+    branches:
+      - master 
+  pull_request:
```
adpi2, line 6: We need to run the CI on pull requests that target `master` as well.

```
 pull_request:
   branches: 
      -  master
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          java-version: "adopt@1.8"
+      - uses: coursier/cache-action@v5
+      - name: Run tests
+        run: |
+          cd scaladex
+          ./bin/scalafmt --test
+          sbt "data/run elastic"
+          sbt test
+          sbt server/universal:packageBin
```
adpi2, line 35: Could you make individual steps out of those?
So that it is easier to spot what is failing

adpi2 (changes requested):

.github/workflows/ci.yml
tgodzik: I think it's the default, but I will add it just in case

.github/workflows/ci.yml
tgodzik: Sure!

.github/workflows/ci.yml
tgodzik: Added!

.github/workflows/ci.yml
tgodzik: Changed to separate steps:
https://github.com/tgodzik/scaladex/runs/1380895668?check_suite_focus=true

what do you think?

.github/workflows/ci.yml
adpi2: okay I thought it was missing and that was why the CI was not running

.github/workflows/ci.yml
adpi2: Perfect thank you!

adpi2 (approved):

Issue #634, opened 2020-11-10
Rename "dotty" family to "scala3"


adpi2:
In the front page and the search page, the `dotty` family name should be renamed `scala3`.
![Screenshot_2020-11-10 Scaladex](https://user-images.githubusercontent.com/13123162/98674578-4c0e5000-2359-11eb-995e-d5663ad2defd.png)
In the search page:
![Screenshot_2020-11-10 Search cats](https://user-images.githubusercontent.com/13123162/98674596-53355e00-2359-11eb-91bc-455891e52b20.png)
![Screenshot_2020-11-10 Search cats(1)](https://user-images.githubusercontent.com/13123162/98674600-54ff2180-2359-11eb-8fb4-63140938900c.png)

In the project page:
- `dotty 0.27` should not be renamed even if it is part of the `scala3 family`
- `3.0.0-M1` binary version should appear as `scala 3.0.0-M1`

![Screenshot_2020-11-10 typelevel cats](https://user-images.githubusercontent.com/13123162/98674649-67795b00-2359-11eb-8145-7253e4116029.png)

PR #643, opened 2020-11-10
display correctly scala 3 versions 

### LINKED ISSUE #635: scala3- <compiler / library> artifacts are not indexed
In https://index.scala-lang.org/lampepfl/dotty/, the `scala3-library` and `scala3-index` are not indexed.

- Did Maven Central pushed them?
- Are they present in the prod `scaladex-index` files?
- Did the indexing in elasticsearch fail?

mlachkar:
image

Fix #635, fix #635 and fix https://github.com/scalacenter/scaladex/issues/634

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
@@ -6,7 +6,7 @@ import fastparse._
 
 /**
  * A [[LanguageVersion]] is the binary version of the compiler frontend.
- * It can be either a [[ScalaVersion]] or a [[DottyVersion]]
+ * It can be either a [[ScalaVersion]] or a [[Scala3Version]]
  */
 sealed trait LanguageVersion {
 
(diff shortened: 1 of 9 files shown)


model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
```diff
@@ -34,8 +34,12 @@ final case class ScalaVersion(version: BinaryVersion) extends LanguageVersion {
 }
 
 final case class DottyVersion(version: BinaryVersion) extends LanguageVersion {
-  def family = "dotty"
-  def render: String = s"dotty $version"
+  def family = "3.0"
```
adpi2, line 37: I think `scala3` is a better family name than `3.0` because it also contains `0.27` versions.

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
```diff
@@ -34,8 +34,12 @@ final case class ScalaVersion(version: BinaryVersion) extends LanguageVersion {
 }
 
 final case class DottyVersion(version: BinaryVersion) extends LanguageVersion {
```
adpi2, line 36: I think we should rename the class to `Scala3Version`|

adpi2 (changes requested):

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
mlachkar: yes

adpi2 (approved):

PR #644, opened 2020-11-10
Fix #638: Add filters in the compatibility grid page.

### LINKED ISSUE #638: Add filters in the compatibility grid page
Mimic the target and binary version filters in the search page (https://index.scala-lang.org/search?q=cats) to filter the versions that are shown in the compatibility grid page: https://index.scala-lang.org/artifacts/typelevel/cats

Also add a version filter to filter on the version of the library.

sjrd:
We add checkboxes in the header cells of the table. Only lines
that match all selected checkboxes will be displayed.

The filtering is done entirely on the client-side.

image


adpi2 (approved):
LGTM thanks!

PR #647, opened 2021-01-08
Exclude Dotty versions

adpi2:


mlachkar (approved):

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
```diff
@@ -67,8 +67,8 @@ object ScalaVersion {
 object Scala3Version {
   def isValid(version: BinaryVersion): Boolean =
     version match {
-      case MinorBinary(major, minor) if major == 0 && minor < 30 =>
-        true // dotty versions
+      // case MinorBinary(major, minor) if major == 0 && minor < 30 =>
```
tgodzik, line 70: Why not just remove it?

PR #650, opened 2021-01-26
Replace Scala Native 0.4.0-M2 target with binary version 0.4

WojciechMazur:
This PR removes listing of 0.4.0-M2 Scala Native release in favor of  0.4.0.

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
@@ -120,12 +120,12 @@ object ScalaJs {
 
 object ScalaNative {
   val `0.3`: BinaryVersion = MinorBinary(0, 3)
-  val `0.4.0-M2`: BinaryVersion =
-    release.PreReleaseBinary(0, 4, Some(0), Milestone(2))
+  val `0.4`: BinaryVersion = MinorBinary(0, 4)
 
-  def isValid(version: BinaryVersion): Boolean = {
-    version == `0.3` || version >= `0.4.0-M2`
-  }
+  private val stableBinaryVersions = Set(`0.3`, `0.4`)
+
+  def isValid(version: BinaryVersion): Boolean =
+    stableBinaryVersions.contains(version)
 
   def Parser[_: P]: P[ScalaTarget] =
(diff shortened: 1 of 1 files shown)


model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
```diff
@@ -120,12 +120,12 @@ object ScalaJs {
 
 object ScalaNative {
   val `0.3`: BinaryVersion = MinorBinary(0, 3)
-  val `0.4.0-M2`: BinaryVersion =
-    release.PreReleaseBinary(0, 4, Some(0), Milestone(2))
+  val `0.4`: BinaryVersion = MinorBinary(0, 4)
 
-  def isValid(version: BinaryVersion): Boolean = {
-    version == `0.3` || version >= `0.4.0-M2`
```
adpi2, line 127: I would have expected `0.4` to be greater than `0.4.0-M2` but apparently it is not:

```scala
assert(MinorBinary(0, 4) > PreReleaseBinary(0, 4, None, Milestone(2))) // success
assert(MinorBinary(0, 4) > PreReleaseBinary(0, 4, Some(0), Milestone(2))) // failure
```
Looks like a bug to me... Would you try to fix it?

adpi2 (approved):
So apparently there is a bug in Scaladex. But your PR is nevertheless probably the right thing to do.

Do you know which library has already been released for Scala Native `0.4`?

WojciechMazur:
There's been a few, I'll list just a few I know should be indexed in scaladex: 
* [utest](https://mvnrepository.com/artifact/com.lihaoyi/utest)
* [os-lib](https://mvnrepository.com/artifact/com.lihaoyi/os-lib)
* [snUnit](https://mvnrepository.com/artifact/com.github.lolgab/snunit)
* [scalaCheck](https://mvnrepository.com/artifact/org.scalacheck/scalacheck) (this one was published for all possible targets 0.3.x, 0.4.0-M2, 0.4.x)

model/src/main/scala/ch.epfl.scala.index.model/release/ScalaTarget.scala
WojciechMazur: Sure, not a problem

PR #652, opened 2021-02-04
Upgrade to Elastic 7.10.2

adpi2:
Also dev is now deployed to icvm0042.epfl.ch where elasticsearch 7.10.2 is installed.

https://index-dev.scala-lang.org/


search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
```diff
@@ -95,85 +148,92 @@ class DataRepository(esClient: TcpClient, indexName: String)(implicit
...
       val requests = dependencies.map { d =>
-        indexInto(indexName / dependenciesCollection)
+        indexInto(dependencyIndex)
           .source(DependencyDocument(d))
       }
-      esClient.execute(bulk(requests))
+      Future
+        .sequence(
+          requests
+            .grouped(1000)
```
vincenzobaz, line 184: Do you to do this manually?

search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
```diff
@@ -1,41 +1,53 @@
...
+import com.sksamuel.elastic4s.requests.searches.aggs.responses.bucket.Terms
+import com.sksamuel.elastic4s.http.JavaClient
 
 /**
  * @param esClient TCP client of the elasticsearch server
  */
-class DataRepository(esClient: TcpClient, indexName: String)(implicit
+class DataRepository(
+    esClient: ElasticClient,
+    container: Option[ElasticsearchContainer],
```
vincenzobaz, line 39: I don't like having his here, I am not sure of how to manage it :/
It is here only for `close` if I understand correctly...

vincenzobaz (approved):
The PR mostly adapts the codebase to the new ES API.
The changes introduced by testcontainer are minimal.

search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
```diff
@@ -50,43 +62,84 @@ class DataRepository(esClient: TcpClient, indexName: String)(implicit
...
-        if (exists) esClient.execute(deleteIndex(indexName))
-        else Future.successful(())
-    } yield ()
+    def delete(index: String): Future[Unit] = {
+      for {
+        exists <- esClient.execute(indexExists(index)).map(_.result.isExists)
+        _ <-
+          if (exists) esClient.execute(deleteIndex(index))
+          else Future.successful(())
+      } yield ()
```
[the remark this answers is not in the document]
adpi2: you suggestion returns `Future[Future[Unit]]` which I need to flatten

search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
adpi2: I had to chunk this request because of the size limit imposed by the default configuration of elasticsearch 7

search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
adpi2: I agree. A potential solution is to create a `Closeable` abstraction on top of `ElasticClient` that can encapsulate the container.
But the more I think of it, the more I convince myself that the container lifecycle should be managed by sbt. So I don't want to add more code that I will probably revert soon.

search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
vincenzobaz: I understand. What I do not like is that *you* have to do it *manually* and here. Isn't `bulk` meant to address this problem?

adpi2: `bulk` is an HTTP endpoint to insert many document in a single request. So the logic is not in the library but in the elasticsearch server.

search/src/main/scala/ch.epfl.scala.index.search/DataRepository.scala
vincenzobaz: The container is required only for testing. So tests could be another point where this could be handled instead of sbt

adpi2: It is also required for running the Scaladex locally. So it's rather a matter of local/remote than a matter of main/test.

PR #656, opened 2021-02-16
Align search API with search page

adpi2:
This PR deprecates the current `api/search` endpoint by renaming it `api/search-old`.

Then it introduces a new `api/search` endpoint whose query inputs, behavior and response are similar to the `search` page enpoint.

This new API can be tested on the dev environment: https://index-dev.scala-lang.org

Example:

- Returns the 20 first Scala 3 projects in terms of number of stars:
https://index-dev.scala-lang.org/api/search?q=*&scalaVersions=scala3&sort=stars&total=20


adpi2:
@romanowski

server/src/main/scala/ch.epfl.scala.index.server/routes/api/SearchApi.scala
```diff
@@ -201,16 +212,35 @@ class SearchApi(
...
             }
           }
       }
     }
 
+  private def convertProject(
+      cli: Boolean
+  )(project: Project): SearchApi.Project = {
+    val artifacts = if (cli) project.cliArtifacts.toList else project.artifacts
+    SearchApi.Project(
```
romanowski, line 228: Can we also include defaultArtifact since search API under the hood redirects to it anyway.

server/src/main/scala/ch.epfl.scala.index.server/routes/package.scala
```diff
@@ -10,29 +10,33 @@ package object routes {
       (
         "q" ? "*",
         "page".as[Int] ? 1,
+        "total".as[Int] ? SearchParams.resultsPerPage,
```
romanowski, line 13: Could we also include information about available pages? Currently one needs to call next page until the response is empty (but this is probably not in the scope of this PR).

romanowski (approved):

bishabosha (approved):

server/src/main/scala/ch.epfl.scala.index.server/routes/package.scala
adpi2: I have added the `currentPage` and `pageTotal` fields: https://index-dev.scala-lang.org/api/search?q=*&scalaVersions=scala3&sort=stars&total=20&page=2

romanowski (approved):

server/src/main/scala/ch.epfl.scala.index.server/routes/api/SearchApi.scala
```diff
@@ -91,7 +104,8 @@ class SearchApi(
   val routes: Route =
     pathPrefix("api") {
       cors() {
-        path("search") {
+        // deprecated endpoint replaced by api/search endpoint
+        path("search-old") {
```
julienrf, line 108: Should we instead keep this one and introduce a new endpoint `api/search/v2`, to not break existing users?

PR #663, opened 2021-05-12
Manage elasticsearch server in sbt

adpi2:
- Move the elasticsearch container logic to the sbt build
- No need to distinguish between remote and local elasticsearch in the application config anymore
- Can run  server and data, run and test at the same time
- No need to wait for the container to stop between two runs: we now support ~reStart or ~test


project/Elasticsearch.scala
```diff
@@ -0,0 +1,84 @@
...
+  override lazy val globalSettings: Seq[Setting[_]] = Seq(
+    startElasticsearch := {
+      val esData = (ThisBuild / baseDirectory).value / ".esdata"
+      val logger = streams.value.log
+      if (alreadyStarted) {
+        logger.info("elasticsearch has already started")
+      } else {
+        logger.info("Trying to start elasticsearch container")
+        if (!esData.exists) IO.createDirectory(esData)
+        IO.setPermissions(esData, "rwxrwxrwx")
```
vincenzobaz, line 38: Should we use more strict permissions?

vincenzobaz (approved):

vincenzobaz:
LGTM

project/Elasticsearch.scala
adpi2: elasticsearch needs read and write access to it, but I don't know if I can change the container user

build.sbt
```diff
@@ -149,7 +148,18 @@ lazy val server = project
...
       .dependsOn(Assets / WebKeys.assets)
       .value,
-    reStart := reStart.dependsOn(Assets / WebKeys.assets).evaluated,
+    Test / test := {
+      startElasticsearch.value
+      (Test / test).value
+    },
+    reStart := {
+      startElasticsearch.value
+      reStart.evaluated
```
julienrf, line 157: Should we wait for elasticsearch to be ready before we invoke reStart?

adpi2: `startElasticsearch` only checks that the http GET request replies with OK. But then the application waits for elastic to be ready using  the health api.

julienrf: Here, `startElasticsearch` and `reStart` will be executed in parallel. So, I think it’s possible that sometimes the application tries to reach the health API before elasticsearch is even reachable. I think we should consider using `Def.sequential` here.

adpi2: Oh yes that's right. Thanks for catching it.

project/Elasticsearch.scala
adpi2: It also needs execution right

build.sbt
```diff
@@ -149,7 +148,9 @@ lazy val server = project
     Universal / packageBin := (Universal / packageBin)
       .dependsOn(Assets / WebKeys.assets)
       .value,
-    reStart := reStart.dependsOn(Assets / WebKeys.assets).evaluated,
+    Test / test := (Test / test).dependsOn(startElasticsearch).value,
+    reStart := reStart.dependsOn(startElasticsearch).evaluated,
```
rtyley, line 152: I think there was an unintended change here - `~server/reStart` in the sbt console no longer constructs assets, so the [dev instructions in the guide for running Scaladex locally](https://github.com/scalacenter/scaladex/blob/main/CONTRIBUTING.md#how-to-run-scaladex-locally) give an unstyled site, with CSS & JS failing to load:

image

...you might have to run `sbt clean` first to see this problem, if there are still assets lying around from an earlier run.

Issue #664, opened 2021-05-14
Scala 3.0.0 is not shown


julienrf:
![Screenshot from 2021-05-14 09-55-20](https://user-images.githubusercontent.com/332812/118240035-cbace100-b49a-11eb-8706-becf7dc7debd.png)

Releases for Scala 3.0.0-RC3 are shown but not for 3.0.0.

adpi2:
Fixed by Sonatype

PR #666, opened 2021-05-23
Attempt to fix issue #664: parse artifacts for Scala 3 final

rtyley:
See https://github.com/scalacenter/scaladex/issues/664 - Scala 3 was officially released on Friday 14th May 2021, with the 3.0.0 final release arriving on Maven Central on the prior day. Scala libraries built with Scala 3.0.0 final now have just the Scala major version number (ie `_3`) as their Scala language-version binary family qualifier, rather than the fully qualified version that was used pre-release (eg `_3.0.0-RC3`), or the major-minor version that was used with Scala 2 (eg `_2.13`).

So the maven artifact name (eg for an artifact called `scala-collection-plus`, at version 0.9) will now look like this:

```
scala-collection-plus_3-0.9.jar
```

Scaladex didn't have code to parse this new 'major-only' version string, so new artifacts released for Scala 3 final were not showing up in the Scaladex.

The fix in this commit should hopefully fix the issue going forward, though as an outside developer, I'm really not sure how updates get into Scaladex's elasticsearch instance, and I think there might have to be some kind of backfill? I only have the old snapshot in https://github.com/scalacenter/scaladex-small-index to go on, which was pre-Scala 3, so this fix was, to an extent, developed in the dark!

cc @julienrf

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
@@ -37,6 +37,7 @@ final case class ScalaVersion(version: BinaryVersion) extends LanguageVersion {
 final case class Scala3Version(version: BinaryVersion) extends LanguageVersion {
   def family = "scala3"
   def render: String = version match {
+    case MajorBinary(major) => s"scala $major"
     case MinorBinary(major, minor) if major == 0 && minor < 30 =>
       s"dotty $version"
     case PreReleaseBinary(major, _, _, _) if major == 3 => s"scala $toString"
--- a/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
@@ -66,9 +66,12 @@ object ScalaVersion {
 }
 
 object Scala3Version {
+  val `3`: Scala3Version = Scala3Version(MajorBinary(3))
+
   def isValid(version: BinaryVersion): Boolean =
     version match {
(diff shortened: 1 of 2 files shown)


adpi2 (approved):
Thanks for the help @rtyley

adpi2:
> The fix in this commit should hopefully fix the issue going forward, though as an outside developer, I'm really not sure how updates get into Scaladex's elasticsearch instance, and I think there might have to be some kind of backfill?

All new Scala pom files are pushed by Sonatype to Scaladex, which in turns save them to the file system so that we can reload them later if needed. So yes, in general there is a backfill. But right now Sonatype is not pushing any Scala 3 final artifact, since the name pattern has changed: "_3" instead of "_2.13". I am making contact with them to fix this.

> I only have the old snapshot in https://github.com/scalacenter/scaladex-small-index to go on, which was pre-Scala 3, so this fix was, to an extent, developed in the dark!

I have tested locally by publishing a Scala 3 final pom file. It works fine!

The request for publishing is:

rtyley:
> I have tested locally by publishing a Scala 3 final pom file. It works fine!
> 
> The request for publishing is:

Ah thank you! Following your instructions, I was able to do this locally as well (once I'd tweaked `hasPublishingAuthority` to allow me):

```bash
export POM_PATH=org/typelevel/cats-core_3/2.6.1/cats-core_3-2.6.1.pom

cd `mktemp -d`

wget https://repo1.maven.org/maven2/$POM_PATH -O example.pom

curl --data-binary "@example.pom" \
-XPUT \
--user token:$GITHUB_TOKEN \
"http://localhost:8080/publish?created=1478668532&readme=true&info=true&contributors=true&path=$POM_PATH"
```

PR #667, opened 2021-05-26
Download Scala 3 artifacts from Maven Central

adpi2:
This PR is a modification of the `data central` operation to download all Scala 3 artifacts published to Maven Central.

It works roughly this way:
- list all known organizations
- for each organization list all artifacts using https://search.maven.org API
- filter Scala 3 artifacts
- for each Scala 3 artifact download and save pom

One limitation is that all artifacts published with a new organization name are missed.

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/central/CentralMissing.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/central/CentralMissing.scala
@@ -16,17 +16,7 @@ import akka.stream.scaladsl._
 import ch.epfl.scala.index.data.maven.PomsReader
 import ch.epfl.scala.index.data.project.ArtifactMetaExtractor
 import ch.epfl.scala.index.model.misc.Sha1
-import ch.epfl.scala.index.model.release.{
-  Js,
-  MinorBinary,
-  Native,
-  Sbt,
-  SbtPlugin,
-  ScalaJs,
-  ScalaJvm,
-  ScalaNative,
-  ScalaVersion
-}
+import ch.epfl.scala.index.model.release._
 import de.heikoseeberger.akkahttpjson4s.Json4sSupport._
 import org.joda.time.DateTime
 import org.json4s._
@@ -50,9 +40,8 @@ object CentralMissing {
       v: String,
       timestamp: DateTime
   )
-  private[CentralMissing] case class ArtifactRequest(
-      groupId: String,
-      artifactId: String
+  private[CentralMissing] case class SearchRequest(
+      groupId: String
   )
 
   private[CentralMissing] case class DownloadRequest(
@@ -67,7 +56,7 @@ object CentralMissing {
     }
   }
 
-  case class PomContent(content: String)
+  case class PomContent(content: String, artifact: DownloadRequest)
 }
 
 object TimestampSerializer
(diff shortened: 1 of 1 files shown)


data/src/main/scala/ch.epfl.scala.index.data/central/CentralMissing.scala
```diff
@@ -93,56 +82,45 @@ class CentralMissing(paths: DataPaths)(implicit val system: ActorSystem) {
...
-  private def request(gaRequest: ArtifactRequest): HttpRequest = {
-    import gaRequest._
-
+  private def toHttp(request: SearchRequest): HttpRequest = {
     HttpRequest(
       uri = Uri("/solrsearch/select").withQuery(
         Query(
-          "q" -> s"""g:"$groupId" AND a:"$artifactId" """,
+          "q" -> s"""g:"${request.groupId}"""",
+          "rows" -> "20000",
```
julienrf, line 103: Should we also handle pagination for organizations having more than 20k artifacts?

data/src/main/scala/ch.epfl.scala.index.data/central/CentralMissing.scala
```diff
@@ -169,135 +145,97 @@ class CentralMissing(paths: DataPaths)(implicit val system: ActorSystem) {
...
-      ScalaNative(ScalaVersion.`2.11`, Native.`0.3`)
-    )
-
-    val releasesDownloads =
-      releases.flatMap { case (groupId, artifact) =>
-        allTargets.map(target =>
-          ArtifactRequest(groupId, artifact + target.encode)
-        )
-      }.toList
+    val releasesDownloads = allGroups.toList.map(SearchRequest(_))
```
julienrf, line 206: IIUC, we only perform requests to find Scala 3 artifacts belonging to an organization for which there are already releases indexed in scaladex?

I guess this is also what you mean by this comment in the description:

> One limitation is that all artifacts published with a new organization name are missed.

julienrf (approved):
This script looks good to me, but I wonder what we want to do with it in the long run. AFAIU, we want to run it once, and we expect the next releases to be pushed to us by Sonatype? Also, do we still want to keep a script that downloads all the Scala releases of Maven Central (regardless of their Scala binary version), in order to be able to re-compute the index?

data/src/main/scala/ch.epfl.scala.index.data/central/CentralMissing.scala
adpi2: Yes, it would be nice to handle pagination.

data/src/main/scala/ch.epfl.scala.index.data/central/CentralMissing.scala
adpi2: Exact

adpi2:
> AFAIU, we want to run it once, and we expect the next releases to be pushed to us by Sonatype?

So yes, I hope Sonatype will fix the issue with Scala 3 artifacts. Until then, we can run this script again. 

> Also, do we still want to keep a script that downloads all the Scala releases of Maven Central (regardless of their Scala binary version), in order to be able to re-compute the index?

I think it is useless to maintain different scripts, but this one can easily be adapted to some other need.

PR #668, opened 2021-05-26
Exclude Scala 3 milestone versions

adpi2:
I think it is time to exclude the Scala 3 milestone versions


julienrf (approved):
What happens to already indexed releases? Will they still show up on the website?

adpi2:
They will disappear from the website after I re-index everything. But the pom files stay on the disk.

PR #669, opened 2021-05-26
Fix binary version ordering

adpi2:
`Scala 3` should appear before `Scala 3.0.0-RC3`.
It entails that `MajorBinary(3) > PreReleaseBinary(3, 0, Some(0), ReleaseCandidate(3))`

More generally:
`MajorBinary(3) > MinorBinary(3, x) > PatchBinary(3, x, y) > PreReleaseBinary(3, x, y, z)`


julienrf (approved):
Nice improvement!

It would be valuable to add a test checking that the `Ordering` instance is symmetric and transitive. Would you mind adding that?

PR #670, opened 2021-05-28
Provide a 'Create Badge' function for users

rtyley:
This work follows on from https://github.com/scalacenter/scaladex/pull/660 - the markdown required to display a badge is fairly verbose:

[algebra-laws Scala version support](https://index.scala-lang.org/typelevel/algebra/algebra-laws)
```
[![algebra-laws Scala version support](https://index.scala-lang.org/typelevel/algebra/algebra-laws/latest-by-scala-version.svg)](https://index.scala-lang.org/typelevel/algebra/algebra-laws)
```
...it's nice to be able to help people to create it:

https://user-images.githubusercontent.com/52038/120015000-a8c21700-bfda-11eb-9590-4c55250d8b13.mp4

 Travis CI, GitHub and others follow this pattern of showing both the badge, and the recommended markdown to display it.

cc @adpi2

### DIFF
--- a/server/src/main/assets/css/partials/_main.scss
+++ b/server/src/main/assets/css/partials/_main.scss
@@ -647,55 +647,6 @@ main {
               padding: 6px 12px;
             }
         }
-        .overlay {
-            position: fixed;
-            z-index: 10;
-            top: 0;
-            bottom: 0;
-            left: 0;
-            right: 0;
-            background: rgba(0, 0, 0, 0.7);
-            transition: opacity 500ms;
-            visibility: hidden;
-            opacity: 0;
-        }
-        .overlay:target {
-            visibility: visible;
-            opacity: 1;
-        }
-
-        .popup {
-            margin: 70px auto;
-            padding: 20px;
-            background: #fff;
-            border-radius: 5px;
-            width: 60%;
-            position: relative;
-            transition: all 5s ease-in-out;
-        }
-
-        .popup h2 {
-            margin-top: 0;
-            color: #333;
-            font-family: Tahoma, Arial, sans-serif;
-        }
-        .popup .close {
-            position: absolute;
-            top: 20px;
-            right: 30px;
-            transition: all 200ms;
-            font-size: 30px;
-            font-weight: bold;
-            text-decoration: none;
-            color: #333;
-        }
-        .popup .close:hover {
-            color: #06D85F;
-        }
-        .popup .content {
-            max-height: 30%;
-            overflow: auto;
-        }
     }
     .artifact-project {
         margin-top: 10px;
@@ -726,6 +677,58 @@ main {
             display: block;
             margin: 5px;
         }
+    }
+}
+
+.badge-showcase {
+    .overlay {
+        position: fixed;
+        z-index: 10;
+        top: 0;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        background: rgba(0, 0, 0, 0.7);
+        transition: opacity 500ms;
+        visibility: hidden;
+        opacity: 0;
+    }
+    .overlay:target {
+        visibility: visible;
+        opacity: 1;
+    }
+
+    .popup {
+        margin: 70px auto;
+        padding: 20px;
+        background: #fff;
+        border-radius: 5px;
+        width: 60%;
+        position: relative;
+        transition: all 5s ease-in-out;
+    }
+
+    .popup h2 {
+        margin-top: 0;
+        color: #333;
+        font-family: Tahoma, Arial, sans-serif;
+    }
+    .popup .close {
+        position: absolute;
+        top: 20px;
+        right: 30px;
+        transition: all 200ms;
+        font-size: 30px;
+        font-weight: bold;
+        text-decoration: none;
+        color: #333;
+    }
+    .popup .close:hover {
+        color: #06D85F;
+    }
+    .popup .content {
+        max-height: 30%;
+        overflow: auto;
     }
 }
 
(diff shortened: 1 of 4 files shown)


server/src/main/assets/css/partials/_main.scss
```diff
@@ -643,6 +647,55 @@ main {
...
+            top: 0;
+            bottom: 0;
+            left: 0;
+            right: 0;
+            background: rgba(0, 0, 0, 0.7);
+            transition: opacity 500ms;
+            visibility: hidden;
+            opacity: 0;
+        }
+        .overlay:target {
```
rtyley, line 662: This is using the 'Pure-CSS lightbox' technique to create a modal overlay without any javascript:

https://developer.mozilla.org/en-US/docs/Web/CSS/:target#pure-css_lightbox

server/src/main/assets/css/partials/_main.scss
```diff
@@ -14,6 +14,10 @@ main {
     flex: 1;
 }
 
+.copyable-incantation {
+  user-select: all;
```
rtyley, line 18: This has the nice effect of the user selecting the entire text sample, rather just a word, when they click on it - at that point it's very easy to copy & paste the entire expression, so it's a nice complement (or possibly even replacement) to the Javascript 'Copy' button.

adpi2:
Thanks for this contribution, that's a really nice idea!

I have deployed this PR on the dev server for testing: [https://index-dev.scala-lang.org/](https://index-dev.scala-lang.org/)

A few remarks about the end result:
- I think the badge should be placed in the right side bar in its own box, before the installation box. That's because the badge can be quite verbose, also we may want to add the Scala.js and Scala Native badges too.
- It doesn't work when the main artifact is an sbt plugin: see for example [sbt-assembly](https://index-dev.scala-lang.org/sbt/sbt-assembly/sbt-assembly/0.15.0?target=_2.12_1.0)

server/src/main/assets/css/partials/_main.scss
adpi2: Nice trick!

server/src/main/assets/css/partials/_main.scss
adpi2: Really nice too!

rtyley:
Hi @adpi2 , thanks for you feedback!

### Handling non-JVM artifacts (Scala.JS, sbt, etc)

> * It doesn't work when the main artifact is an sbt plugin: see for example [sbt-assembly](https://index-dev.scala-lang.org/sbt/sbt-assembly/sbt-assembly/0.15.0?target=_2.12_1.0)

I've pushed commit 3e7498e, which hopefully goes some way to addressing this. The badge now uses the `targetType` that matches the selected `target` for the page - so if the user has selected https://index-dev.scala-lang.org/typelevel/cats/cats-core/2.6.1?target=_sjs1.x_3.x , the badge will be showing Scala.JS compatability, rather than the default JVM-based badge.

For `sbt` in particular, the badge displayed looks like this:

[sbt-assembly Scala version support-green.svg)](https://index.scala-lang.org/sbt/sbt-assembly/sbt-assembly)

..this isn't really ideal - for `sbt`, where `sbt` version fully determines the Scala version used, it would be better for the badge to simply display `sbt` versions, rather than Scala versions, like this:

[sbt-assembly Scala version support-green.svg)](https://index.scala-lang.org/sbt/sbt-assembly/sbt-assembly)

...that's probably a fix for a separate PR tho'.

### Badge placement

> * I think the badge should be placed in the right side bar in its own box, before the sbt/Maven installation box. That's because the badge can be quite verbose, also we may want to add the Scala.js and Scala Native badges too.

The good news is that it might not be necessary to display multiple badges at once, if the _"badge-targetType-matches-the-selected-target"_ tweak above is acceptable.

I wasn't sure if moving the badge to the RHS bar was better - the space is even more narrow, so most of the badge gets lost, and I couldn't find any clean CSS techniques to fade out the right -hand side of the badge to indicate it was truncated. As I wasn't certain about it, I've put that change into a [`badge-display-in-own-box-on-RHS-bar`](https://github.com/rtyley/scaladex/tree/badge-display-in-own-box-on-RHS-bar) branch for now, and you can see the screenshots of the two options below:

#### Badge placed in the Project Info

This does have the advantage of being a wide part of the layout, so we get to see a fair amount of the badge:
image

#### Badge placed in it's own box on the RHS bar

This is unfortunately quite a narrow space - we don't get to see much of the badge:
image

adpi2:
New changes deployed: https://index-dev.scala-lang.org/

## Handling non-JVM artifacts (Scala.JS, sbt, etc)

Looks good to me.

> ..this isn't really ideal - for sbt, where sbt version fully determines the Scala version used, it would be better for the badge to simply display sbt versions, rather than Scala versions

I agree. It can be done in another PR.

## Badge placement

I still think it should be placed on the RHS because:
- the library author is likely to put it in the readme as well
- it is kind of redundant with the versions drop down (when you choose a Scala version, it refreshes with the latest version)

According to me, this 'Create Badge' section is useful to copy-paste the markdown piece but it should not be placed in the project header, because it would add too much redundancy.

badges

> I wasn't sure if moving the badge to the RHS bar was better - the space is even more narrow, so most of the badge gets lost, and I couldn't find any clean CSS techniques to fade out the right -hand side of the badge to indicate it was truncated

Have you tried `overflow: scroll` as in the Installation section?

adpi2 (approved):

adpi2:
Thanks again @rtyley for your contribution! It's deployed in production: https://index.scala-lang.org/

I agree the current placement is not ideal. We, at the Scala Center team, are currenlty thinking of redesigning the project page in different tabs: project tab, and releases tab. With this redesign the version dropdowns will probably go to the release tab, and then we will be able to find a better place for the badge.

rtyley:
Thanks @adpi2 ! That's cool - I'm on holiday at the moment so wasn't able to respond properly to your feedback, but many thanks for releasing!

Issue #659, opened 2021-04-25
Scaladex badge that supplies info on supported Scala versions


rtyley:
As a user of Scala libraries, when I come to a project on GitHub I want to know:

*  what's **the latest version of Scala** that it supports, and **the corresponding artifact version**
* ...when I'm not on the latest version of Scala, what's the **latest available artifact version for that _legacy_ version of Scala**

As a _developer_ of Scala libraries, I know that keeping the information up to date in the the GitHub `ReadMe.md` is a bit of a drag. It would be lovely to have a modified badge that conveys all that information, and does so automatically.

### Current Scaladex badge

As detailed on the Scaladex [FAQ](https://github.com/scalacenter/scaladex/wiki/FAQ#how-to-include-a-scaladex-badge-in-my-projects-readme): ![Latest version](https://index.scala-lang.org/rtyley/scala-textmatching/scala-textmatching/latest.svg)
_(image url: https://index.scala-lang.org/playframework/play-json/play-json/latest.svg)_

### New badge format

  I'd like to suggest a modified larger badge that, beyond the latest artifact version, tells us the important Scala-version-dependent information, like this: ![Latest version](https://img.shields.io/badge/scala--textmatching-2.5%20(Scala%202.13%2C%202.12)%2C%202.4%20(Scala%202.11)%2C%202.1%20(Scala%202.10)-green?logo=scala&logoColor=DE3423)
_(image url: https://img.shields.io/badge/scala--textmatching-2.5%20(Scala%202.13%2C%202.12)%2C%202.4%20(Scala%202.11)%2C%202.1%20(Scala%202.10)-green?logo=scala&logoColor=DE3423)_

The goal is for the badge to be human readable, give salient but not comprehensive information, keeping the label as concise as possible!

Given an artifact id, minus the Scala version suffix:

1. Group all artifact versions by Scala binary version. If there are pre-release versions (M3, RC2) of a Scala Binary version, only  use the most recent one- and if a stable version is available, only list that.
2. Find the latest artifact version for each Scala binary version. Sort the artifact versions into descending order
3. Against each artifact version, list the Scala binary versions supported by it (again, sorted in descending order). If the Scala binary version is supported by a more recent artifact version, don't bother to list it against older ones.

### Non-JVM artifacts

If a project cross-builds to Scala.js or Scala Native, there's a lot more to explain, and the badge gets bigger. If, for this badge at least, the focus is the _Scala_ version, rather than supported runtime, we could still add in some runtime summary:

[...]

PR #674, opened 2021-07-04
Better support for sbt artifacts in the Scaladex badge

rtyley:
Each release of sbt uses a specific version of Scala (1.0 uses Scala 2.12, 0.13 uses Scala 2.10), so the sbt version of a plugin fully determines the Scala language binary version it uses - this means, for a user, it's only really interesting to know the sbt version of the plugin, not the Scala language version.

At the moment, the Scaladex badge for, eg, the `sbt-assembly` plugin reads:

[sbt-assembly Scala version support-green.svg)](https://index.scala-lang.org/sbt/sbt-assembly/sbt-assembly) (`1.0.0 (Scala 2.12, 2.10)`)

...for the user, it would be more helpful if it read:

[sbt-assembly Scala version support-green.svg)](https://index.scala-lang.org/sbt/sbt-assembly/sbt-assembly) (`1.0.0 (sbt 1.0, 0.13)`)

...listing the versions of sbt you can actually use it with! Although the Scaladex badge did have some logic to display the platform versions supported by _all_ listed Scala language versions, that doesn't help in this situation, as the Scala language version differs for each platform version of sbt. See also [this comment](https://github.com/scalacenter/scaladex/pull/670#issuecomment-853426504) from a prior PR.

This PR refactors the Scaladex badge logic to have two version-summary strategies:

* `SummariseLanguageVersion` - which was the existing strategy, listing notable Scala version support, secondarily adding in a summary of platform editions (eg 'Scala 2.13 - Native 0.4+0.3')
* `SummarisePlatformEdition` - the new strategy, which is used if *all* fetched `ScalaTarget`s for an artifact have a `targetType` for which the platform version fully determines the Scala version (as is the case for sbt). It omits the Scala version from the summary, and just relates the platform versions (eg `sbt 1.0, 0.13`)

The refactor is quite a substantial rearrangement of the code in `BadgesSupport.scala` unfortunately, and so the diff is quite hard to read - might be better to just read the updated final version. Tests in `BadgesSupportTest.scala` have been added/updated to check for the new sbt case.

cc @adpi2

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/BadgesSupport.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/BadgesSupport.scala
@@ -10,93 +10,203 @@ object BadgesSupport {
       allAvailableReleases: Seq[Release],
       specificArtifact: String,
       specificTargetType: ScalaTargetType
-  ): String = summaryOfLatestVersions((for {
-    release <- allAvailableReleases if release.isValid
-    ref = release.reference if ref.artifact == specificArtifact
-    scalaTarget <- ref.target if scalaTarget.targetType == specificTargetType
-  } yield ref.version -> scalaTarget).groupMap(_._1)(_._2))
-
-  def summaryOfLatestVersions(
-      scalaTargetsByArtifactVersion: Map[SemanticVersion, Seq[ScalaTarget]]
-  ): String = {
-    val allTargets = scalaTargetsByArtifactVersion.values.flatten.toSet
-
-    val latestScalaVersionsOfAvailableBinaryVersions =
-      allTargets
-        .map(_.languageVersion)
-        .groupBy(_.family)
-        .view
-        .mapValues(_.max)
-        .values
-        .toSet
-
-    val latestArtifactVersionByLanguageVersion = (for {
-      (artifactVersion, scalaTargets) <- scalaTargetsByArtifactVersion.toSeq
-      scalaTarget <- scalaTargets
-      if latestScalaVersionsOfAvailableBinaryVersions.contains(
-        scalaTarget.languageVersion
-      )
-    } yield scalaTarget.languageVersion -> artifactVersion)
+  ): String = summaryOfLatestVersions(
+    (for {
+      release <- allAvailableReleases if release.isValid
+      ref = release.reference if ref.artifact == specificArtifact
+      scalaTarget <- ref.target if scalaTarget.targetType == specificTargetType
+    } yield ref.version -> scalaTarget)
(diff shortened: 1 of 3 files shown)


server/src/main/scala/ch.epfl.scala.index.server/BadgesSupport.scala
```diff
@@ -10,93 +10,203 @@ object BadgesSupport {
...
+          .map(_.collect { case st: ScalaTargetWithPlatformBinaryVersion =>
+            st.platformEdition
+          })
+          .reduce(_ & _)
+
+      val targetsSummary: Option[String] =
+        Option.when(platformEditionsSupportedForAllLanguageVersions.nonEmpty)(
+          " - " + summarisePlatformTargets(
+            platformEditionsSupportedForAllLanguageVersions,
+            "+"
```
adpi2, line 178: Why using a different character in platforms compare to Scala versions?
I was surprised by that when looking at this badge:
image

adpi2 (approved):
Two minor comments, otherwise LGTM. Thanks for improving this!

server/src/main/scala/ch.epfl.scala.index.server/BadgesSupport.scala
rtyley: The `+` is current behaviour for the badge - but you don't actually get to see it...! I was thinking of an older version of the badge (that no longer exists!) that would show multiple platform types in the same badge, eg

`cats-core 2.6.1 (Scala 3, 2.13, 2.12 - JS 1.x+0.6, Native 0.4+0.3)`

I thought if the platform editions were separated by `,` rather than `+` that would become less clear - it would look like:

`cats-core 2.6.1 (Scala 3, 2.13, 2.12 - JS 1.x, 0.6, Native 0.4, 0.3)`

'Native' there seems to be mixed in with some random version numbers, so I thought `+` would at least group the platform versions by platform type.

However, thanks to [feedback](https://github.com/scalacenter/scaladex/issues/659#issuecomment-826381083) the badge now only ever shows one platform at most, so the situation only ever gets this bad:

`cats-core 2.6.1 (Scala 3, 2.13, 2.12 - JS 1.x, 0.6)`

...that does look ok to me, so I'l take out the custom separator parameter!

rtyley: That's fixed with 34949f3!

rtyley:
Thanks! Looks good in prod on, eg, https://index-dev.scala-lang.org/sbt/sbt-assembly/sbt-assembly#badge-markdown :

image

PR #675, opened 2021-07-08
rename modules

mlachkar:


infra/src/main/scala/ch/epfl/scala/utils/DoobieUtils.scala
```diff
@@ -0,0 +1,68 @@
...
+  object Fragments {
+    val empty = fr0""
+    val space = fr0" "
+
+    def buildInsert(
+        table: Fragment,
+        fields: Fragment,
+        values: Fragment
+    ): Fragment =
+      fr"INSERT INTO" ++ table ++ fr0" (" ++ fields ++ fr0") VALUES (" ++ values ++ fr0")"
```
vincenzobaz, line 55: Is it possible to use interpolation with `fr`?

build.sbt
```diff
@@ -189,4 +197,8 @@ lazy val data = project
     reStart / javaOptions ++= Seq("-Xmx4g")
   )
   .enablePlugins(BuildInfoPlugin, JavaAppPackaging)
-  .dependsOn(model, search)
+  .dependsOn(core, infra)
+
+lazy val V = new {
```
adpi2, line 202: Why `lazy val V = new {` rather than `object V {`?

build.sbt
```diff
@@ -189,4 +197,8 @@ lazy val data = project
...
+  val doobieVersion = "0.13.4"
```
adpi2, line 203: Would be better to group all the versions together. So I suggest to move the old ones here. https://github.com/scalacenter/scaladex/blob/9dc8ca0f6f92bf6c75625724c3ac660ebfa9558d/build.sbt#L6-L11

PR #676, opened 2021-07-08
Doobie

mlachkar:


core/src/main/scala/ch/epfl/scala/utils/Secret.scala
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.scala.utils
+
+case class Secret(private val value: String) extends AnyVal {
```
adpi2, line 3: I don't think we will need that class.

core/src/main/scala/ch/epfl/scala/services/ScaladexDb.scala
```diff
@@ -0,0 +1,8 @@
+package ch.epfl.scala.services
+
+import ch.epfl.scala.index.model.Project
+
+trait ScaladexDb {
```
adpi2, line 5:
```suggestion
trait Database {
```

core/src/main/scala/ch/epfl/scala/utils/Secret.scala
mlachkar: for password for the real database ?

core/src/main/scala/ch/epfl/scala/services/ScaladexDb.scala
mlachkar: ok

PR #684, opened 2021-07-13
Bootstrap the table projects

mlachkar:
- add postgresql container


core/src/main/scala/ch/epfl/scala/services/DatabaseApi.scala
```diff
@@ -0,0 +1,10 @@
+package ch.epfl.scala.services
+
+import scala.concurrent.Future
+
+import ch.epfl.scala.index.newModel.NewProject
+
+trait DatabaseApi {
+  def insertProject(p: NewProject): Future[NewProject]
+  def indexedProjects(): Future[Long]
```
adpi2, line 9:
```suggestion
  def countProjects(): Future[Long]
```

adpi2 (approved):

PR #687, opened 2021-08-05
Validate scala3 and 3.x.y versions

### LINKED ISSUE #654: search API cannot parse scalaVersion=scala3
https://index.scala-lang.org/api/search?target=JVM&scalaVersion=scala3&q=cats&page=1

vincenzobaz:
Fixes #654

### DIFF
--- a/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
+++ b/model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
@@ -37,10 +37,8 @@ final case class ScalaVersion(version: BinaryVersion) extends LanguageVersion {
 final case class Scala3Version(version: BinaryVersion) extends LanguageVersion {
   def family = "scala3"
   def render: String = version match {
-    case MajorBinary(major) => s"scala $major"
-    case MinorBinary(major, minor) if major == 0 && minor < 30 =>
-      s"dotty $version"
-    case PreReleaseBinary(major, _, _, _) if major == 3 => s"scala $toString"
+    case MajorBinary(3) => s"scala 3"
+    case PreReleaseBinary(3, _, _, _) => s"scala $toString"
     case _ => toString()
   }
   def isValid: Boolean = Scala3Version.isValid(version)
--- a/build.sbt
+++ b/build.sbt
@@ -108,6 +108,7 @@ lazy val template = project
 lazy val search = project
   .settings(commonSettings)
   .settings(
+    Test / test := (Test / test).dependsOn(startElasticsearch).value,
     libraryDependencies ++= Seq(
       "com.sksamuel.elastic4s" %% "elastic4s-client-esjava" % elastic4sVersion,
       "org.json4s" %% "json4s-native" % "3.6.9",
--- a/model/src/test/scala/ch.epfl.scala.index.model/release/LanguageVersionTests.scala
+++ b/model/src/test/scala/ch.epfl.scala.index.model/release/LanguageVersionTests.scala
@@ -1,0 +1,47 @@
+package ch.epfl.scala.index.model.release
+
+import org.scalatest.FunSpec
+import org.scalatest.Matchers
+
+class LanguageVersionTests extends FunSpec with Matchers {
+  describe("Scala3Version") {
+    it("should accept version 3") {
+      assert(Scala3Version.isValid(MajorBinary(3)))
+    }
+
+    it("should render version 3") {
+      assert(Scala3Version(MajorBinary(3)).render == "scala 3")
+    }
+
+    it("should not accept 3.x versions") {
+      assert(!Scala3Version.isValid(MinorBinary(3, 0)))
+    }
+
+    it("should accept 3.x.y versions") {
+      assert(!Scala3Version.isValid(PatchBinary(3, 0, 1)))
+    }
+  }
+
+  describe("parseScalaTarget") {
+    it("should not recognize scala3") {
+      val res = LanguageVersion.tryParse("scala3")
+      assert(res.isEmpty, res)
+    }
+
+    it("should recognize 3") {
+      val res = LanguageVersion.tryParse("3")
+      assert(res.nonEmpty, res)
+    }
+
+    it("should not recognize 3.0.1") {
+      val res = LanguageVersion.tryParse("3.0.1")
+      assert(res.isEmpty, res)
+    }
+
+    it("should recognize 3.0.1-RC1") {
+      val res = LanguageVersion.tryParse("3.0.1-RC1")
+      assert(res.isDefined, res)
+    }
+
+  }
+}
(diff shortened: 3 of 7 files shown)


model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
```diff
@@ -72,12 +72,16 @@ object Scala3Version {
...
+      case PatchBinary(3, _, _) => true
       case _ => false
     }
 
+  def FamilyParser[_: P]: P[Scala3Version] =
+    ("scala3" ~ End).map(_ => Scala3Version(MajorBinary(3)))
+
   def Parser[_: P]: P[Scala3Version] = {
-    BinaryVersion.Parser
+    FamilyParser | BinaryVersion.Parser
```
vincenzobaz, line 84: Accept binary version or family (`scala3`)

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
```diff
@@ -37,10 +37,10 @@ final case class ScalaVersion(version: BinaryVersion) extends LanguageVersion {
...
   def family = "scala3"
   def render: String = version match {
-    case MajorBinary(major) => s"scala $major"
-    case MinorBinary(major, minor) if major == 0 && minor < 30 =>
-      s"dotty $version"
-    case PreReleaseBinary(major, _, _, _) if major == 3 => s"scala $toString"
+    case MajorBinary(3) => s"scala 3"
+    case MinorBinary(0, minor) if minor < 30 => s"dotty $version"
+    case MinorBinary(3, _) => s"scala 3"
+    case PatchBinary(3, _, _) => s"scala 3"
```
vincenzobaz, line 43: IIUC for all versions 3.x.y we will only have `_3` artifacts so it makes sense to consider all versions simply as Scala 3

adpi2: Here `version` is the binary version so it is always supposed to be `Major(3)` except for pre-release version of 3.0.0 (eg. `PreRelease(3, 0, Some(0), "RC1")`.

So `case MinorBinary(3, _)` and `case PatchBinary(3, _, _)` should never happen.

Also `case PreReleaseBinary(major, _,  _, _)` is the old dotty case (eg. `0.27.0-RC1`) and is is already filtered out by `Scala3Version.isValid` method, so it should never happen.

 ```suggestion
    case MajorBinary(3) => s"scala 3"
```

File: model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala [hunk 3]
```diff
@@ -72,12 +73,17 @@ object Scala3Version {
   def isValid(version: BinaryVersion): Boolean =
     version match {
       case MajorBinary(3) => true
+      case MinorBinary(3, _) => true
+      case PatchBinary(3, _, _) => true
```
adpi2, line 77: Isn't it always supposed to be `Major(3)` ("_3")?

server/src/test/scala/ch.epfl.scala.index.server/routes/api/SearchApi.scala
```diff
@@ -0,0 +1,33 @@
+package ch.epfl.scala.index
+package server.routes.api
+
+import org.scalatest.FunSpec
+import org.scalatest.Matchers
+
+class SearchApiTest extends FunSpec with Matchers {
```
adpi2, line 7: This class and file should be called `SearchApiTests`.

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
adpi2: This is confusing because this `Parser` is also used to parse artifact names. So it means `cats-core_scala3` are valid Scala targets which is wrong.

I suggest to create a new `FamilyOrVersionParser` in `LanguageVersion`:

```scala
object LanguageVersion {
  def FamilyOrVersionParser[_: P]: P[LanguageVersion] = {
    (Scala3Version.FamilyParser | LanguageVersion.Parser) ~ End
  }
}
```

then use it in `SearchApi.parseScalaTarget`

adpi2 (changes requested):
It is not clear to me why you added `case MinorBinary(3, _)` and `case PatchBinary(3, _, _)` as valid Scala 3 binary versions.

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
vincenzobaz: Can we change the class constructor or put an assertion in it?
Having a `case _ => toString` looks like a hack not to have a exhaustivity check warning.
Maybe we could encode the fact that we are only expecting `Major(3)` into the types...

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
vincenzobaz: I don't know. This method is used to validate searches as well, so if I search for `3.0.1` binary artifacts I would like the validation to be true for all `3.x.y`

adpi2: Currently the `scalaVersion` field in the search API is the Scala binary version so it should be `2.13` or `3` rather than `2.13.5` or `3.0.1`.

vincenzobaz: Can we accept any version and internally transform it to binary version? I don't know how many Scala developers are familiar with this concept and its evolution in Scala 3.
We already try to hide this information in `sbt` with `%%`, it seems reasonable to accept any version here

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
adpi2: I agree about the `case _ => toString` but the `case PreRelease(3, 0, Some(0), _)` is still valid.

model/src/main/scala/ch.epfl.scala.index.model/release/LanguageVersion.scala
adpi2: I am not totally against it but I don't want to make this API more complex than it already is. In particular I think it is unfortunate that the fields of the search API are different form the fields of the search page and I think we should aim at merging them somehow. Your proposition goes in the opposite direction.

In the long run, I think we should design a new API that would be used by the front-end to load the different parts of the page asynchronously. Ideally this API should be as simple as possible (not smarter than what needs the scaladex website). Then we can also provide a nice library on top of it that can do the smart stuff (like computing the binary version).

vincenzobaz:
Thank you for your review @adpi2 

I appreciate your comment about making stuff easier. To that goal I can drop the search with `scala3` and `FamilyParser`.
I will use the search API with the correct binary version (I opened this PR to search Scala 3 artifacts).

We can open a new issue or project about unifying the two APIs and think about the access library.

adpi2 (approved):
Thank you for the fix and also for adding the tests!

PR #688, opened 2021-08-11
Insert GithubInfo into db

mlachkar:


infra/src/main/scala/ch/epfl/scala/utils/DoobieUtils.scala
```diff
@@ -71,4 +70,41 @@ object DoobieUtils {
...
+      )(_.map(toJson(_)).mkString(","))
+    }
+    implicit val githubIssuesMeta: Meta[List[GithubIssue]] = {
+      implicit val githubIssueDecoder: Decoder[GithubIssue] =
+        deriveDecoder[GithubIssue]
+      implicit val githubIssueEncoder: Encoder[GithubIssue] =
+        deriveEncoder[GithubIssue]
+      stringMeta.timap(
+        _.split(",").filter(_.nonEmpty).map(fromJson[GithubIssue](_).get).toList
+      )(_.map(toJson(_)).mkString(","))
```
adpi2, line 96: Why don't you use circe derivation to create an encoder and a decoder of `List[GithubIssue]`

PR #691, opened 2021-08-13
Add a table for Releases

mlachkar:


infra/src/main/scala/ch/epfl/scala/services/storage/sql/tables/ReleaseTable.scala
```diff
@@ -0,0 +1,50 @@
...
+    "released",
+    "licenses",
+    "isNonStandardLib"
+  )
+  private val tableFr: Fragment = Fragment.const0(table)
+  private val fieldsFr: Fragment = Fragment.const0(fields.mkString(", "))
+
+  private def values(r: NewRelease): Fragment =
+    fr0"${r.maven.groupId}, ${r.maven.artifactId}, ${r.version}, ${r.organization}, ${r.repository}," ++
+      fr0" ${r.artifact}, ${r.target}, ${r.description}, ${r.released}, ${r.licenses}, ${r.isNonStandardLib}"
```
adpi2, line 39: We could use the `Write[NewRelease]` instead of this, right?

adpi2 (approved):

infra/src/main/scala/ch/epfl/scala/services/storage/sql/tables/ReleaseTable.scala
mlachkar: I think so! thanks

mlachkar: next PR

PR #695, opened 2021-08-16
Upgrade scalatest to 3.2.9

mlachkar:
in the develop branch we upgraded scalatest. To avoid conflicts everytime we add new tests in main, I updated scalatest in main too.


adpi2 (approved):
Thanks a lot! Did you use scalafix to update all the Spec classes?

mlachkar:
forgot about the scalafix Rules.. I fixed it by hand.

PR #697, opened 2021-08-18
Add a tables for dependencies

mlachkar:


infra/src/main/scala/ch/epfl/scala/services/storage/sql/tables/DependenciesTable.scala
```diff
@@ -0,0 +1,39 @@
...
+    "target_artifactId",
+    "target_version",
+    "scope"
+  )
+  private val tableFr: Fragment = Fragment.const0(table)
+  private val fieldsFr: Fragment = Fragment.const0(fields.mkString(", "))
+
+  private def values(d: NewDependency): Fragment =
+    fr0"${d.source.groupId}, ${d.source.artifactId} ,${d.source.version}, ${d.target.groupId}," ++
+      fr0" ${d.target.artifactId}, ${d.target.version}, ${d.scope}"
```
adpi2, line 29: Since we have a `Writer[NewDependency]` this method should not be needed.

PR #698, opened 2021-08-18
 PublishApi publish now in the db

mlachkar:


data/src/main/scala/ch.epfl.scala.index.data/DataPaths.scala
```diff
@@ -55,7 +55,7 @@ scaladex-credentials (optionnal)
 └── search-credential
  */
 
-sealed trait LocalRepository
+sealed trait LocalRepository extends Product with Serializable
```
adpi2, line 58: Why do we need that?

data/src/main/scala/ch.epfl.scala.index.data/project/ProjectConvert.scala
```diff
@@ -213,6 +212,55 @@ class ProjectConvert(paths: DataPaths, githubDownload: GithubDownload)
     }
     (projectAndReleases, allDependencies)
   }
+
+  private def getDependencies(pom: ReleaseModel): List[NewDependency] =
+    pom.dependencies
+      .map(dep => NewDependency(pom.mavenRef, dep.mavenRef, dep.scope))
+      .distinct
+
+  def convertOne(
```
adpi2, line 221: Nice!

infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
```diff
@@ -1,37 +1,56 @@
...
+      projectRef: Project.Reference
+  ): Future[Option[NewProject]] =
+    for {
+      project <- run(ProjectTable.selectOne(projectRef.org, projectRef.repo))
+      userForm <- run(
+        ProjectUserFormTable.selectOne(projectRef.org, projectRef.repo)
+      )
+      githubInfoTable <- project
+        .map(p => run(GithubInfoTable.selectOne(p.organization, p.repository)))
+        .getOrElse(Future.successful(None))
```
adpi2, line 45: If this is too slow we can make it faster by running the 3 futures in parallel. (later)

server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/IndexingActor.scala
```diff
@@ -78,95 +77,31 @@ class IndexingActor(
...
+        data.created,
+        repo,
+        project
+      )
+      _ = converted
+        .map { case (p, r, d) =>
+          for {
+            _ <- db.insertProject(p)
+            _ <- db.insertReleases(Seq(r))
+            _ <- db.insertDependencies(d)
```
adpi2, line 98: Here we can also save some time by running the 3 futures in parallel.
Also it would be good to know which request failed if it failed.

server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/IndexingActor.scala
```diff
@@ -12,20 +12,19 @@ import akka.actor.Actor
...
-import ch.epfl.scala.index.model.Release
 import ch.epfl.scala.index.model.misc.GithubRepo
-import ch.epfl.scala.index.newModel.NewDependency
 import ch.epfl.scala.index.search.ESRepo
+import ch.epfl.scala.services.DatabaseApi
 import org.slf4j.LoggerFactory
 
 class IndexingActor(
     paths: DataPaths,
     dataRepository: ESRepo,
```
adpi2, line 26: Do we still need it?

adpi2 (approved):

PR #702, opened 2021-08-23
Thread pool

mlachkar:
The thread size is quiet arbitrary: I am using the value in doobie example. 
Otherwise Zalando wrote this article about thread pool sizing.
https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html

### DIFF
--- a/infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
+++ b/infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
@@ -15,10 +15,11 @@ import ch.epfl.scala.services.storage.sql.tables.ProjectTable
 import ch.epfl.scala.services.storage.sql.tables.ProjectUserFormTable
 import ch.epfl.scala.services.storage.sql.tables.ReleaseTable
 import ch.epfl.scala.utils.DoobieUtils
+import ch.epfl.scala.utils.ScalaExtensions.TraversableOnceFutureExtension
 import doobie.implicits._
 
-class SqlRepo(conf: DbConf) extends DatabaseApi {
-  private[sql] val (xa, flyway) = DoobieUtils.create(conf)
+class SqlRepo(conf: DbConf, xa: doobie.Transactor[IO]) extends DatabaseApi {
+  private[sql] val flyway = DoobieUtils.flyway(conf)
   def createTables(): IO[Unit] = IO(flyway.migrate())
   def dropTables(): IO[Unit] = IO(flyway.clean())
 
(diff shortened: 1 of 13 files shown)


adpi2 (approved):
I don't understand much in this PR but I trust you :)

infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
```diff
@@ -97,3 +107,6 @@ class SqlRepo(conf: DbConf) extends DatabaseApi {
   private def run[A](v: doobie.ConnectionIO[A]): Future[A] =
     v.transact(xa).unsafeToFuture()
 }
+object SqlRepo {
+  val sizeOfInsertMany = 10000
```
julienrf, line 111: Does that mean that we might attempt to concurrently insert 10k entries?

julienrf (commented):
It is not clear to me either what would be the shape of the threads, looking at the source code. Could you please summarize how many “thread pools” this is using, what are their configurations, and in which situation they are used?

infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
```diff
@@ -64,13 +65,22 @@ class SqlRepo(conf: DbConf) extends DatabaseApi {
...
   def insertRelease(release: NewRelease): Future[NewRelease] =
     run(ReleaseTable.insert, release)
 
-  override def insertDependencies(deps: Seq[NewDependency]): Future[Int] =
-    run(DependenciesTable.insertMany(deps))
+  override def insertDependencies(deps: Seq[NewDependency]): Future[Int] = {
+    deps
+      .grouped(SqlRepo.sizeOfInsertMany)
+      .map(d => run(DependenciesTable.insertMany(d)))
+      .sequence
```
julienrf, line 81: IIUC, this is going to insert all the content of `deps` in parallel.

I guess your intention was to insert by batches of the size `sizeOfInsertMany`, instead? I don’t think the call to `grouped` helps here, because anyway the call to `map` won’t “wait” that the previous group finished being inserted before inserting the next one.

I think here the parallelism level (of the database operations) won’t be defined by `sizeOfInsertMany`, but by the size of the thread pool used by doobie in `run`.

infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
adpi2: That means an insertMany request cannot be bigger than 10k entries. It is orthogonal to the size of the thread pool which control the level of concurrency.

infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
julienrf: Maybe this is not important if that works, but I insist on the fact that `grouped(sizeOfInsertMany)` has no effect here. The calls to `run` will all happen concurrently.

mlachkar: SQL allows to insert many rows at the same time using one connexion. 
in fact the size of the thread pool used by doobie controls parallelism , which 32. 
So we insert 10 000 rows in each connexion using a thread pool size of 32

adpi2: > Maybe this is not important if that works, but I insist on the fact that grouped(sizeOfInsertMany) has no effect here.

The `grouped(sizeOfInsertMany)` creates n request of 10000 entries. The n requests are submitted simultaneously. That's what we want.

julienrf: Ah, thank you for clarifying, I understand my mistake now! Indeed, there is a single SQL `INSERT` with 10k rows on every call to `run`, but all the calls to `run` happen concurrently (I thought we were inserting the rows one by one in the `insertMany` but this is not the case!).

Sorry for bringing confusion!

PR #705, opened 2021-08-30
Project page

mlachkar:


core/src/main/scala/ch.epfl.scala.index.model/DataPaths.scala
```diff
@@ -1,11 +1,9 @@
-package ch.epfl.scala.index
-package data
+package ch.epfl.scala.index.model
```
adpi2, line 1: Why did you move it to `model`? In the future, I would expect it to be goes to `infra`?

core/src/main/scala/ch/epfl/scala/index/newModel/NewProject.scala
```diff
@@ -11,42 +11,43 @@ case class NewProject(
...
       Organization(org),
       repository = Repository(repo),
       githubInfo = githubInfo,
       esId = None,
-      formData = formData
+      dataForm = formData
     )
 
-  case class FormData(
+  case class DataForm(
```
adpi2, line 41: Not sure if it's a better name. Maybe we should call them `ProjectSettings` because that's the thing the project owner can modify on its project.

core/src/main/scala/ch/epfl/scala/services/LocalStorageApi.scala
```diff
@@ -0,0 +1,10 @@
+package ch.epfl.scala.services
+
+import scala.concurrent.Future
+
+import ch.epfl.scala.index.model.Project
+import ch.epfl.scala.index.model.ProjectForm
+
+trait LocalStorageApi {
+  def saveDataForm(ref: Project.Reference, userData: ProjectForm): Future[Unit]
+}
```
adpi2, line 10: Nice to have those interfaces

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
```diff
@@ -126,6 +129,80 @@ class GithubDownload(
       )
     )
   }
+  def update(
+      project: Project,
+      projectForm: ProjectForm,
+      fromStored: Boolean = false
+  ): Project = {
```
adpi2, line 136: This is not the right place because it does not download anything from github or reading/validating the github json from the filesystem.

I would put it in `Project` instead.

template/src/main/twirl/ch.epfl.scala.index.views/search/resultList.scala.html
```diff
@@ -12,7 +11,7 @@
         <div class="col-md-8">
           @if(params.contributingSearch){
             <div class="contributing-result">
-              @contributingInfo(project, params = Some(params))
+              @contributingInfo(???, params = Some(params))
```
adpi2, line 14: ???

core/src/main/scala/ch.epfl.scala.index.model/DataPaths.scala
mlachkar: make sense, it should be next to ESRepo.

data/src/main/scala/ch.epfl.scala.index.data/github/GithubDownload.scala
mlachkar: we agreed to remove the update from github and do it later

core/src/main/scala/ch/epfl/scala/index/newModel/NewProject.scala
mlachkar: in another commit. It needs a bigger change to include database rename

template/src/main/twirl/ch.epfl.scala.index.views/search/resultList.scala.html
mlachkar: done

template/src/main/twirl/ch.epfl.scala.index.views/search/resultList.scala.html
```diff
@@ -1,7 +1,7 @@
 @import ch.epfl.scala.index.views.html._
```
adpi2, line 1: not used

adpi2 (approved):

template/src/main/twirl/ch.epfl.scala.index.views/project/badges.scala.html
```diff
@@ -13,14 +16,14 @@ <h2>Create badge</h2>
...
-        <img src="@release.reference.badgeUrl" />
+        <img src="@release.badgeUrl(env)" />
         <div>
           <h5>Markdown:</h5>
           <pre
           aria-label="Badge markdown"
           id="copy-badge-markdown"
           class="copyable-incantation"
-          >[![@release.reference.artifact Scala version support](@release.reference.badgeUrl)](@release.reference.artifactFullHttpUrl)</pre>
+          >[![@release.artifactName Scala version support](@release.badgeUrl(env)](@release.artifactFullHttpUrl(env)</pre>
```
rtyley, line 26: Two small typos introduced here as only the opening bracket is introduced with `(env`! Fix in https://github.com/scalacenter/scaladex/pull/786

PR #706, opened 2021-08-30
Make system shutdown captured by scaladex server

mlachkar:
- when running server/run, previously, the system shutdown didn't stop the server.
It's now fixed.
- Also added in this PR configuration to configure the endpoint and the port


project/Deployment.scala
```diff
@@ -128,7 +128,7 @@ class Deployment(
           |  -Dconfig.file=/home/$userName/scaladex-credentials/application.conf \\
           |  -Dsentry.dsn=$sentryDsn \\
           |  -Dsentry.release=$version \\
-          |  $port \\
+          |  -Dapi.port=$port \\
```
adpi2, line 131: If it is in config it should better go to the app.conf file

PR #708, opened 2021-09-01
Make datapath configurable through application.conf

mlachkar:
- make prostgresql write locally in .postgresql directory:
it will allow not to insert data everytime locally

### DIFF
--- a/data/src/main/scala/ch.epfl.scala.index.data/DataPaths.scala
+++ b/data/src/main/scala/ch.epfl.scala.index.data/DataPaths.scala
@@ -5,6 +5,7 @@ import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
 
+import ch.epfl.scala.utils.Env
 import org.slf4j.LoggerFactory
 
 /*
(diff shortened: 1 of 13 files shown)


project/Deployment.scala
```diff
@@ -128,10 +128,11 @@ class Deployment(
...
           |  -Dsentry.release=$version \\
-          |  -Dapi.port=$port \\
-          |  /home/$userName/scaladex-contrib \\
-          |  /home/$userName/scaladex-index \\
-          |  /home/$userName/scaladex-credentials \\
+          |  -Dapp.port=$port \\
+          |  -Dapp.env=prod \\
+          |  -Ddata-paths.contrib=/home/$userName/scaladex-contrib \\
+          |  -Ddata-paths.index=/home/$userName/scaladex-index \\
+          |  -Ddata-paths.credentials=/home/$userName/scaladex-credentials \\
```
adpi2, line 135: Can we remove that and put them in `~/scaladex-credentials/app.conf` ?

data/src/main/scala/ch.epfl.scala.index.data/DataPaths.scala
```diff
@@ -70,48 +71,42 @@ object LocalPomRepository {
 
 object DataPaths {
 
-  private val log = LoggerFactory.getLogger(getClass)
-
   private val base = build.info.BuildInfo.baseDirectory.toPath.getParent
```
adpi2, line 74: Do we still need that?

data/src/main/scala/ch.epfl.scala.index.data/DataPaths.scala
```diff
@@ -70,48 +71,42 @@ object LocalPomRepository {
...
-          Paths.get(index),
-          Paths.get(credentials)
-        )
-      case _ => {
-        (
-          defaultContrib,
-          defaultIndex,
-          defaultCredentials
-        )
+  base.resolve(Paths.get("scaladex-credentials"))
```
adpi2, line 75: and that?

server/src/main/resources/reference.conf
```diff
@@ -10,11 +10,18 @@ org.scala_lang.index.server {
...
+app {
+   env = "local"
    endpoint = "localhost"
    port = 8082
 }
 
+data-paths {
+  contrib = "scaladex-contrib"
+  index = "scaladex-small-index"
+  credentials = "scaladex-credentials"
```
adpi2, line 22: locally we were using `../scaladex-contrib`?

project/Deployment.scala
mlachkar: yes! next step . just to make this release not dependent on anything

PR #711, opened 2021-09-14
Add pathPrefix for Assets

mlachkar:
- change the path for jQuery from
/assets/lib/jQuery/dist/jquery.min.js to /assets/lib/jquery/dist/jquery.min.js

the reason is that locally on macos (at least), the path is case sensitive. It was tested on the prod
server and the new path is found.

### DIFF
--- a/server/src/main/resources/reference.conf
+++ b/server/src/main/resources/reference.conf
@@ -13,7 +13,7 @@ org.scala_lang.index.server {
 app {
    env = "local"
    endpoint = "localhost"
-   port = 8082
+   port = 8080
 }
 
 data-paths {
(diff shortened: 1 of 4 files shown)


server/src/main/resources/reference.conf
```diff
@@ -13,7 +13,7 @@ org.scala_lang.index.server {
 app {
    env = "local"
    endpoint = "localhost"
-   port = 8082
+   port = 8080
```
adpi2, line 16: what's the reason for modifying it?

adpi2 (approved):

server/src/main/resources/reference.conf
mlachkar: authentication always redirect to 8080/callback... 
I don't know how the port is passed, I think it's some default one

adpi2: so this setting is not used anywhere?

PR #712, opened 2021-09-16
Refactor actors to use akka-typed and no blocking

vincenzobaz:


vincenzobaz:
We need tests to verify that this kind of refactor does not impact functionality.

PR #717, opened 2021-09-24
Update the frontpage to use the databaseApi

mlachkar:
- getAllTopics


core/src/main/scala/ch.epfl.scala.index.model/Release.scala
```diff
@@ -55,11 +52,11 @@ object Release {
       repository: String,
       artifact: String,
       version: SemanticVersion,
-      target: Option[ScalaTarget]
+      target: Platform
```
adpi2, line 55: Sweet!

core/src/main/scala/ch.epfl.scala.index.model/release/Platform.scala
```diff
@@ -0,0 +1,204 @@
...
+    def isValid(version: BinaryVersion): Boolean =
+      stableBinaryVersions.contains(version)
+  }
+
+  case object Java extends Platform {
+    override def scalaVersion: Option[ScalaLanguageVersion] = None
+    override def platformVersion: Option[BinaryVersion] = None
+    override def render: String = "Java"
+    override def encode: String = ""
+    override def shortName: String = "java"
```
adpi2, line 94: should be `Java`

server/src/main/scala/ch.epfl.scala.index.server/routes/FrontPage.scala
```diff
@@ -34,12 +37,27 @@ class FrontPage(dataRepository: ESRepo, session: GithubUserSession)(implicit
...
+      allPlatforms <- allPlatformsF
+      platformTypeWithCount = FrontPage.getPlatformTypeWithCount(allPlatforms)
+      scalaFamilyWithCount = FrontPage.getScalaLanguageVersionWithCount(
+        allPlatforms
+      )
+      scalaJsVersions = FrontPage
+        .getPlatformWithCount(allPlatforms) { case p: Platform.ScalaJs =>
+          p.scalaJsV
+        }
+        .sorted
```
adpi2, line 50: This `sorted` is repeated 5 times. It could be put in `getPlatformWithCount` instead.

server/src/test/scala/ch.epfl.scala.index.server/routes/FrontPageTests.scala
```diff
@@ -0,0 +1,68 @@
...
+      ),
+      ref2 -> Set(
+        Platform.ScalaNative(`2.13`, nat03),
+        Platform.ScalaJvm(`2.13`),
+        Platform.ScalaJvm(`2.12`),
+        Platform.Java,
+        `3.0.0-RC3`,
+        `3.0.0-RC1`
+      )
+    )
```
adpi2, line 43: All these could be put in `Values` and be used in other test classes, in `PlatformTests` or `BadgesSupportTest` for instance.

template/src/main/twirl/ch.epfl.scala.index.views/frontpage.scala.html
```diff
@@ -1,14 +1,16 @@
...
-  scalaJsVersions: List[(String, Long)],
-  scalaNativeVersions: List[(String, Long)],
-  sbtVersions: List[(String, Long)], 
+@import ch.epfl.scala.index.model.release.BinaryVersion
+@(topics: List[(String, Int)],
+  platformWithCount: List[(Platform.PlatformType, Int)],
+  scalaVersionsWithCount: List[(String, Int)],
+  scalaJsVersions: List[(BinaryVersion, Int)],
+  scalaNativeVersions: List[(BinaryVersion, Int)],
+  sbtVersions: List[(BinaryVersion, Int)],
```
adpi2, line 13: Types :heart:

adpi2 (changes requested):
Overall this looks very good! Thanks again for refactoring the `Platform` model, it looks simpler now. I made some suggestions and I think we should add and fix the proposed test in Badges.

server/src/main/scala/ch.epfl.scala.index.server/BadgesSupport.scala
```diff
@@ -191,21 +192,23 @@ object BadgesSupport {
...
     override def summarise(
-        scalaTargets: Set[ScalaTargetWithPlatformBinaryVersion],
-        interestingKeyVersions: Set[PlatformEdition]
+        scalaTargets: Set[Platform],
+        interestingKeyVersions: Set[BinaryVersion],
+        platformType: Platform.PlatformType
     ): String =
       summarisePlatformTargets(
-        scalaTargets.map(_.platformEdition).filter(interestingKeyVersions)
+        "sbt",
```
adpi2, line 210: Should be `platformType.toString`

PR #718, opened 2021-09-27
Rework containers

adpi2:
In order of commits:

- [Load database config in infra](https://github.com/scalacenter/scaladex/pull/718/commits/b2c8a68b872034cfbd83b6e4cb276ec21d6755e5):
To run the database tests in infra we need to load the database config.
- [Rework Elasticsearch and Postgres containers](https://github.com/scalacenter/scaladex/pull/718/commits/b0639743a6324d529c38adb04bfdacf4d5cb54b3):
We separate the task that starts the elasticsearch container and the task that start the postgres container. The postgres database name is now configurable. We don't use fixed port anymore so that we are sure the container can start on some available port. The port are passed to the application using java properties: e.g. `-Ddatabase.port=12345`.
- [Rework infra tests](https://github.com/scalacenter/scaladex/pull/718/commits/27751db6f70e165b8d499c8f3cf0b7eceaea420f)
  - Moved the creation of the `SqlRepo` instance out of `Values` to a new `BaseDatabaseSuite`.
  - Grouped the values in `Values` by project in object. Example: `Values.Cats.project`, `Values.Cats.core`...
  - Made `SqlRepoTests` tests asynchronous (by not using `Await.result`)
  - Made `EsRepoTests` run on an empty elasticsearch index by adding new documents in each tests
- [Move RelevanceTest to IntegrationTest](https://github.com/scalacenter/scaladex/pull/718/commits/1d98fc04c3fb2d6d1537664b16cb24e31585a7af):
`SeedElasticsearch` is now run inside the test directly.


infra/src/main/scala/ch/epfl/scala/services/storage/sql/SqlRepo.scala
```diff
@@ -143,13 +146,20 @@ class SqlRepo(conf: DatabaseConfig, xa: doobie.Transactor[IO])
...
   // to use only when inserting one element or updating one element
   // when expecting a row to be modified
-  private def run[A](i: A => doobie.Update0, v: A): Future[A] =
-    i(v).run.transact(xa).unsafeToFuture.flatMap {
-      case 1 => Future.successful(v)
-      case code =>
-        Future.failed(new Exception(s"Failed to insert $v (code: $code)"))
+  private def strictRun(
+      update: doobie.Update0
+  )(expectedRows: Int): Future[Unit] =
```
mlachkar, line 151: maybe a default value to 1!

This run was meant to be used only by insert, since when inserting a row, we expect to have only 1 modified line, but by mistake I used it once when doing insertOrUpdate and it was failing when we don't update anything. 
Did you change it because of something similar ?

infra/src/test/scala/ch/epfl/scala/services/storage/sql/SqlRepoTests.scala
```diff
@@ -1,91 +1,86 @@
...
+import scala.concurrent.ExecutionContextExecutor
 
-class SqlRepoTests
-    extends AsyncFunSpec
-    with Matchers
-    with IOChecker
-    with BeforeAndAfterAll {
-  private val db = Values.db
-  val transactor = Values.xa
+class SqlRepoTests extends AsyncFunSpec with BaseDatabaseSuite with Matchers {
```
mlachkar, line 10: nice to use AsyncFunSpec ..

infra/src/test/scala/ch/epfl/scala/services/storage/sql/SqlRepoTests.scala
```diff
@@ -1,91 +1,86 @@
...
 
-  override def beforeAll(): Unit = db.migrate().unsafeRunSync()
+  // TODO: use a fixed pool execution context
```
mlachkar, line 12: for tests I don't think it's an issue!

data/src/main/scala/ch.epfl.scala.index.data/elastic/SeedElasticSearch.scala
```diff
@@ -1,34 +1,108 @@
...
+      )
+      _ = logFailures(
+        insertedAnProject,
+        (p: NewProject) => p.reference.toString,
+        "projects"
+      )
+      _ = logger.info("inserting releases to database")
+      insertedReleases <- db.insertReleasesWithFailures(
+        releases.map(NewRelease.from)
+      )
```
mlachkar, line 61: We love for comprehension !

mlachkar (approved):
except maybe the discussion we had about the configuration and where "it should be", everything looks nice! 
Maybe we can merge like that, we just need to make sure that everything is working (I mean main and Server), but otherwise I am ok!

infra/src/test/scala/ch/epfl/scala/services/storage/sql/SqlRepoTests.scala
adpi2: As long as one has more than one core it should work. But it is always good to know exactly how much cores are needed.

data/src/main/scala/ch.epfl.scala.index.data/elastic/SeedElasticSearch.scala
adpi2: :)

PR #722, opened 2021-10-07
Offer a platform specific badge

vincenzobaz:


adpi2:
It's even better if you can reuse the existing tests and remove the old code.

vincenzobaz:
`latestByScalaVersion` accepts an optional `targetType` which my endpoint requires.
Should I assume `JVM` as it is done there?

rtyley:
👋    Hi there! I just spotted this PR today in passing, and as I was the original contributor of the `latest-by-scala-version.svg` badge with PR https://github.com/scalacenter/scaladex/pull/660 (which this PR is [replacing](https://github.com/scalacenter/scaladex/pull/722#issuecomment-938672629), right?) I guess I'm susceptible to woeful feelings of personal-ownership, etc, which I must try to cope with :)

Regardless of that, I guess I would be curious to know how the proposed badge text improves on the old badge text? I don't see any examples of the new badge text in the PR description/comments, so it's hard to compare. For instance, for these badges, what would the proposed new text be?

* [cats-core Scala version support,_2.0.0_(Scala_2.11),_1.2.0_(Scala_2.10)-green.svg?)](https://index.scala-lang.org/typelevel/cats/cats-core)
* [sbt-sonatype Scala version support](https://index.scala-lang.org/xerial/sbt-sonatype/sbt-sonatype)

The use-case for the original `latest-by-scala-version.svg` badge, as described in https://github.com/scalacenter/scaladex/issues/659 were:

> As a user of Scala libraries, when I come to a project on GitHub I want to know:
> 
> *  what's **the latest version of Scala** that it supports, and **the corresponding artifact version**
> * ...when I'm not on the latest version of Scala, what's the **latest available artifact version for that _legacy_ version of Scala**

...that's definitely a use-case that everyone recognises! I think it would be unfortunate if support for those particular needs goes away (especially as users may have installed the badge for precisely that functionality), but if the new functionality in this PR supports them, then that's great 👍

adpi2:
@rtyley thank you for opening this discussion. It is true that we did not communicate our intention here and I would like to apologize for it.

The current `latest-by-scala-version.svg` badge is great in many aspects and we would like to keep it. But there are 2 problems we are trying to solve:
- some badges are large and they will grow larger over time with new Scala versions and platform versions
ex: cats-core Scala version support
- in some rare case a badge does not display the platform version: https://github.com/scalacenter/scaladex/blob/97c997e2b2c633caa0c283f2da8117494c97ba6b/server/src/test/scala/ch.epfl.scala.index.server/BadgesSupportTest.scala#L155-L167

We can solve these two issues by adding a platform version parameter, in addition to the platform type parameter:
- `typelevel/cats/cats-core/latest-by-scala-version.svg?targetType=Js` will only show the latest available Scala.js version:
-green.svg?)
- we can also specify the platform version with:
`typelevel/cats/cats-core/latest-by-scala-version.svg?targetType=Js&platformVersion=0.6`
,_1.2.0_(Scala_2.10)-green.svg?)
- `https://index.scala-lang.org/typelevel/cats/cats-core` will still choose the Jvm as the default platform
[cats-core Scala version support,_2.0.0_(Scala_2.11),_1.2.0_(Scala_2.10)-green.svg?)](https://index.scala-lang.org/typelevel/cats/cats-core)
- sbt must be treated as an exception so that `https://index.scala-lang.org/xerial/sbt-sonatype/sbt-sonatype/latest-by-scala-version.svg?targetType=Sbt` shows
[sbt-sonatype Scala version support,_2.3_(sbt_0.13)-green.svg?)](https://index.scala-lang.org/xerial/sbt-sonatype/sbt-sonatype)

So there will be 6 badges (Jvm, Js 1.x, Js 0.6, Native 0.4, Native 0.3, sbt) instead of 4 (Jvm, Js, Native, sbt) and it will be the responsability of the library maintainer to decide which platform versions to show and which not to show.

In return  the badges will be more concise and the code will  be slightly more simple.

adpi2:
Superseded by #893

PR #736, opened 2021-11-08
Retrieve newest projects

mlachkar:


core/src/main/scala/ch/epfl/scala/index/newModel/NewProject.scala
```diff
@@ -63,6 +66,7 @@ object NewProject {
       Organization(org),
       repository = Repository(repo),
       githubInfo = githubInfo,
+      created = Some(Instant.now),
```
adpi2, line 69: I would prefer to say `None` that `Instant.now` which is not right.

data/src/main/scala/ch.epfl.scala.index.data/project/PomMeta.scala
```diff
@@ -120,12 +122,12 @@ object PomMeta {
...
   def from(
       pom: ReleaseModel,
-      created: DateTime,
+      created: Instant,
       localRepository: LocalRepository,
       paths: DataPaths,
       sha1: String
   ): Option[PomMeta] = {
-    val createdString = PomMeta.format.print(created) // todo: keep Datetime
+    val createdString = created.toEpochMilli.toString // todo: keep Datetime
```
adpi2, line 130:
```suggestion
    val createdString = created.toEpochMilli.toString
```

server/src/main/scala/ch.epfl.scala.index.server/routes/FrontPage.scala
```diff
@@ -13,28 +13,29 @@ import ch.epfl.scala.index.search.ESRepo
...
 
 class FrontPage(
     dataRepository: ESRepo,
-    db: DatabaseApi,
+    db: WebDatabase,
     session: GithubUserSession
 )(implicit ec: ExecutionContext) {
   import session.implicits._
 
+  val limitFrontPage = 12
```
adpi2, line 28: limit of what?

server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/PublishProcess.scala
```diff
@@ -82,7 +80,7 @@ private[api] class PublishProcess(paths: DataPaths, dataRepository: ESRepo, db:
 
                   Meta.append(
                     paths,
-                    Meta(data.hash, data.path, data.created),
+                    Meta(data.hash, data.path, data.datetimeCreated),
```
adpi2, line 83: Why not change that to `Instant` and remove the dependency to `joda` here?

server/src/main/scala/scaladex/server/service/Scheduler.scala
```diff
@@ -51,8 +55,28 @@ class Scheduler(val name: String, runnable: Runnable)(implicit ec: ExecutionCont
...
+        catch {
+          case NonFatal(e) =>
+            Future.successful(
+              logger.info(
+                s"Scheduler ${name}: Failed because ${e.getMessage}"
+              )
+            )
+            Future.successful(logger.info(e.getStackTrace.mkString("\n")))
+        }
+      Await.result(future, Duration.Inf)
```
adpi2, line 79: The `try {} catch` here will never catch anything because it happens inside a future.
Also the logs inside the catch should probably be warnings and it is not really useful to get the `Instant.now` in the logs since all logs are already dated.


```suggestion
      logger.info(s"Scheduler ${name}: Starting")
       try {
         Await.result(job(), Duration.Inf)
         logger.info("Scheduler $name: Finished")
       } catch {
         case NonFatal(e) =>
            logger.warning(s"Scheduler ${name}: Failed because ${e.getMessage}")
            logger.warning(e.getStackTrace.mkString("\n")))
       }
```

server/src/main/scala/scaladex/server/service/SchedulerService.scala
```diff
@@ -1,57 +1,55 @@
...
 import ch.epfl.scala.utils.ScalaExtensions._
 import com.typesafe.scalalogging.LazyLogging
+import scaladex.server.service.SchedulerService._
+import scaladex.template.SchedulerStatus
+
+class SchedulerService(db: SchedulerDatabase) extends LazyLogging {
+  implicit val ec: ExecutionContextExecutor = ExecutionContext.global
+
+  private val mostDependentProjectScheduler = new Scheduler("project-trends", mostDependentProjectJob, 1.hour)
+  private val mostRecentProject = new Scheduler("update-projects", updateProjectJob, 30.minutes)
```
adpi2, line 19: It would be better to have the same name for the variable and the scheduler.

```suggestion
  private val mostDependentProjectScheduler = new Scheduler("most-dependent", mostDependentProjectJob, 1.hour)
  private val updateProject = new Scheduler("update-projects", updateProjectJob, 30.minutes)
```

server/src/main/scala/scaladex/server/service/SchedulerService.scala
```diff
@@ -1,57 +1,55 @@
...
+    for {
+      refs <- db.getAllProjectRef()
+      _ <- updateCreatedTimeIn(db, refs)
+    } yield ()
 }
 
 object SchedulerService {
-  def updateProjectDependenciesTable(db: DatabaseApi)(implicit ec: ExecutionContext): Future[Unit] =
+
+  def updateProjectDependenciesTable(db: SchedulerDatabase)(implicit ec: ExecutionContext): Future[Unit] =
```
adpi2, line 52: Could be moved to the `SchedulerService` class since it needs the `db` and `ec`.

server/src/main/scala/scaladex/server/service/SchedulerService.scala
```diff
@@ -67,5 +65,18 @@ object SchedulerService {
             s"not able to insertProjectDependencies because of ${e.getMessage}"
           )
         )
+
     } yield ()
+
+  private def updateCreatedTimeIn(db: SchedulerDatabase, refs: Seq[NewProject.Reference])(
```
adpi2, line 71: Same here, it could be move to the class.

core/src/main/scala/ch/epfl/scala/index/newModel/NewProject.scala
mlachkar: ok, I thought it was only used in tests

server/src/main/scala/ch.epfl.scala.index.server/routes/FrontPage.scala
mlachkar: ok

server/src/main/scala/ch.epfl.scala.index.server/routes/api/impl/PublishProcess.scala
mlachkar: we cannot, this is what we discussed yesterday. The file is written with datetime.

adpi2: ok

server/src/main/scala/scaladex/server/service/Scheduler.scala
```diff
@@ -51,8 +55,23 @@ class Scheduler(val name: String, runnable: Runnable)(implicit ec: ExecutionCont
...
+      logger.info(s"Scheduler ${name}: Starting")
+      try {
+        Await.result(job(), Duration.Inf)
+        logger.info("Scheduler $name: Finished")
+      } catch {
+        case NonFatal(e) =>
+          logger.warn(s"Scheduler ${name}: Failed because ${e.getMessage}")
+          logger.warn(e.getStackTrace.mkString("\n"))
+      }
+      Await.result(future, Duration.Inf)
```
adpi2, line 74:
```suggestion
```

adpi2 (approved):
It does not compile but otherwise it looks good to me!

PR #743, opened 2021-11-19
scheduler for github synchronization

mlachkar:
This pull request adds a scheduler that will run asynchronously to update github infos.
- With a 5000 limit per hour on github api, we are able to synchronize all github info in 8 hours which is quiet slow. There are ways to faster this process: for example only update the readme and repo information and 3 other github request do that less often (once per day for example)
- We used to have 8500 github information, but now we only have +6700. In fact 1200 project are not found anymore. I think it could be nice since all our scaladex relies on github, to have a new field called : GithubStatus (Found, NotFound, Forbiden, MovedPermantly), so we can decide later what to do with project that do not exist anymore.
- Right now, when we cancel githubinfo scheduler, the job is not canceled. this shoul be fixed.


data/src/test/scala/ch.epfl.scala.index/github/LinkParserTests.scala
```diff
@@ -1,30 +1,30 @@
...
+//package data
+//package github
+//
+//import org.scalatest.funspec.AsyncFunSpec
+//
+//class LinkParserTests extends AsyncFunSpec {
+//  describe("Link parser") {
+//    it("quote") {
+//      assert(
+//        extractLastPage(
```
adpi2, line 11: Why is this commented?

infra/src/main/resources/reference.conf
```diff
@@ -11,4 +11,8 @@ database {
   name = ${?POSTGRESQL_DATABASE}
   port = 5432
   url = "jdbc:postgresql://"${database.user}":"${database.password}"@localhost:"${database.port}"/"${database.name}
+}
+github {
+  # follow this tutoriel to create your token https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
```
adpi2, line 16:
```suggestion
  # follow this tutorial to create your token https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
```

infra/src/main/scala/ch/epfl/scala/services/github/GithubImplementation.scala
```diff
@@ -0,0 +1,236 @@
...
+import akka.util.ByteString
+import ch.epfl.scala.index.model.misc.GithubInfo
+import ch.epfl.scala.index.model.misc.GithubRepo
+import ch.epfl.scala.index.model.misc.Url
+import ch.epfl.scala.services.GithubService
+import ch.epfl.scala.services.github.GithubModel.Contributor
+import ch.epfl.scala.utils.ScalaExtensions.TraversableOnceFutureExtension
+import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
+
+class GithubImplementation(githubConfig: GithubConfig)(implicit system: ActorSystem) extends GithubService {
```
adpi2, line 39: Maybe `AkkaGithubClient` or just `GithubClient`?

infra/src/main/scala/ch/epfl/scala/services/github/GithubModel.scala
```diff
@@ -0,0 +1,131 @@
...
+      forks_count: Int,
+      mirror_url: Option[String], // "mirror_url": "git://git.apache.org/spark.git",
+      forks: Int,
+      open_issues: Int,
+      default_branch: String, // master
+      subscribers_count: Int, // Watch
+      topics: Seq[String]
+  )
+
+  implicit val repositoryDecoder: Decoder[Repository] = new Decoder[Repository] {
```
adpi2, line 32: Maybe add a comment to explain why the `Decoder` cannot be derived automatically

infra/src/main/scala/ch/epfl/scala/services/github/GithubImplementation.scala
adpi2: I don't really like the name `Implementation`...

mlachkar: Agree

infra/src/main/scala/ch/epfl/scala/services/github/GithubModel.scala
mlachkar: ok

data/src/test/scala/ch.epfl.scala.index/github/LinkParserTests.scala
mlachkar: I dont know

mlachkar: it's the old test of githubReader, that make sure we do the pagination correctly. It's not needed anymore

PR #747, opened 2021-12-09
My projects

mlachkar:


infra/src/main/scala/ch/epfl/scala/services/github/GithubModel.scala
```diff
@@ -128,4 +132,63 @@ object GithubModel {
         labelNames <- labelsJson.traverse(_.hcursor.downField("name").as[String])
       } yield if (isPullrequest.isEmpty) Some(OpenIssue(number, title, url, labelNames)) else None
   }
+
+  case class GithubRepoWithPermission(nameWithOwner: String, viewerPermission: String)
+  case class GithubRepoWithPermissionPage(
```
adpi2, line 137:
```suggestion
  case class RepoAndPermission(nameWithOwner: String, viewerPermission: String)
  case class RepoAndPermissionPage(
```

infra/src/main/scala/ch/epfl/scala/services/github/GithubClient.scala
```diff
@@ -184,24 +220,117 @@ class GithubClient(githubConfig: GithubConfig)(implicit system: ActorSystem) ext
     }
   }
 
+  def fetchMyRepo(myToken: Secret): Future[Map[GithubRepo, String]] = {
```
adpi2, line 223:
```suggestion
  def fetchUserRepos(userToken: Secret): Future[Map[GithubRepo, String]] = {
```

adpi2: Also I would pass the requested permission as a list and filter them at the client side. Otherwise it is not clear that `String` in `Map[GithubRepo, String]` is a permission.
```suggestion
  def fetchMyRepo(myToken: Secret, permissions: Seq[String]): Future[Seq[GithubRepo]] = {
```

server/src/main/scala/ch.epfl.scala.index.server/Github.scala
```diff
@@ -11,26 +11,14 @@ import akka.http.scaladsl.model._
...
-  case class Organization(login: String)
-
-  case class User(login: String, name: Option[String], avatarUrl: String) {
-    def convert(token: String): UserInfo =
-      UserInfo(login, name, avatarUrl, token)
-  }
 }
 
-class Github()(implicit sys: ActorSystem) extends Json4sSupport {
+class Github(githubClient: GithubService)(implicit sys: ActorSystem) extends Json4sSupport {
```
adpi2, line 21: Maybe we can rename it `GithubAuth` and move it to infra

adpi2 (approved):
Looks good enough for now.

infra/src/main/scala/ch/epfl/scala/services/github/GithubClient.scala
mlachkar: yep!

mlachkar:
I cannot understand what's happening with these tests :s

PR #757, opened 2021-12-17
upgrade elasticsearch used locally to be the same on preprod/prod env…

mlachkar:
…ironments

- silence the warn log from elastic search


mlachkar:
the logback.xml that I change in this PR wont change the one used when deployed. it seems that there is another one in scaladex-credentials

PR #760, opened 2021-12-17
move releases to be linked to the new project

mlachkar:


server/src/main/scala/scaladex/server/service/MoveReleasesSynchronizer.scala
```diff
@@ -0,0 +1,29 @@
...
+import ch.epfl.scala.index.model.misc.GithubStatus
+import ch.epfl.scala.index.newModel.NewProject
+import ch.epfl.scala.services.SchedulerDatabase
+import ch.epfl.scala.utils.ScalaExtensions.TraversableOnceFutureExtension
+
+class MoveReleasesSynchronizer(db: SchedulerDatabase)(implicit ec: ExecutionContext)
+    extends Scheduler("move-releases-synchronizer", 5.minutes) {
+  override def run(): Future[Unit] =
+    for {
+      allProjects <- db.getAllProjects()
```
adpi2, line 16: This methods gets all the project with their gthubInfo which can be huge because it contains all readme and issues...

It would be better to have a method `getAllProjectsStatus` that return `Seq[(Project.Reference, GithubStatus)]`

Or event better, after we split the columns of github_status:
```
getAllMovedProjects: Seq[Project.Reference]
```

mlachkar: yes. will implement it when I change de coder for githubStatus

infra/src/main/scala/ch/epfl/scala/services/storage/sql/tables/ReleaseTable.scala
```diff
@@ -43,6 +44,11 @@ object ReleaseTable {
   def indexedReleased(): Query0[Long] =
     buildSelect(tableFr, fr0"count(*)").query[Long]
 
+  def updateProjectRef(): Update[(NewProject.Reference, MavenReference)] =
```
adpi2, line 47:
```suggestion
  val updateProjectRef: Update[(NewProject.Reference, MavenReference)] =
```

PR #761, opened 2021-12-20
make all organization and repository lowercase

mlachkar:


infra/src/main/resources/reference.conf
```diff
@@ -14,5 +14,5 @@ database {
 }
 github {
   # follow this tutorial to create your token https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
-  token = ${?GITHUB_TOKEN}
+  token = ${?TOKEN_FOR_GITHUB}
```
adpi2, line 17: We could have both so that we don't need to change anything in our local and server configuration:

```
  token = ${?GITHUB_TOKEN}
  token = ${?TOKEN_FOR_GITHUB}
```

mlachkar: true! I just found this can be confusing

adpi2: ok

PR #763, opened 2021-12-21
Add plausible

adpi2:
It only activates plausible on production. But after this will be merged to develop we will be able to use the `Env` object to activate plausible to pre-prod too.

### DIFF
--- a/server/src/main/scala/ch.epfl.scala.index.server/Server.scala
+++ b/server/src/main/scala/ch.epfl.scala.index.server/Server.scala
@@ -49,12 +49,13 @@ object Server {
     // because of the sbtResolver mode
     val data = DataRepository.open()
 
-    val searchPages = new SearchPages(data, session)
+    val searchPages = new SearchPages(config.production, data, session)
     val userFacingRoutes = concat(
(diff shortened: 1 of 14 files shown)


server/src/main/scala/ch.epfl.scala.index.server/Server.scala
```diff
@@ -49,12 +49,13 @@ object Server {
     // because of the sbtResolver mode
     val data = DataRepository.open()
 
-    val searchPages = new SearchPages(data, session)
+    val searchPages = new SearchPages(config.production, data, session)
```
mlachkar, line 52: I was about to comment.. what is this config ? it should be config.env ^^

mlachkar (approved):
I don't get why we would need analytics on preprod! Preprod is only for development purposes, so we don't really expect users on it ?

server/src/main/scala/ch.epfl.scala.index.server/Server.scala
adpi2: ^^

adpi2:
I think we need it on preprod just to test that it works.

PR #768, opened 2021-12-22
remove contributorcount and create a new githubInfo for ES

mlachkar:


core/src/main/scala/ch.epfl.scala.index.model/misc/GithubInfo.scala
```diff
@@ -35,29 +39,45 @@ case class GithubInfo(
...
     chatroom: Option[Url] = None,
-    beginnerIssues: List[GithubIssue] = List()
-)
+    beginnerIssues: List[GithubIssue] = List() // right now it's all issues, not only beginners issues
+) {
+  val organization: Organization = projectRef.organization
+  val repository: Repository = projectRef.repository
+  val contributorCount: Int = contributors.size
+
+  def toGithubInfoDocument: GithubInfoDocument =
```
adpi2, line 53:
```suggestion
  def toDocument: GithubInfoDocument =
```

adpi2 (approved):
Looks Good

PR #774, opened 2021-12-23
Rename database

mlachkar:


infra/src/main/scala/scaladex/infra/github/GithubModel.scala
```diff
@@ -1,8 +1,13 @@
 package scaladex.infra.github
 
+import java.time.Instant
+
+import scala.util.Try
+
 import cats.implicits.toTraverseOps
 import io.circe._
 import io.circe.generic.semiauto._
+import org.joda.time.format.ISODateTimeFormat
```
adpi2, line 10: We should stop using `joda.time`. Here you should be able to use `DateTimeFormatter.ISO_DATE_TIME` form `java.time`.

```suggestion
import java.time.format.DateTimeFormatter
```

infra/src/main/scala/scaladex/infra/storage/sql/tables/ProjectTable.scala
```diff
@@ -64,12 +61,12 @@ object ProjectTable {
     buildSelect(
       fullTableFr,
       allFieldsFr,
-      fr0"WHERE creation_date IS NOT NULL ORDER BY creation_date DESC LIMIT $limit"
+      fr0"WHERE p.creation_date IS NOT NULL ORDER BY p.creation_date DESC LIMIT ${limit.toLong}"
```
adpi2, line 64: Is this `limit.toLong` needed because of doobie check?

```suggestion
      fr0"WHERE p.creation_date IS NOT NULL ORDER BY p.creation_date DESC LIMIT ${limit.toLong}"
```

adpi2 (approved):

infra/src/main/scala/scaladex/infra/storage/sql/tables/ProjectTable.scala
mlachkar: yes

Issue #749, opened 2021-12-09
Add an endpoint handling Scaladoc linking


pikinier20:
The idea is to add some endpoint that for given library and member coordinates:
```scala
case class Library(
  org: String,
  lib: String,
  version: String
)
case class DRI(
  location: String,
  anchor: String
)
```
redirects to the documentation of the member.
It should be aware of which version of Scaladoc it links to.

julienrf:
@pikinier20 In the web interface, we show a “Scaladoc” link that points to the scaladoc URL of the given artifact (it defaults to javadoc.io, but users can override it).

![Screenshot from 2022-01-10 16-15-07](https://user-images.githubusercontent.com/332812/148789997-8279254a-50be-48a3-a5ee-4b79522e1a09.png)

Do I understand well that this is this link that you would like to retrieve via the endpoint? Could you please give more details about the response model you suggest in the issue description? What are “location” and “anchor”?

pikinier20:
To be precise, I would like to retrieve this link + link to member inside this documentation. So, I would like to pass information to Scaladex like:
Library(
  "org.virtuslab",
  "scala-cli",
  "0.0.9"
 )
 DRI(
   "runner/SomeRunner"
   "run-..."
 )
 and have a link to documentation of scala-cli pointing on the method run in class SomeRunner.
 The DRI model is the same as we use in Scaladoc and actually basing only on it and a kind of documentation (whether it's Scaladoc 2, 3 or some Javadoc) we can construct a valid link to member. 
 
 We had a discussion about the way Scaladex retrieves documentation location. Sbt has an option to attach the location of documentation to POM (then uses this information when `autoAPIMapping` is enabled) and we could extend this option to add there a version of Scaladoc. Scaladex might want to use this information.

adpi2:
I deployed a new scaladoc endpoint in the dev environment: https://index-dev.scala-lang.org

The template of this endpoint is:
```
https://index-dev.scala-lang.org/artifacts/<group-id>/<artifact-id>/<version>/scaladoc<optional-path>
```

where `<optional-path>` can contain a path (e.g., `/api/io/iteratee/Enumeratee.html`), an anchor (e.g., `#io.iteratee.Enumeratee`) or both.

Examples:
- Scala 3:
https://index-dev.scala-lang.org/artifacts/io.iteratee/iteratee-core_3/0.20.0/scaladoc/api/io/iteratee/Enumeratee.html
- Scala 2.13:
[...]

PR #813, opened 2022-01-10
Use endpoints4s for RPC between web client and server

julienrf:
I wanted to do this as a warmup before #749.

The idea is to avoid duplicating the communication protocol details between the [client](https://github.com/scalacenter/scaladex/blob/d310d1768f0f63b0b0907ff4741854ddd3ae16be/core/shared/src/main/scala/scaladex/core/api/SearchRequest.scala#L16-L24) and the [server](https://github.com/scalacenter/scaladex/blob/d310d1768f0f63b0b0907ff4741854ddd3ae16be/server/src/main/scala/scaladex/server/route/package.scala#L40-L50).

The communication protocol is defined once (in `AutocompletionEndponits`) and reused by both the web client and the server.

Discussion point: currently, the server uses the same logic to process autocompletion requests and search requests, but this PR only changes the logic for the autocompletion endpoint. I wonder if I should re-unify them. (I see little benefits in doing so, except maybe we can factor out similar code)


core/shared/src/main/scala/scaladex/core/api/AutocompletionEndpoints.scala
```diff
@@ -0,0 +1,64 @@
...
+trait AutocompletionEndpoints extends algebra.Endpoints with algebra.JsonEntitiesFromSchemas {
+
+  /** Enrich a request with session information */
+  def withOptionalSession[A](request: Request[A]): Request[WithSession[A]]
+
+  /** A type `A` enriched with session information (unknown yet, defined by the web client and server) */
+  type WithSession[A]
+
+  // JSON schema of the autocompletion response entity
+  implicit val autocompletionResponseSchema: JsonSchema[AutocompletionResponse] =
```
adpi2, line 34: Out of curiosity, is it possible to derive a `JsonSchema` from a circe `Codec`?

julienrf: No, because Circe codecs are not “introspectable” (they are just functions `Json => DecodingResult[A]`). But it is possible to derive a Circe codec from a `JsonSchema`, if needed.

adpi2: I guess not because the documentation of the schema would be missing. But I see that we can do the other way around: derive a `Codec` from a `JsonSchema` (https://endpoints4s.github.io//interpreters/circe.html).

server/src/main/scala/scaladex/server/route/api/SearchApi.scala
```diff
@@ -252,4 +248,19 @@ class SearchApi(searchEngine: SearchEngine, database: WebDatabase, session: Gith
...
+
+  object AutocompletionApi extends AutocompletionEndpoints with server.Endpoints with server.JsonEntitiesFromSchemas {
+    // On the server-side, the session is made of a user id
+    case class WithSession[A](data: A, maybeUserId: Option[UUID])
+    def withOptionalSession[A](request: Request[A]): Request[WithSession[A]] =
+      new Request[WithSession[A]] {
+        val directive: Directive1[WithSession[A]] = Directive[Tuple1[WithSession[A]]] { k =>
+          optionalSession(refreshable, usingCookies) { maybeUserId =>
+            request.directive(a => k(Tuple1(WithSession(a, maybeUserId))))
+          }
```
julienrf, line 260: I manually tested (locally) that autocompleting “my projects” works as expected.

adpi2 (approved):
Thanks  @julienrf this looks good. Would you be able to set up a documentation endpoint for this autocomplete request?

julienrf:
Yes, sure. I thought that endpoint would not be part of the public API since it is used by the web frontend only, but maybe it is useful to expose it publicly as well. Maybe @mlachkar has an opinion on that matter?

adpi2:
@julienrf

You are probably right but it this documentation endpoint could be used as an example to help us.

julienrf:
I have just pushed a commit that adds an endpoint serving the OpenAPI documentation at `/api/open-api.json`.

Currently, it only documents `/api/autocomplete`. Here is the document:

<details>
{
  "openapi":"3.0.0",
  "info":{
    "title":"Scaladex API",
    "version":"0.1.0"
  },
  "paths":{
    "/api/autocomplete":{
      "get":{
        "responses":{
          "400":{
            "description":"Client error",
            "content":{
              "application/json":{
                "schema":{
                  "$ref":"#/components/schemas/endpoints.Errors"
                }
              }
            }
          },
          "500":{
            "description":"Server error",
            "content":{
              "application/json":{
                "schema":{
                  "$ref":"#/components/schemas/endpoints.Errors"
                }
              }
            }
          },
          "200":{
            "description":"",
            "content":{
              "application/json":{
                "schema":{
                  "type":"array",
                  "items":{
                    "type":"object",
                    "properties":{
                      "organization":{
                        "type":"string"
                      },
                      "repository":{
                        "type":"string"
                      },
                      "description":{
                        "type":"string"
                      }
                    },
                    "required":[
                      "organization",
                      "repository",
                      "description"
                    ]
                  }
                }
              }
            }
          }
        },
        "parameters":[
          {
            "name":"q",
            "in":"query",
            "schema":{
              "type":"string"
            },
            "description":"Main query (e.g., 'json', 'testing', etc.)",
            "required":true
          },
          {
            "name":"you",
            "in":"query",
            "schema":{
              "type":"string"
            },
            "description":"Used internally by Scaladex web user interface"
          },
          {
            "name":"topics",
            "in":"query",
            "schema":{
              "type":"array",
              "items":{
                "type":"string"
              }
            },
            "description":"Filter the results matching the given topics only"
          },
          {
            "name":"targetTypes",
            "in":"query",
            "schema":{
              "type":"array",
              "items":{
                "type":"string"
              }
            },
            "description":"Filter the results matching the given target types only (e.g., 'jvm', 'js', 'native', 'sbt', 'java')"
          },
          {
            "name":"scalaVersions",
            "in":"query",
            "schema":{
              "type":"array",
              "items":{
                "type":"string"
              }
            },
            "description":"Filter the results matching the given Scala versions only (e.g., 'scala3', '2.13', '2.12', '2.11')"
          },
          {
            "name":"scalaJsVersions",
            "in":"query",
            "schema":{
              "type":"array",
              "items":{
                "type":"string"
              }
            },
            "description":"Filter the results matching the given Scala.js versions only (e.g., '1.x', '0.6')"
          },
          {
            "name":"scalaNativeVersions",
            "in":"query",
            "schema":{
              "type":"array",
              "items":{
                "type":"string"
              }
            },
            "description":"Filter the results matching the given Scala Native versions only (e.g., '0.4', '0.3')"
          },
          {
            "name":"sbtVersions",
            "in":"query",
            "schema":{
              "type":"array",
              "items":{
                "type":"string"
              }
            },
            "description":"Filter the results matching the given sbt versions only (e.g., '1.0', '0.13')"
          },
          {
            "name":"contributingSearch",
            "in":"query",
            "schema":{
              "type":"boolean"
            }
          }
        ]
      }
    }
  },
  "components":{
    "schemas":{
      "endpoints.Errors":{
        "type":"array",
        "items":{
          "type":"string"
        }
      }
    },
    "securitySchemes":{
      
    }
  }
}
</details>

And here is an example of how it looks like on Swagger UI:

Screenshot from 2022-01-10 13-49-02

Another thing we could publish, in addition to the `open-api.json` document, would be a Scala artifact defining the API (the `AutocompletionEndpoints` trait, for now). These artifacts could be used by Scala users (e.g., Metals, scala-cli) to call the Scaladex API with a high-level Scala API.

adpi2 (approved):
That's amazing thanks

PR #815, opened 2022-01-12
Fix local packaging of webclient-fastopt.js

adpi2:


project/ScalaJSHelper.scala
```diff
@@ -7,35 +7,25 @@ object ScalaJSHelper {
...
         )
       ).toSeq
     }.taskValue,
-    Compile / packageBin / mappings := {
-      val mappingExcludingNonOptimized =
-        (Compile / packageBin / mappings).value.filterNot {
-          case (f, r) =>
-            f.getName.endsWith("-fastopt.js") ||
-              f.getName.endsWith("js.map")
-        }
```
julienrf, line 24: Does that mean that the package will contain both the `-fastopt.js` and the `-opt.js`?

adpi2: yes

PR #817, opened 2022-01-12
Rework publication

adpi2:


infra/src/main/resources/reference.conf
```diff
@@ -18,4 +18,10 @@ scaladex {
     # follow this tutorial to create your token https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
     token = ${?GITHUB_TOKEN}
   }
+  filesystem {
```
mlachkar, line 21: so the goal is to remove dataPath and replace it by fylesystemConfig ?

infra/src/main/scala/scaladex/infra/storage/DataPaths.scala
```diff
@@ -54,47 +57,27 @@ scaladex-contrib
...
-      contrib: String,
-      index: String,
-      credentials: String,
+      config: FilesystemConfig,
       env: Env
   ): DataPaths = {
-    val contribPath = Paths.get(contrib)
-    val indexPath = Paths.get(index)
-    val credentialsPath = Paths.get(credentials)
+    import config._
```
mlachkar, line 69: I m not completly a fan of this thing

infra/src/main/resources/reference.conf
adpi2: Yes I would like to remove `DataPaths` completely and move all methods that read/write into the filesystem to the `LocalStorageRepo` class. So that we have a better view of all that depends on reading/writing to disk. Also to create a mock of `LocalStorage` and test the things that depends on it.

That's quite some work though.

So in the end `LocalStorageRepo` should be the only thing consuming the `FilesystemConfig`.

mlachkar: which make sense!

core/shared/src/main/scala/scaladex/core/service/PomResolver.scala
```diff
@@ -0,0 +1,9 @@
+package scaladex.core.service
+
+import java.nio.file.Path
+
+import scala.concurrent.Future
+
+trait PomResolver {
+  def resolve(groupId: String, artifactId: String, version: String): Future[Option[Path]]
```
mlachkar, line 8: nice

infra/src/main/resources/migrations/V2__fix_platform.sql
```diff
@@ -0,0 +1,8 @@
+UPDATE artifacts SET platform = '_3' WHERE platform = '_3.x';
+UPDATE artifacts SET platform = '_sjs1_2.11' WHERE platform ='_sjs1.x_2.11';
+UPDATE artifacts SET platform = '_sjs1_2.12' WHERE platform = '_sjs1.x_2.12';
+UPDATE artifacts SET platform = '_sjs1_2.13' WHERE platform = '_sjs1.x_2.13';
+UPDATE artifacts SET platform = '_sjs1_3.0.0-RC1' WHERE platform = '_sjs1.x_3.0.0-RC1';
+UPDATE artifacts SET platform = '_sjs1_3.0.0-RC2' WHERE platform = '_sjs1.x_3.0.0-RC2';
+UPDATE artifacts SET platform = '_sjs1_3.0.0-RC3' WHERE platform = '_sjs1.x_3.0.0-RC3';
+UPDATE artifacts SET platform = '_sjs1_3' WHERE platform = '_sjs1.x_3.x';
```
mlachkar, line 8: nice!

infra/src/main/scala/scaladex/infra/config/FilesystemConfig.scala
```diff
@@ -0,0 +1,18 @@
...
+
+object FilesystemConfig {
+  def from(config: Config): FilesystemConfig = {
+    val temp = Paths.get(config.getString("scaladex.filesystem.temp"))
+    val index = Paths.get(config.getString("scaladex.filesystem.index"))
+    val contrib = Paths.get(config.getString("scaladex.filesystem.contrib"))
+    val credentials = Paths.get(config.getString("scaladex.filesystem.credentials"))
+    FilesystemConfig(temp, index, contrib, credentials)
+  }
+}
```
mlachkar, line 18: I would love to change that to use pureconfig here... It will remove many of this code

infra/src/main/scala/scaladex/infra/storage/local/LocalStorageRepo.scala
```diff
@@ -28,26 +31,60 @@ class LocalStorageRepo(dataPaths: DataPaths) extends LocalStorageApi {
...
+  override def saveAllProjectSettings(settings: Map[Project.Reference, Project.Settings]): Unit =
     Files.write(
-      dataPaths.liveProjects,
+      projectSettings,
       Printer.noSpaces.print(settings.asJson).getBytes(StandardCharsets.UTF_8)
     )
+
+  override def createTempFile(content: String, prefix: String, suffix: String): Path = {
+    val tempFile = Files.createTempFile(temp, prefix, suffix)
+    Files.write(tempFile, content.getBytes(StandardCharsets.UTF_8))
```
mlachkar, line 48: hope that `Files` close it's own resource. Otherwise I have seen people writing this method like this 
```
  final def writeFile(file: AbsolutePath, content: String): Try[Unit] =
    Try {
      val bw = new BufferedWriter(new FileWriter(file.toFile))
      bw.write(content)
      bw.close()
    }
```

mlachkar (approved):

infra/src/main/scala/scaladex/infra/storage/local/LocalStorageRepo.scala
adpi2: Yes `Files.write` manage the resource itself.

infra/src/main/scala/scaladex/infra/config/FilesystemConfig.scala
adpi2: The place where we load the config is now well contained in a few files. So it should not be hard to replace `typesafe.config` with `pureconfig`.

PR #818, opened 2022-01-13
Create .scala-steward.conf

mlachkar:


.scala-steward.conf
```diff
@@ -0,0 +1 @@
+pullRequests.frequency = "0 0 1,15 * ?"
```
adpi2, line 1: What does it mean?

mlachkar: every first and 15 of the month

adpi2 (approved):

PR #819, opened 2022-01-13
fix wrong values of platform in postgresql

mlachkar:


adpi2:
Thanks a lot @mlachkar, I used it in https://github.com/scalacenter/scaladex/pull/817

PR #878, opened 2022-01-26
Introduce new index format

adpi2:
Introduce a new index format in the filesystem that looks like this:
```
<organization> / <repository> /
  project.json
  artifacts.json
  dependencies.json
```

As a consequence we don't store all the poms, and we don't have huge files that contain all the metadata or the project settings of all projects.

It makes the generation of the index more simple: for each project in the database, get the artifacts and the dependencies and save them. And the generation of the database from the index more simple too: for each project in the index, read the artifacts and dependencies and insert them in the database.

What if we need to re-read all poms at some point? We hope it won't happen very often, but if we really need to we can download and cache them with Coursier using the `CoursierResolver` class.

What if the schemas of `Project`, `Artifact` and `ArtifactDependency` change? We deploy scaladex and generate a new index. It takes about 30 min to generate the full index.


modules/data/src/main/scala/scaladex/data/central/CentralMissing.scala
```diff
@@ -162,39 +156,14 @@ class CentralMissing(paths: DataPaths)(implicit val system: ActorSystem) {
...
-        .loadAll(paths.poms(LocalPomRepository.MavenCentral))
-        .flatMap {
-          case (pom, _) =>
-            metaExtractor
-              .extract(pom)
-              .filter(meta => meta.platform != Platform.Java)
-              .map(_ => pom.groupId)
-        }
-        .toSet
+    val allGroups: Set[String] = ???
```
mlachkar, line 166: is that ok ?

adpi2: It's done on purpose.

This code won't be useful anymore but I would like to keep for a while to translate it into a scheduled job. That would be the scheduled job that find the missing artifacts from Maven Central.

mlachkar: ok!

mlachkar (approved):

PR #893, opened 2022-02-07
Rework filters, version matrix and badges

adpi2:
## Introduction

1. Remove Scala 3 pre-release version
2. Rework Platform:

`_sjs1_2.13` is the binary version of an artifact compiled with Scala 2.13 for the Scala.js 1.x platform.
As a consequence we now have:
A binary version can also be `_2.13`, which means the artifact is compiled with Scala 2.13 for the JVM platform.
Or it can be empty, which means it is compiled with Java for the JVM platform.

To model this we now have:
```scala
case class BinaryVersion(platform: Platform, language: Language)

trait Language
case object Java extends Language
case object Scala(version: SemanticVersion) extends Language

trait Platform
case object JVM extends Platform
case class ScalaJs(version: SemanticVersion) extends Platform
case class ScalaNative(version: SemanticVersion) extends Platform
case class SbtPlugin(version: SemanticVersion) extends Platform
```

This new model makes it easier to design the version matrix, the filters and the badges.

### Before

image

### After 

image

### Before

image

### After

image

## Badges

Applied changes described in https://github.com/scalacenter/scaladex/pull/722#issuecomment-1001481693.

### Before

image
image

### After

image
image
image


mlachkar (approved):
Nice!

PR #933, opened 2022-03-04
Adding percentage of Scala for a project

### LINKED ISSUE #928: Get percentage of Scala code in a project from Github API
## Motivation

In Scaladex we index all projects from which a Scala artifact was published. But currently we don't know if Scala is the primary language of the project or not.

Comparing the number of stars of Scala projects with projects from other ecosystems is not really fair because the latters can have bigger communities. Indeed in the top 20 projects in https://index.scala-lang.org/search?q=* only 8 of them uses Scala as their primary language.

To mitigate this we could get the percentage of Scala in each project using the Github API  and use it in the Elasticsearch scoring method. 

## Identified Obstacle

How to use the `scalaPercentage` in the scoring method?

## Implementation Guideline

- [x] Add a `scalaPercentage` field in `GithubInfo`
- [x] Get the information from Github in the `GithubClient`
- [ ] Add the `scalaPercentage` field in `GithubInfoDocument` that is indexed in elasticsearch
- [ ] Use this field in the  `githubStarScoring` method of `ElasticsearchEngine`

## Expectations

- A majority of the top 20 projects in https://index.scala-lang.org/search?q=* should use Scala as their primary language.
- the `GithubClient` should be tested

jyoo980:
This is a basic implementation of getting some information regarding how much of a codebase is Scala, i.e. `.scala` or`.sc`files.

I used this [endpoint](https://docs.github.com/en/rest/reference/repos#list-repository-languages), which produces a response like so:

```json
{
  "Scala": 40292600,
  "Python": 7343749,
  "Java": 4526889,
  "Jupyter Notebook": 4310516,
  "HiveQL": 1872438,
  "R": 1272682,
  "PLpgSQL": 352963,
  "Shell": 230745,
  "JavaScript": 222664,
  "q": 98156,
  "ANTLR": 56929,
  "HTML": 42080,
  "Roff": 31648,
  "Batchfile": 27405,
  "CSS": 26221,
  "Dockerfile": 9711,
  "PowerShell": 4221,
  "Makefile": 2374,
  "Thrift": 2016,
  "C": 1493,
  "ReScript": 240
}
```

Given a payload such as
```sh
curl -H "Accept: application/vnd.github.v3+json" \
  https://api.github.com/repos/apache/spark/languages
```

Which is basically a dictionary of languages to the number of bytes they occupy in the project. I understand that this isn't an ideal implementation (perhaps a count of files of a given language might be a better measure), so I'm open to suggestions.

Additionally, more work needs to be done on deciding how this percentage is used in calculating the scoring for Elasticsearch, but I think that can be discussed more in #928.

Should eventually close #928

### DIFF
--- a/modules/infra/src/it/scala/scaladex/infra/GithubClientTests.scala
+++ b/modules/infra/src/it/scala/scaladex/infra/GithubClientTests.scala
@@ -65,8 +65,18 @@ class GithubClientTests extends AsyncFunSpec with Matchers {
   }
 
   it("getUserInfo") {
-    for (userInfo <- client.getUserInfo())
+    for (_ <- client.getUserInfo())
       yield succeed
+  }
+
+  it("getPercentageOfLanguage should return 0, given a repo with none of the target language") {
+    for (percentOfLanguage <- client.getPercentageOfLanguage(Scalafix.reference, "Racket"))
+      yield percentOfLanguage shouldBe 0
+  }
+
+  it("should return a non-zero value, given a repo which contains the target language") {
+    for (percentOfLanguage <- client.getPercentageOfLanguage(Scalafix.reference, "Scala"))
+      yield percentOfLanguage > 0 shouldBe true
   }
 
   if (!isCI) {
--- a/modules/infra/src/main/scala/scaladex/infra/GithubClient.scala
+++ b/modules/infra/src/main/scala/scaladex/infra/GithubClient.scala
@@ -83,6 +83,7 @@ class GithubClient(token: Secret)(implicit val system: ActorSystem)
       contributors <- getContributors(repo.ref)
       openIssues <- getOpenIssues(repo.ref)
       chatroom <- getGitterChatRoom(repo.ref)
+      scalaPercentage <- getPercentageOfLanguage(repo.ref, language = "Scala")
     } yield GithubInfo(
       homepage = repo.homepage.map(Url),
       description = repo.description,
@@ -99,7 +100,8 @@ class GithubClient(token: Secret)(implicit val system: ActorSystem)
       contributingGuide = communityProfile.flatMap(_.contributingFile).map(Url),
       codeOfConduct = communityProfile.flatMap(_.codeOfConductFile).map(Url),
       chatroom = chatroom.map(Url),
-      openIssues = openIssues.map(_.toGithubIssue).toList
+      openIssues = openIssues.map(_.toGithubIssue).toList,
+      scalaPercentage = scalaPercentage
     )
 
   def getReadme(ref: Project.Reference): Future[Option[String]] = {
@@ -185,6 +187,17 @@ class GithubClient(token: Secret)(implicit val system: ActorSystem)
           }
       }
       .fallbackTo(Future.successful(Seq.empty))
+  }
+
+  def getPercentageOfLanguage(ref: Project.Reference, language: String): Future[Option[Int]] = {
+    val request = HttpRequest(uri = s"${repoUrl(ref)}/languages").addHeader(acceptJson).addCredentials(credentials)
+    for {
+      response <- get[Map[String, String]](request)
+      languages = response.transform((_, numBytes) => BigInt(numBytes))
+      maybeNumBytesLang <- Future(languages.get(language))
+      totalNumBytes = languages.values.sum
+      if languages.nonEmpty
+    } yield maybeNumBytesLang.map(numBytesLang => (numBytesLang / totalNumBytes).toInt)
   }
 
   def getGitterChatRoom(ref: Project.Reference): Future[Option[String]] = {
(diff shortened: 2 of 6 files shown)


modules/infra/src/main/scala/scaladex/infra/GithubClient.scala
```diff
@@ -187,6 +189,18 @@ class GithubClient(token: Secret)(implicit val system: ActorSystem)
       .fallbackTo(Future.successful(Seq.empty))
   }
 
+  def getPercentageOfLanguage(ref: Project.Reference, language: String): Future[Int] = {
+    val request =
+      HttpRequest(uri = s"${repoUrl(ref)}/languages").addHeader(acceptJson).addCredentials(credentials)
+    for {
+      response <- get[Map[String, Int]](request)
+      languages = response.transform((_, numBytes) => numBytes.toFloat)
```
jyoo980, line 197: Slightly concerned about using `.toFloat`, overflow might seem possible. Any thoughts on which datatype is preferred?

modules/infra/src/it/scala/scaladex/infra/GithubClientTests.scala
```diff
@@ -65,10 +65,20 @@ class GithubClientTests extends AsyncFunSpec with Matchers {
   }
 
   it("getUserInfo") {
-    for (userInfo <- client.getUserInfo())
```
jyoo980, line 68: Value of `userInfo` wasn't being used in the test at all.

modules/infra/src/main/scala/scaladex/infra/GithubClient.scala
adpi2: `.toFloat` can lose precision on big int values but I don't think it can overflow so it seems fine.

But maybe we don't need to convert all numBytes values to float, only the one of the language:
```scala
for (response <- get[Map[String, Int]](request))
yield response.get(language)
  .fold(0)(bytesLang => ((bytesLang.toFloat / response.sum) * 100).toInt)
```

adpi2 (commented):
Good work @jyoo980

Unfortunately this breaks the `GithubInfoTableTests`:

```
[info] GithubInfoTableTests:
INFO  HikariDataSource - HikariPool-1 - Starting...
INFO  HikariDataSource - HikariPool-1 - Start completed.
[info] - check insert *** FAILED ***
[info]   Update[(scaladex.core.model.Project.Reference, scaladex.core.model.GithubInfo)] defined at (source location unknown)
[info]     INSERT INTO github_info (organization, repository, homepage,
[info]     description, logo, stars, forks, watchers, issues, creation_date,
[info]     readme, contributors, commits, topics, contributing_guide,
[info]     code_of_conduct, chatroom, open_issues) VALUES (?, ?, ?, ?, ?, ?, ?,
[info]     ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
[info]     ✓ SQL Compiles and TypeChecks
[info]     ✓ P01 String                  →  VARCHAR   (varchar)
[info]     ✓ P02 String                  →  VARCHAR   (varchar)
[info]     ✓ P03 Option[String]          →  VARCHAR   (varchar)
[info]     ✓ P04 Option[String]          →  VARCHAR   (varchar)
[info]     ✓ P05 Option[String]          →  VARCHAR   (varchar)
[info]     ✓ P06 Option[Int]             →  INTEGER   (int4)
[info]     ✓ P07 Option[Int]             →  INTEGER   (int4)
[info]     ✓ P08 Option[Int]             →  INTEGER   (int4)
[info]     ✓ P09 Option[Int]             →  INTEGER   (int4)
[info]     ✓ P10 Option[OffsetDateTime]  →  TIMESTAMP (timestamptz)
[info]     ✓ P11 Option[String]          →  VARCHAR   (text)
[info]     ✓ P12 String                  →  VARCHAR   (varchar)
[info]     ✓ P13 Option[Int]             →  INTEGER   (int4)
[info]     ✓ P14 String                  →  VARCHAR   (varchar)
[info]     ✓ P15 Option[String]          →  VARCHAR   (varchar)
[info]     ✓ P16 Option[String]          →  VARCHAR   (varchar)
[info]     ✓ P17 Option[String]          →  VARCHAR   (varchar)
[info]     ✓ P18 String                  →  VARCHAR   (varchar)
[info]     ✕ P19 Option[Int]             →            
[info]       Interpolated value has no corresponding SQL parameter and likely
[info]       appears inside a comment or quoted string. Ior.Left will result in
[info]       a runtime failure; fix this by removing the parameter. (Checker.scala:56)
```

That's because doobie, the sql library, does not know how to insert this new field `scalaPercentage` in the database.

To fix this you can:
1. create a new SQL column in the database
In [modules/infra/src/main/resources/migrations](https://github.com/scalacenter/scaladex/tree/main/modules/infra/src/main/resources/migrations)
you can add a database migration `V6__add_scala_percentage.sql` file that contain the SQL request to add the `scala_percentage` field in the `github_info` table.

2. Declare this new field in `GithubInfoTable.infoFields`  

Hopefully this should fix all the tests. You can check this with:
```
$ sbtn
> infra / test
```

adpi2 (approved):
Thanks! I appreciate the contribution.

PR #934, opened 2022-03-06
Adding `scala_percentage` field to ES mapping + scoring function for Github Stars

### LINKED ISSUE #928: Get percentage of Scala code in a project from Github API

jyoo980:
**Current ordering of projects for `GET /search?q=*`**

1. [shadowsocks/shadowsocks-android](https://github.com/shadowsocks/shadowsocks-android) (0% Scala)
2. [apache/spark](https://github.com/apache/spark) (66.3% Scala)
3. [dmlc/xgboost](https://github.com/dmlc/xgboost) (9.7% Scala)
4. [apache/kafka](https://github.com/apache/spark) (22.8% Scala)
5. [apache/incubator-mixnet](https://github.com/apache/incubator-mixnet) (0% Scala)

**New ordering of projects for `GET /search?q=*` with these code changes**

1. [scala/scala](https://github.com/scala/scala) (97.7% Scala)
2. [akka/akka](https://github.com/akka/akka) (60.1% Scala)
3. [lampepfl/dotty](https://github.com/lampepfl/dotty) (97.5% Scala)
4. [scalanlp/breeze](https://github.com/scalanlp/breeze) (99.8% Scala)
5. [netflix/atlas](https://github.com/netflix/atlas) (98.1% Scala)

Should eventually close #928.


modules/infra/src/main/scala/scaladex/infra/ElasticsearchEngine.scala
```diff
@@ -302,9 +303,13 @@ class ElasticsearchEngine(esClient: ElasticClient, index: String)(implicit ec: E
...
 
   private def gitHubStarScoring(query: Query): Query = {
-    val scorer = fieldFactorScore("githubInfo.stars")
-      .missing(0)
-      .modifier(FieldValueFactorFunctionModifier.LN2P)
+    val githubStarField = fieldAccess("githubInfo.stars", default = "1")
+    val scalaPercentageField = fieldAccess("githubInfo.scalaPercentage", default = "1")
+    val scorer = scriptScore(
+      Script(
+        script = s"Math.log(($githubStarField) * ($scalaPercentageField))"
```
jyoo980, line 310: The argument to `Math.log` can't be 0, since the behaviour of `log(0)` is undefined/diverges toward negative infinity. As a result, the function here behaves as the log of:

  * `1`, if _both_ the star field and Scala percentage field is missing.
  * The value of the star field if the Scala percentage field is missing.
  * The value of the Scala percentage field if the star field is missing.
  * The value of the product of the star field and Scala percentage field if both are defined.

Does this sound reasonable?

adpi2: Yes it does.

However I wonder how this will behave during the transition period: the GithubUpdater will take more than 8 hours to run. During this period of time I would rather not penalize the projects that are not yet updated (Scala percentage is missing). So I suggest to use 100 as the default value of Scala percentage.

(I pushed the change)

adpi2 (approved):
Thanks for the contribution!

I pushed a test in `ElasticsearchEngineTests` to make sure that we won't break this in the future.

adpi2:
I deployed this PR to https://index-dev.scala-lang.org

It will take a bit more than 8 hours to fetch all the new `GithubInfo` with the `scalaPercentage`.

PR #938, opened 2022-03-07
Fix intersection observer

adpi2:
Intersection observer fails to capture some events when the scrolling is too fast. See https://stackoverflow.com/questions/61951380/intersection-observer-fails-sometimes-when-i-scroll-fast

In order to fix this, in the callback of the observer, we check the bounding rect of all sections and add or remove the 'active' class of their corresponding nav items.

Tested in Chrome and Firefox

scrolling-fast

### DIFF
--- a/modules/webclient/src/main/scala/scaladex/client/ActiveNavObserver.scala
+++ b/modules/webclient/src/main/scala/scaladex/client/ActiveNavObserver.scala
@@ -1,8 +1,10 @@
 package scaladex.client
 
+import scala.scalajs.js.timers
+
 import org.scalajs.dom.document
+import org.scalajs.dom.ext._
 import org.scalajs.dom.html.Element
-import org.scalajs.dom.html.LI
 import org.scalajs.dom.html.Link
 import scaladex.dom.IntersectionObserver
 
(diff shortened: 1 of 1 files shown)


modules/webclient/src/main/scala/scaladex/client/ActiveNavObserver.scala
```diff
@@ -11,20 +13,32 @@ import scaladex.dom.IntersectionObserver
...
+          val id = section.getAttribute("id")
+          val link = document.querySelector(s"""nav li a[href="#$id"]""").asInstanceOf[Link]
+          Option(link).map(l => section -> l.parentElement)
+        }
 
-      if (entry.intersectionRatio > 0) navLI.classList.add("active")
-      else navLI.classList.remove("active")
+    val observer = IntersectionObserver { _ =>
+      // ignore the observed entry and update all sections instead
+      timers.setTimeout(0) {
```
sjrd, line 30: So I looked at the solution at https://stackoverflow.com/a/65008310/1829647 and now I understand better why they use a `setTimeout`. They actually use it in conjunction with a `clearTimeout`, with the same handle `debounceTimeout`. With the 150 milliseconds, this effectively achieves a *debounce* mechanism: if the intersection observer fires multiple times within 150 ms, the recomputation will only happen once, after it has "stabilized" for at least 150 ms. I think it would be good to take this mechanism and implement here.

You'll have to declare `debounceTimeout` as a
```scala
var debounceTimeout: js.timers.SetTimeoutHandle = null
```

sjrd (changes requested):

modules/webclient/src/main/scala/scaladex/client/ActiveNavObserver.scala
```diff
@@ -11,18 +13,34 @@ import scaladex.dom.IntersectionObserver
...
+          Option(link).map(l => section -> l.parentElement)
+        }
 
-    if (entry.intersectionRatio > 0) navLI.classList.add("active")
-    else navLI.classList.remove("active")
+    var debounceUpdate: timers.SetTimeoutHandle = null
+    val observer = IntersectionObserver { _ =>
+      // ignore the observed entry and update all sections instead
+      debounceUpdate = timers.setTimeout(150) {
+        timers.clearTimeout(debounceUpdate)
```
sjrd, line 32: Hum, IIUC the `clearTimeout` must come before `debounceUpdate = setTimeout` (but inside the lambda passed to `IntersectionObserver`), shouldn't it?

adpi2: :man_facepalming:

sjrd (approved):

PR #949, opened 2022-03-11
When the element is hidden, it's still rendered

mlachkar:
During the rendering, you can't click on the button that are below the element.


adpi2 (approved):
Thanks!

PR #950, opened 2022-03-13
Adding user sessions to database

### LINKED ISSUE #947: Store sessions in database
## Motivation

Each time we re-deploy Scaladex, all users are logged out. To avoid this we can store the sessions in the database.

## Identified Obstacle

Serialize/Deserialize `UserState`.

## Implementation Guideline

1. Add a `user_sessions` table in the database with a [migration file](https://github.com/scalacenter/scaladex/tree/main/modules/infra/src/main/resources/migrations)
2. Add `insertSession(userId: UUID, userState: UserState)` and `getSession(userId: UUID): Futuer[Option[UserState]]` in `WebDatabase` and implement them in `SqlDatabase`.
4. Use these methods in `GithubUserSession` to save and retrieve the `UserState`

## Expectations

1. The new methods in `SqlDatabase` should be tested in `SqlDatabaseTests`

## Progress

- [x] Remove `env` from the `UserInfo` class and pass as arguments.
- [x] Add a DB migration for adding the `UserSessions` table.
- [ ] Add the `insert` and `get` Sessions functions to `WebDatabase` and implement them in `SqlDatabase`

jyoo980:
Should eventually resolve #947


adpi2 (commented):
I am glad you are starting a new contribution :slightly_smiling_face: 

Here are some indications.

jyoo980:
@adpi2 I think both removing `env` from `UserInfo` and introducing new logic around auth/environments (user sessions) in a single PR might not be such a good idea in case something goes wrong and we need to localize an error. I'll open a new PR just to handle the removal logic.

adpi2 (approved):
Thanks!

jyoo980:
Hey @adpi2, coming back to this since it's the weekend. 

I'm currently working on serializing/deserializing the `UserState`, I have something like 

```scala
object UserSessionsTable {

  private[sql] val table = "user_sessions"

  private val userId = "user_id"
  private val userSessionFields = Seq("repos", "orgs")
  private val userInfoFields = Seq("login", "name", "avatar_url", "secret")
  private val allFields = userId +: (userSessionFields ++ userInfoFields)

  val insertOrUpdate: Update[(UUID, UserState)] =
    insertOrUpdateRequest(table, allFields, allFields)
}
```

I think the next step would be to create a `Write[(UUID, UserState)]` instance, like

```scala
implicit val userSessionWrite: Write[(UUID, UserState)] =
      Write[(String, Option[String], Option[String], String, Option[String], String, String)].contramap {
        case (uuid, UserState(repos, orgs, UserInfo(login, maybeName, avatarUrl, secret))) => ???
      }
```

But maybe I am misunderstanding something. The issue I'm running into is I'm not entirely sure how to serialize the `repos` and `orgs`. Of course, one could do something like `repos.map(_.toString).mkString(",")`, but that doesn't seem quite right.

jyoo980:
@adpi2 I'm going to close this PR and work on the required changes incrementally. I'll add checkboxes to the #947 to track progress.

PR #959, opened 2022-03-17
Add sbt-github-dependency-graph

adpi2:


.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,20 @@
+name: Dependency Graph
+on:
+  push:
+    branches:
+      - main
```
adpi2, line 5: We cannot activate it on pull requests because the GITHUB_TOKEN permission is `read` in pull requests

PR #968, opened 2022-03-26
Adding user_sessions table + tests

jyoo980:
Sorry about the lack of progress on this, been busy with school.

  * Created `Meta[_]` instances, and implicit readers and writers for `UserInfo`
    and `UserState`
  * Test behaviour seems a bit off... `login` is being reported as an
    `Option[String]` when the SQL schema is `NOT NULL`.
  * Also seems to want me to filter out the "secret" key from the `select`,
    which doesn't make any sense.

Part of work for #947

### DIFF
--- a/modules/infra/src/main/resources/migrations/V8__add_user_sessions.sql
+++ b/modules/infra/src/main/resources/migrations/V8__add_user_sessions.sql
@@ -1,0 +1,10 @@
+CREATE TABLE user_sessions(
+    user_id       VARCHAR NOT NULL,
+    repos         VARCHAR NOT NULL,
+    orgs          VARCHAR NOT NULL,
+    login         VARCHAR(39) NOT NULL,
+    name          VARCHAR,
+    avatar_url    VARCHAR NOT NULL,
+    secret        VARCHAR NOT NULL,
+    PRIMARY KEY (user_id)
+);
--- a/modules/infra/src/main/scala/scaladex/infra/sql/DoobieUtils.scala
+++ b/modules/infra/src/main/scala/scaladex/infra/sql/DoobieUtils.scala
@@ -1,6 +1,7 @@
 package scaladex.infra.sql
 
 import java.time.Instant
+import java.util.UUID
 
 import scala.concurrent.ExecutionContext
 import scala.util.Try
@@ -28,6 +29,9 @@ import scaladex.core.model.Project
 import scaladex.core.model.Project._
 import scaladex.core.model.Resolver
 import scaladex.core.model.SemanticVersion
+import scaladex.core.model.UserInfo
+import scaladex.core.model.UserState
+import scaladex.core.util.Secret
 import scaladex.infra.Codecs._
 import scaladex.infra.config.PostgreSQLConfig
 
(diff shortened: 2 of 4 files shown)


jyoo980:
Currently, the `UserSessionsTable` test fails with the output:

```
- check selectUserSessionById *** FAILED ***
 Query[java.util.UUID, scaladex.core.model.UserState] defined at (source location unknown)
   SELECT user_id, repos, orgs, login, name, avatar_url, secret FROM
   user_sessions WHERE user_id=?
   ✓ SQL Compiles and TypeChecks
   ✓ P01 String  →  VARCHAR (text)
   ✓ C01 user_id    VARCHAR (varchar) NOT NULL  →  String
   ✓ C02 repos      VARCHAR (varchar) NOT NULL  →  String
   ✓ C03 orgs       VARCHAR (varchar) NOT NULL  →  String
   ✓ C04 login      VARCHAR (varchar) NOT NULL  →  Option[String]
   ✕ C05 name       VARCHAR (varchar) NULL      →  String
     Reading a NULL value into String will result in a runtime failure.
     Fix this by making the schema type NOT NULL or by changing the
     Scala type to Option[String]
   ✓ C06 avatar_url VARCHAR (varchar) NOT NULL  →  String
   ✕ C07 secret     VARCHAR (varchar) NOT NULL  →
     Column is unused. Remove it from the SELECT statement. (Checker.scala:56)
```

I feel like there is something here that I am misunderstanding:

  * Why is `login` being reported as an `Option[String]`, where I represent it in the implicit as a `String` and in the SQL as a `NOT NULL` type? I would expect `name` to be an `Option[String]` since it's not defined as a `NOT NULL` type.

  * The `secret` column appears to be unused, but this should not be the case. That failure goes away when I write `allFields.filterNot(_ == "secret")` in  [this line of code](https://github.com/scalacenter/scaladex/pull/968/files#diff-edbc704540104eb1f23f76c23ed18cfe26b0dfe44f0c2fb5756520efebf1fa32R24) but that should not be needed (we _want_ the secret to be read).

 
Any suggestions?

jyoo980:
For people in the future who may run into the same issue above, this was resolved by correctly setting the fields to be read [here](https://github.com/scalacenter/scaladex/pull/968/files#diff-edbc704540104eb1f23f76c23ed18cfe26b0dfe44f0c2fb5756520efebf1fa32R24).

Previously, this was implemented via:

```scala
val selectUserSessionById: Query[UUID, UserState] =
     selectRequest(table, allFields, Seq("user_id"))
```

This was not correct, as this introduces an additional field `user_id` to be read using the implicit readers defined for `UserInfo` and `UserState`. This effectively "shifts" the set of fields to be read down by 1 (this is why the testing framework was reporting `Secret` as being an unused column – it was replaced with `user_id`!) To mitigate this, we instead write:

```scala
val selectUserSessionById: Query[UUID, UserState] =
     selectRequest(table, userStateFields ++ userInfoFields, Seq("user_id"))
```

Which set the correct number of fields to be read.

adpi2 (approved):
Thanks

PR #972, opened 2022-03-28
Add a scheduler to retrieve missing release dates for artifacts

mlachkar:
Maybe we should update the small-index once we have all the release dates!


adpi2 (approved):
Looks good! Thanks!

PR #973, opened 2022-03-28
Load and set user sessions in Server

### LINKED ISSUE #947: Store sessions in database

jyoo980:
This work removes the locally-stored user sessions in `GithubSession.scala` and replaces it with a call to the SQL database to read and write user sessions.

This also involves having to make calls in `Server.scala` and `SearchApi.scala` to load the user sessions. Unfortunately, this currently involves the use of `Await`, which I don't think is a good idea, but could not find an alternative. Maybe someone more experienced could take a look?

Eventually closes #947

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/Server.scala
+++ b/modules/server/src/main/scala/scaladex/server/Server.scala
@@ -146,7 +146,7 @@ object Server extends LazyLogging {
     import actor.dispatcher
 
     val githubAuth = new GithubAuthImpl()
-    val session = new GithubUserSession(config.session)
+    val session = new GithubUserSession(config.session, webDatabase)
 
     val searchPages = new SearchPages(config.env, searchEngine)
     val frontPage = new FrontPage(config.env, webDatabase, searchEngine)
--- a/modules/server/src/main/scala/scaladex/server/Server.scala
+++ b/modules/server/src/main/scala/scaladex/server/Server.scala
@@ -1,6 +1,4 @@
 package scaladex.server
-
-import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
@@ -167,15 +165,15 @@ object Server extends LazyLogging {
     )
     import session.implicits._
     val userFacingRoute =
-      optionalSession(refreshable, usingCookies) { userId =>
-        val futureMaybeUser = userId.traverse(session.getUser).map(_.flatten)
+      optionalSession(refreshable, usingCookies) { maybeUserId =>
+        val futureMaybeUser = maybeUserId.traverse(session.getUser).map(_.flatten)
         val futureRoute = futureMaybeUser.map { maybeUser =>
           frontPage.route(maybeUser) ~ adminPages.route(maybeUser) ~ awesomePages.route(maybeUser) ~
             redirectToNoTrailingSlashIfPresent(StatusCodes.MovedPermanently) {
               projectPages.route(maybeUser) ~ searchPages.route(maybeUser)
             }
         }
-        Await.result(futureRoute, Duration.Inf)
+        context => futureRoute.flatMap(_.apply(context))
       }
     val exceptionHandler = ExceptionHandler {
       case ex: Exception =>
--- a/modules/server/src/main/scala/scaladex/server/Server.scala
+++ b/modules/server/src/main/scala/scaladex/server/Server.scala
@@ -1,5 +1,6 @@
 package scaladex.server
 
+import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
@@ -12,6 +13,7 @@ import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server._
 import cats.effect.ContextShift
 import cats.effect.IO
+import cats.implicits.toTraverseOps
 import com.softwaremill.session.SessionDirectives._
 import com.softwaremill.session.SessionOptions._
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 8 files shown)


modules/server/src/main/scala/scaladex/server/Server.scala
```diff
@@ -164,13 +166,16 @@ object Server extends LazyLogging {
...
-            projectPages.route(user) ~ searchPages.route(user)
-          }
+        val futureMaybeUser = userId.traverse(session.getUser).map(_.flatten)
+        val futureRoute = futureMaybeUser.map { maybeUser =>
+          frontPage.route(maybeUser) ~ adminPages.route(maybeUser) ~ awesomePages.route(maybeUser) ~
+            redirectToNoTrailingSlashIfPresent(StatusCodes.MovedPermanently) {
+              projectPages.route(maybeUser) ~ searchPages.route(maybeUser)
+            }
+        }
+        Await.result(futureRoute, Duration.Inf)
```
jyoo980, line 178: The code above expects an `Option[UUID] => Route`, but there's currently no way I can think of to call `GithubSession#getUser`, which evaluates to a `Future[_]` and then chain it to a raw `Route` type... any ideas?

adpi2: The definition of `Route` in akka is:
```scala
type Route = RequestContext => Future[RouteResult]
```

Here we have a `Future[Route]` and we can transform it into a route with:
```scala
context => futureRoute.flatMap(_.apply(context))
```

modules/server/src/test/scala/scaladex/server/route/ControllerBaseSuite.scala
```diff
@@ -22,7 +29,26 @@ trait ControllerBaseSuite extends AsyncFunSpec with Matchers with ScalatestRoute
...
+  val transactor: Transactor.Aux[IO, Unit] =
+    Transactor
+      .fromDriverManager[IO](
+        databaseConfig.driver,
+        databaseConfig.url,
+        databaseConfig.user,
+        databaseConfig.pass.decode
+      )
+
+  val sqlDatabase = new SqlDatabase(databaseConfig, transactor)
```
jyoo980, line 49: @adpi2 

Looking into the [failing unit tests](https://github.com/scalacenter/scaladex/runs/5727364612?check_suite_focus=true), I think it might be because of these lines.

I copied over this logic from here

https://github.com/scalacenter/scaladex/blob/86357e5846f7e79448d80d50a549e3a250e3db46/modules/infra/src/test/scala/scaladex/infra/BaseDatabaseSuite.scala#L18-L19

Because I didn't think extending this trait in a test suite that's not directly testing the database was a wise choice.

adpi2: Ok I was surprised by the failing tests but now I understand better. The problem is that we don't start the PostgreSQL container when running `server / test`.

The current design for the test is:
- the database and other third-parties (`GithubClient`, `ElasticSearchEngine`...) are unit-tested in `infra / test` and `infra / IntegrationTest / test`
- we use some mock implementations in `server / test`

The mock for the database is `InMemoryDatabase`. The mock for the `GithubAuth` is `MockGithubAuth`. I think you should be able to run all the tests in `server / test` by adapting those mocks instead of plugging in a real database and calling the real Github API.

modules/server/src/test/scala/scaladex/server/route/ControllerBaseSuite.scala
```diff
@@ -22,7 +29,26 @@ trait ControllerBaseSuite extends AsyncFunSpec with Matchers with ScalatestRoute
...
+
+  val githubUserSession = new GithubUserSession(config.session, sqlDatabase)
   val githubAuth = MockGithubAuth
 
   val database: InMemoryDatabase = new InMemoryDatabase()
```
adpi2, line 54:
```suggestion
  val githubAuth = MockGithubAuth
  val database: InMemoryDatabase = new InMemoryDatabase()
   val githubUserSession = new GithubUserSession(config.session, database)
```

modules/server/src/test/scala/scaladex/server/route/ControllerBaseSuite.scala
adpi2: You should see my suggestion below

modules/server/src/test/scala/scaladex/server/route/ControllerBaseSuite.scala
adpi2: Actually this is enough to make the tests pass, there is nothing to change in `InMemoruDatabase` or `MockGithubAuth`.
I am pushing the change.

adpi2 (approved):
Thanks for the great contribution.

adpi2:
Deployed successfully in production environment.

jyoo980:
Thanks for helping out! I really appreciate it

PR #974, opened 2022-03-28
Remove withSession from SearchEndpoints

adpi2:
Instead we move it to Akka implementation

@jyoo980 this will solve [this issue](https://github.com/scalacenter/scaladex/pull/973#discussion_r836625521) in #973

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/route/api/SearchApi.scala
+++ b/modules/server/src/main/scala/scaladex/server/route/api/SearchApi.scala
@@ -2,32 +2,25 @@ package scaladex.server.route.api
 
 import scala.concurrent.ExecutionContext
 
-import akka.http.scaladsl.model.Uri
-import akka.http.scaladsl.server.Directive
-import akka.http.scaladsl.server.Directive1
 import akka.http.scaladsl.server.Route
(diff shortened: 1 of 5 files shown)


jyoo980:
Sound good, I'll merge it into my branch once this is on `main`

modules/server/src/main/scala/scaladex/server/route/api/SearchApi.scala
```diff
@@ -2,32 +2,25 @@ package scaladex.server.route.api
 
 import scala.concurrent.ExecutionContext
 
-import akka.http.scaladsl.model.Uri
-import akka.http.scaladsl.server.Directive
-import akka.http.scaladsl.server.Directive1
 import akka.http.scaladsl.server.Route
 import ch.megard.akka.http.cors.scaladsl.CorsDirectives._
-import com.softwaremill.session.SessionDirectives._
-import com.softwaremill.session.SessionOptions._
```
julienrf, line 11: Can we remove the dependency from the build?

adpi2: We still use it on the server side with akka-http

PR #980, opened 2022-03-30
Adding scheduler to update/remove user sessions

### LINKED ISSUE #977: Add a mechanism to update `UserState`
### Motivation

Currently the user state is loaded once at first login but then it is never updated.

To solve this issue I see two possible directions:
1. Update each time we receive a `userId` from the client. The response should not wait for the userState update because the update can be slow (~10 seconds).
2. Add a scheduled job that update all user states every once in while

I would suggest solution 2 because it is easier to monitor: we can easily track the number of updated user states and catch failures.

### Identified Obstacles

If the user revokes its token, the GET requests in `GithubClient` will fail. We should catch the failure and identify that the token is no longer valid and remove the corresponding user session from the database. It will automatically log the user out.

### Implementation Guideline

1. In `SchedulerService` create a new scheduler
2. The scheduler should load all `UserState` from the database
3. For each user it should use its token to create a `GithubClient`
4. It uses the `GithubClient` to load the new state
5. Update the state in the database

### Expectations

1. The scheduled job should run once in while and not consume too much resources (avoid too much parallelization)
2. The failure from `GithubClient` should distinguish between revoked token or some other failure.

### Tasks
- [ ] Add a database query that reads all user states from the database.
- [ ] Add scheduled job that uses the query above (ignore the failures as a first step).
- [ ] Identify revoked tokens and remove corresponding user states from the database.

jyoo980:
Should eventually resolve #977

~Tests still needed for `UserSessionSynchronizer.scala`, will work on this when I get a chance this weekend.~

Looks like the scheduler/synchronizer classes don't have tests, I could add them if you gave me some pointers on where to start? It seems a bit difficult, given that there's no directly observable result since most of the return values are of type `Future[Unit]`.


modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
```diff
@@ -0,0 +1,73 @@
...
+  private def createClientsForEach(sessions: Seq[(UUID, UserState)]): Seq[(UUID, GithubService)] =
+    sessions.map {
+      case (userId, userState) => (userId, new GithubClient(userState.info.token))
+    }
+
+  private def getUserInfo(userId: UUID, service: GithubService): Future[Option[(UUID, UserInfo)]] =
+    service.getUserInfo().map(userInfo => Option.apply((userId, userInfo)))
+      .recoverWith {
+        case _ =>
+          logger.info(s"Token for user with id: '$userId' is likely expired")
```
jyoo980, line 44: Not sure what the retention policy is like for logs, would this be GDPR-compliant? We're already storing the `userID` as a primary key in the database anyway, just wanted to be sure.

adpi2: We keep the logs for only one month. Not sure about GDPR compliance but I would say that a expired token is not a personal piece of information.

modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
```diff
@@ -0,0 +1,58 @@
...
+  private def updateUserSession(session: Session, service: GithubService): Future[Either[UUID, Session]] =
+    service
+      .getUserInfo()
+      .map { updatedUserInfo =>
+        session match {
+          case (userId, staleUserState) => Right((userId, staleUserState.copy(info = updatedUserInfo)))
+        }
+      }
+      .recoverWith {
+        case _ =>
```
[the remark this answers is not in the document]
adpi2: > I think GithubService#getUserInfo doesn't fail in a way that lets the consumer know why something failed, but perhaps I'm misunderstanding something.

That's right.

We cannot assume that all failures are due to revoked tokens because we can also have some network outage or the Github API could be unstable. In those cases we don't want the scheduler to delete all the sessions.

So we need to change the implementation of `GithubService#getUserInfo`.
I see two possible solutions:
1. In https://github.com/scalacenter/scaladex/blob/91932e3d98b3262f2ad1d1201a95e2e30fb76e10/modules/infra/src/main/scala/scaladex/infra/GithubClient.scala#L368 we throw a custom `HttpException(code, reason)` that we can then catch and inspect.
2. Or, we change the signature and implementation of  `GithubService#getUserInfo` to return a `Future[GithubResponse[UserInfo]]`

I let you choose the one you prefer.

adpi2 (commented):
> Looks like the scheduler/synchronizer classes don't have tests, I could add them if you gave me some pointers on where to start? It seems a bit difficult, given that there's no directly observable result since most of the return values are of type Future[Unit].

We never did any test for the scheduled jobs because it requires quite some effort:
- we often need a lot of data
- we need to call some external API (Github or Sonatype) or to mock them

We could do:
1. Unit tests by mocking the database and the external APIs
2. Integration tests (like the `RelevanceTests`) with real data and real APIs

For now I am okay if we don't test them, because those schedulers are not critical for the user-facing experience, and because we can monitor them in the admin page.

modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
```diff
@@ -0,0 +1,65 @@
...
+      service: GithubService
+  ): Future[Either[UUID, UserInfo]] =
+    service.getUserInfo().flatMap {
+      case GithubResponse.Ok(res)               => Future.successful(Right(res))
+      case GithubResponse.MovedPermanently(res) => Future.successful(Right(res))
+      case GithubResponse.Failed(code, errorMessage) if code == StatusCodes.Unauthorized.intValue =>
+        logger.info(s"Token for user with id: '$userId' is likely expired, with error: $errorMessage")
+        Future.successful(Left(userId))
+      case _ =>
+        Future.successful(Right(staleUserInfo))
```
jyoo980, line 58: I return the stale information here, since we assume that we want to revoke the token if, and only if the API response returns a 401 (unauthorized).

modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
```diff
@@ -0,0 +1,65 @@
...
+  type Session = (UUID, UserState)
+
+  override def run(): Future[Unit] =
+    for {
+      sessions <- database.getAllSessions()
+      updatedEitherUserSessions <- sessions.traverse { case (session, client) => updateUserSession(session, client) }
+      expiredSessionIds = updatedEitherUserSessions.collect { case Left(expiredSessionId) => expiredSessionId }
+      updatedUserSessions = updatedEitherUserSessions.collect { case Right(sessions) => sessions }
+      _ <- updatedUserSessions.traverse { case (userId, userState) => database.insertSession(userId, userState) }
+      _ <- expiredSessionIds.traverse(database.deleteSession)
```
adpi2, line 31: I would suggest to do the `database.insertSession` or `database.deleteSession` inside the `updateUserSession` method, because we don't need to wait for all the user sessions to start inserting (or deleting) them in the database. Also I think this would make the code a bit more concise.

modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
```diff
@@ -0,0 +1,65 @@
...
+            else session
+          })
+    }
+
+  private def getUserInfo(
+      userId: UUID,
+      staleUserInfo: UserInfo,
+      service: GithubService
+  ): Future[Either[UUID, UserInfo]] =
+    service.getUserInfo().flatMap {
```
adpi2, line 51: Sorry I should have catched this earlier: we also need to update the `repos` and `organizations` in `UserState`.

So maybe you should add a `getUserState` method in `GithubService` and call it instead of `getUserInfo`. There is one already in `GithubAuthImpl` that you can move to `GithubClient`.

modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
```diff
@@ -0,0 +1,65 @@
...
+  private def updateUserSession(session: Session): Future[Either[UUID, Session]] =
+    session match {
+      case (userId, staleUserState) =>
+        val service = new GithubClient(staleUserState.info.token)
+        getUserInfo(userId, staleUserState.info, service)
+          .map(_.map { updatedUserInfo =>
+            if (updatedUserInfo != staleUserState.info)
+              (userId, staleUserState.copy(info = updatedUserInfo))
+            else session
+          })
```
adpi2, line 43: What if you match on `GithubResponse` here and make the call to the database directly:

```suggestion
          service.getUserState().flatMap  {
            case GithubResponse.Ok(state) => database.insertSession(userId, state)
            case GithubResponse.MovedPermanently(state) => database.insertSession(userId, state)
            case GithubResponse.Failed(code, errorMessage) =>
              if (code == StatusCode.Unauthorized.intValue) {
                logger.info(s"Token for user with id: '$userId' is likely expired, with error: $errorMessage")
                database.deleteSession(userId)
              } else {
                logger.warn(s"Failed to get user state from Github because of $code: $errorMessage")
                Future.successful(()) // we cannot update the user state
              }
          }
```

adpi2 (commented):
Thanks for the recent changes. I hope you are okay with some more comments ^^

modules/server/src/main/scala/scaladex/server/service/UserSessionSynchronizer.scala
jyoo980: That sounds good. Here's what I have so far

```scala
  override def getUserState(token: Secret): Future[GithubResponse[UserState]] = {
    val permissions = Seq("WRITE", "MAINTAIN", "ADMIN")
    (for {
      response <- getUserInfo()
      userInfo <- response match {
        case GithubResponse.Ok(userInfo)               => Future.successful(userInfo)
        case GithubResponse.MovedPermanently(userInfo) => Future.successful(userInfo)
        case GithubResponse.Failed(code, errorMessage) =>
          Future.failed(new Exception(s"Request to GithubClient#getUserInfo failed with code: $code, reason: $errorMessage"))
      }
      userOrgs <- getUserOrganizations(userInfo.login).recover(_ => Seq.empty)
      userOrgRepos <- userOrgs
        .traverse(getOrganizationRepositories(userInfo.login, _, permissions))
        .recover(_ => Seq.empty)
        .map(_.flatten)
      userRepos <- getUserRepositories(userInfo.login, permissions).recover(_ => Seq.empty)
    } yield GithubResponse.Ok(UserState(userOrgRepos.toSet ++ userRepos, userOrgs.toSet, userInfo)))
      .recoverWith {
        // TODO: Figure out how to assign the correct code.
        case t: Throwable => GithubResponse.Failed()
      }
  }
```

As you can see, I'm having some issues mapping the error in the for-comprehension into the `.recoverWith { ... }` at the end. Maybe I'm overthinking this and this isn't that hard... but I think it's hard to get information about what specific type of error that was thrown.

I could imagine making the result of a the for-comp an `Either[A, UserState]`, where `A` contains specific information about the type of failure, and then transforming the `Either[A, UserState]` into a `GithubResponse[UserState]` at the end, but this seems like overkill... any thoughts?

jyoo980: I got this working


```scala
  override def getUserState(token: Secret): Future[GithubResponse[UserState]] = {
    val permissions = Seq("WRITE", "MAINTAIN", "ADMIN")
    for {
      response <- getUserInfo()
      eitherInfo = response match {
        case GithubResponse.Ok(userInfo)               => userInfo.asRight
        case GithubResponse.MovedPermanently(userInfo) => userInfo.asRight
        case failed: GithubResponse.Failed             => failed.asLeft
      }
      eitherUserOrgs <- eitherInfo.traverse(info => getUserOrganizations(info.login).recover(_ => Seq.empty))
      eitherUserOrgRepos <- (eitherUserOrgs, eitherInfo).traverseN { (userOrgs, info) =>
        userOrgs
          .traverse(getOrganizationRepositories(info.login, _, permissions))
          .map(_.flatten)
      }
      eitherUserRepos <- eitherInfo.traverse(info =>
        getUserRepositories(info.login, permissions).recover(_ => Seq.empty)
      )
    } yield (eitherInfo, eitherUserOrgs, eitherUserOrgRepos, eitherUserRepos)
      .mapN { (info, userOrgs, userOrgRepos, userRepos) =>
        GithubResponse.Ok(UserState(repos = userOrgRepos.toSet ++ userRepos, orgs = userOrgs.toSet, info = info))
      }
      .fold(identity, identity)
  }
```

But maybe it's not the most readable? with `.traverse`, `.traverseN`, and `.mapN`?

modules/infra/src/it/scala/scaladex/infra/GithubClientTests.scala
```diff
@@ -72,8 +72,8 @@ class GithubClientTests extends AsyncFunSpec with Matchers {
       yield issues shouldBe empty
   }
 
-  it("getUserInfo") {
-    for (_ <- client.getUserInfo())
+  it("getUserState") {
+    for (_ <- client.getUserState())
       yield succeed
   }
```
jyoo980, line 78: @adpi2 

I noticed that this test is now failing on CI.

Previously, `GithubClient#getUserInfo` didn't depend on:

  * `GithubClient#getOrganizationRepositories`
  * `GithubClient#getUserRepositories`  
  *  `GithubClient#getUserOrganizations`

Which seem to be [skipped on CI](https://github.com/scalacenter/scaladex/blob/ce3da52ced67f3fea339ea708f73f3b7414c2489/modules/infra/src/it/scala/scaladex/infra/GithubClientTests.scala#L90-L103).

Does this mean we should also skip this test on CI as well? Sorry for all the questions, I appreciate your help!

jyoo980, line 78: @adpi2 

I noticed that this test is now failing on CI.

Previously, `GithubClient#getUserInfo` didn't depend on:

  * `GithubClient#getOrganizationRepositories`
  * `GithubClient#getUserRepositories`  
  *  `GithubClient#getUserOrganizations`

Which seem to be [skipped on CI](https://github.com/scalacenter/scaladex/blob/ce3da52ced67f3fea339ea708f73f3b7414c2489/modules/infra/src/it/scala/scaladex/infra/GithubClientTests.scala#L90-L103).

Does this mean we should also skip this test on CI as well? Sorry for all the questions, I appreciate your help!

adpi2: Yes unfortunately we need a personnal access token to run those tests but we cannot use it inside the Github infrastructure.

Let's just test `getUserClient` and not `getUserState` (I pushed the change).

adpi2 (approved):
:tada:

PR #984, opened 2022-03-31
Hide artifacts of vigoo/zio-aws

adpi2:
The `vigoo/zio-aws` project has more than 190 000 published artifacts. That's almost 1/5 of all artifacts in Scaladex.

As consequence it consumes a lot of CPU to load the page of that project.

This temporary fix is to not load any artifacts for this particular project.

After the redesign of the project page we should be able to remove this temporary fix.


adpi2:
Already deployed in production: https://index.scala-lang.org/vigoo/zio-aws

adpi2:
superseded by #985

PR #990, opened 2022-04-01
Include ETag header in badge response

armanbilge:
According to https://github.com/github/markup/issues/224#issuecomment-33454537 this is required to prevent badge caching.
> Assets must include `Cache-Control: no-cache` and `ETag` headers. If a badge is not updating, then it means they are not properly setting these headers.

I believe we can use the `status` as the `ETag`, since this is the dynamic part of the reponse (i.e. the result of the version lookup). Everything else should be determined by the URL itself.


adpi2 (approved):
Thanks, that is very useful!

PR #991, opened 2022-04-02
GitHub info

mlachkar:


modules/server/src/main/scala/scaladex/server/service/PublishProcess.scala
```diff
@@ -34,7 +37,7 @@ class PublishProcess(
     database: WebDatabase,
     pomsReader: PomsReader,
     env: Env
-)(implicit ec: ExecutionContext)
+)(implicit ec: ExecutionContext, system: ActorSystem)
```
adpi2, line 40: Why do we need an `ActorSystem` here?

modules/core/shared/src/main/scala/scaladex/core/service/WebDatabase.scala
```diff
@@ -7,11 +7,13 @@ import scala.concurrent.Future
...
 import scaladex.core.model.Artifact
 import scaladex.core.model.ArtifactDependency
+import scaladex.core.model.GithubInfo
+import scaladex.core.model.GithubResponse
 import scaladex.core.model.Project
 import scaladex.core.model.UserState
 
 trait WebDatabase {
-  def insertArtifact(artifact: Artifact, dependencies: Seq[ArtifactDependency], time: Instant): Future[Unit]
+  def insertArtifact(artifact: Artifact, dependencies: Seq[ArtifactDependency], time: Instant): Future[Boolean]
```
adpi2, line 16: Can you add a comment to explain what this boolean means?

modules/infra/src/main/scala/scaladex/infra/migrations/V7_2__edit_platform_and_language.scala
```diff
@@ -24,12 +24,10 @@ class V7_2__edit_platform_and_language extends BaseJavaMigration with ScaladexBa
       (for {
         oldArtifacts <- run(xa)(selectArtifact.to[Seq])
         groupedArtifacts = oldArtifacts.grouped(10000).toSeq
-        numberOfArtifactsUpdated <- groupedArtifacts
+        _ <- groupedArtifacts
           .map(artifacts => run(xa)(updatePlatformAndLanguage.updateMany(artifacts.map(_.update))))
           .sequence
-        _ = logger.info(s"Updated ${numberOfArtifactsUpdated.sum} artifacts")
```
adpi2, line 30: So we can change the code of migration files that were already executed?

adpi2 (commented):
I left some minor comments but that looks great!

modules/server/src/main/scala/scaladex/server/service/PublishProcess.scala
mlachkar: for githubClient

modules/core/shared/src/main/scala/scaladex/core/service/WebDatabase.scala
mlachkar: Not my favourite trick.. I will add a comment

modules/infra/src/main/scala/scaladex/infra/migrations/V7_2__edit_platform_and_language.scala
mlachkar: yes!

mlachkar: I mean at least for the one with code!

adpi2: That's a good news! :slightly_smiling_face:

PR #992, opened 2022-04-02
Implementing the start of the `/artifacts` endpoint

### LINKED ISSUE #982: Add /artifacts endpoint in the programmatic API
### Motivation

The Scaladex API could be used to list the available artifacts for a particular Scala version and/or platform.

We can do so by adding a new endpoint:
```
/api/artifacts?platform=<...>&language=<...>
```
Where `platform` and `language` are optional parameters.

Examples:
- all artifacts: `/api/artifacts`
- all Scala 2.13 artifacts: `/api/artifacts?language=2.13`
- all Scala.js 1.x artifacts: `/api/artifacts?platform=sjs1`
- all Scala.js 1.x artifacts on Scala 3: `/api/artifacts?platform=sjs1&language=3`

It returns a list of `groupid` and `artifact-id` in json:
```json
{
  "items": [
    {
      "groupid": "ch.epfl.scala",
      "artifactid": "scala-debug-adapter_2.13"
    },
   {
      "groupid": "ch.epfl.scala",
      "artifactid": "scalafix_2.13"
   }
  ]
}
```

### Identified Obstacles

We will probably implement paging as a second step. But can it work without a paging?

### Implementation Guideline

1. Add  a new `ArtifactEndpoints` in `modules/core/shared/src/main/scala/scaladex/core/api` and define `ArtifactParams` and `ArtifactResponse`. We use `endpoints4s` to document the API.
2. Add a new method `getAllArtifacts(language: Option[Language], platform: Option[Platform])` in `WebDatabase` and implement it in the `SqlDatabase` class.
3. Add an `ArtifactApi` class in `server/src/main/scala/scaladex/server/route/api/` to define the akka-http route that implement the endoint in `ArtifactEndpoint` using the new method in the database. Configure this route in `Server`.
4. Also add the `ArtifactEndpoint` endpoint in `DocumentationRoutes` for documentation

### Expectations

1. The `getAllArtifacts` method and SQL requests should be tested in `SqlDatabaseTests` and `ArtifactTableTests`
2. The `ArtifactApi` route should be tested in a new `ArtifactApiTests` using the `ControllerBaseSuite` that contains a mock of the database.
3. The new endpoint should be deployed and it should appear in the documentation endpoint `/api/open-api.json` 

### Tasks

 - [x] Add  a new `ArtifactEndpoints` in `modules/core/shared/src/main/scala/scaladex/core/api` and define `ArtifactParams` and `ArtifactResponse`. We use `endpoints4s` to document the API.
 - [x] Add a new method `getAllArtifacts(language: Option[Language], platform: Option[Platform])` in `WebDatabase` and implement it in the `SqlDatabase` class.
   - [x] The methods above should be tested
- [x] Add an `ArtifactApi` class in `server/src/main/scala/scaladex/server/route/api/` to define the akka-http route that implement the endoint in `ArtifactEndpoint` using the new method in the database. Configure this route in `Server`.
- [x] Also add the `ArtifactEndpoint` endpoint in `DocumentationRoutes` for documentation

jyoo980:
* Artifact endpoints/params to use in making the actual `/artifacts` route.
   * Functions in `WebDatabase` that allow users to query the DB for artifacts matching a given language or platform.
     * Tests for the functions above.
   * Updated `Server.scala` with endpoints
   * Adding tests for the artifact endpoints

Should resolve #982 

Is there a way to check if pagination is something needed for this endpoint?


modules/core/shared/src/main/scala/scaladex/core/api/SearchEndpoints.scala
```diff
@@ -6,7 +6,7 @@ import endpoints4s.algebra
 // There definition is shared here, for consistency.
 trait SearchEndpoints extends algebra.Endpoints with algebra.JsonEntitiesFromSchemas {
 
-  implicit val response: JsonSchema[AutocompletionResponse] =
+  implicit val autocompletionResponse: JsonSchema[AutocompletionResponse] =
```
jyoo980, line 9: Renamed this for consistency, along with `params`.

Since `ApiDocumentation` extends both `SearchEndpoints` and `ArtifactEndpoint`, there would be an issue if both have fields named `param` and `response`.

adpi2: :+1

modules/server/src/main/scala/scaladex/server/route/api/ArtifactApi.scala
```diff
@@ -0,0 +1,42 @@
...
+    with server.JsonEntitiesFromSchemas {
+
+  val routes: Route =
+    cors() {
+      artifact.implementedByAsync {
+        case ArtifactParams(maybeLanguageStr, maybePlatformStr) =>
+          val maybeLanguage = maybeLanguageStr.flatMap(Language.fromLabel)
+          val maybePlatform = maybePlatformStr.flatMap(Platform.fromLabel)
+          for (artifacts <- database.getAllArtifacts(maybeLanguage, maybePlatform))
+            yield artifacts.map(artifact => ArtifactResponse(artifact.groupId.value, artifact.artifactId))
```
adpi2, line 30: We need to filter duplicates here because many artifacts can have the same groupid and artifactid (but different versions).
```suggestion
            yield artifacts.map(artifact => ArtifactResponse(artifact.groupId.value, artifact.artifactId)).distinct
```

modules/server/src/main/scala/scaladex/server/route/api/ArtifactApi.scala
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+}
+
+object ArtifactApi {
+
+  implicit val formatArtifactResponse: OFormat[ArtifactResponse] =
+    Json.format[ArtifactResponse]
```
adpi2, line 38: Do we need this?

modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactResponse.scala
```diff
@@ -0,0 +1,3 @@
+package scaladex.core.api.artifact
+
+final case class ArtifactResponse(groupId: String, artifactId: String)
```
mlachkar, line 3: It's a detail that can be changed later, but maybe we can also have the version of the artifact ?

mlachkar (approved):
looks good to me!

modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactEndpoints.scala
```diff
@@ -0,0 +1,31 @@
...
+  ) & qs[Option[String]](
+    name = "platform",
+    docs = Some("Filter the results matching the given platforms only (e.g., 'jvm', 'sjs1', 'native0.4', 'sbt1.0')")
+  )).xmap((ArtifactParams.apply _).tupled)(Function.unlift(ArtifactParams.unapply))
+
+  // Artifact endpoint definition
+  val artifact: Endpoint[ArtifactParams, Seq[ArtifactResponse]] =
+    endpoint(
+      get(path / "api" / "artifacts" /? artifactEndpointParams),
+      ok(jsonResponse[Seq[ArtifactResponse]])
```
adpi2, line 29: Currently the json  response looks like this.

```json
[
  {
    "groupid": "ch.epfl.scala",
    "artifactid": "scala-debug-adapter_2.13"
  },
 {
    "groupid": "ch.epfl.scala",
    "artifactid": "scalafix_2.13"
 }
]
```
That's perfect but we will probably need to add paging later on this endpoint and it will break the compatibility.
```json
{
  "currentPage": 1
  "pageCount": 10
  "totalSize": 20
  "items": [
    {
      "groupid": "ch.epfl.scala",
      "artifactid": "scala-debug-adapter_2.13"
    },
   {
      "groupid": "ch.epfl.scala",
      "artifactid": "scalafix_2.13"
   }
  ]
}
```

To make things easier for later I suggest to return `Page[ArtifactResponse]]` rather than `Seq[ArtifactResponse]` even if we don't implement paging right now (the first page contains all the `ArtifactResponse`).

adpi2 (commented):
Thanks for such productivity!

modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactResponse.scala
adpi2: I would rather create another endpoint to get the versions of a particular artifact. Otherwise this endpoint would return a lot of results (we have more than 1,000,000 artifacts in the database).

mlachkar: make sense!

jyoo980: I can open an issue to create an endpoint that gets the versions of a given artifact 👍

mlachkar: It can be interesting for a group_id/artifact_id

modules/server/src/main/scala/scaladex/server/route/api/ArtifactApi.scala
jyoo980: Yeah, otherwise the [unmarshalling in `ArtifactApiTests.scala`](https://github.com/scalacenter/scaladex/pull/992/files#diff-9523193ad631d9906f11f9ec3dab8b6a641973808d238dd90fde442adc6babb9R32) won't work.

modules/core/shared/src/main/scala/scaladex/core/api/PaginationSchema.scala
```diff
@@ -0,0 +1,16 @@
+package scaladex.core.api
+
+import endpoints4s.algebra.JsonEntitiesFromSchemas
+import scaladex.core.model.search.Pagination
+
+trait PaginationSchema extends JsonEntitiesFromSchemas {
```
jyoo980, line 6: Putting pagination schema into a trait so it can be mixed into other APIs that may use it in the future.

jyoo980:
The tests need an unmarshaller for `Page[ArtifactResponse]`, I can try to work on that this coming weekend. Other than that, this looks done

modules/server/src/main/scala/scaladex/server/route/api/ArtifactApi.scala
julienrf: If this is used in the tests only, I suggest defining it in the tests instead of the main artifact. Also, this definition is redundant with the definition `artifactResponseSchema`, and that could lead to inconsistencies (false-positive or, worse, false-negative test failures).

You can define a play-json codec from the `artifactResponseSchema` by applying the [json-schema-playjson](https://endpoints4s.github.io/interpreters/play-json.html) to it:

~~~ scala
object PlayJsonCodecs extends ArtifactSchemas with endpoints4s.playjson.JsonSchemas

PlayJsonCodecs.artifactResponseSchema.reads // : play.api.libs.json.Reads[ArtifactResponse]
~~~

(Assuming you would move the definition of `artifactResponseSchema` into a separate trait `ArtifactSchemas`, similar to what you did with `PaginationSchema`)

modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactEndpoints.scala
```diff
@@ -0,0 +1,40 @@
...
+      }
+
+  val artifactEndpointParams: QueryString[ArtifactParams] = (qs[Option[String]](
+    name = "language",
+    docs = Some(
+      "Filter the results matching the given language version only (e.g., '3', '2.13', '2.12', '2.11', 'java')"
+    )
+  ) & qs[Option[String]](
+    name = "platform",
+    docs = Some("Filter the results matching the given platforms only (e.g., 'jvm', 'sjs1', 'native0.4', 'sbt1.0')")
```
julienrf, line 31:
```suggestion
    docs = Some("Filter the results matching the given platform only (e.g., 'jvm', 'sjs1', 'native0.4', 'sbt1.0')")
```

modules/server/src/main/scala/scaladex/server/route/api/ApiDocumentation.scala
```diff
@@ -4,16 +4,21 @@ import endpoints4s.openapi
...
+    with openapi.Endpoints
+    with openapi.JsonEntitiesFromSchemas {
 
   val api: OpenApi = openApi(
     Info(
       title = "Scaladex API",
       version = "0.1.0"
     )
-  )(autocomplete)
+  )(autocomplete, artifact)
```
julienrf, line 23: :tada:

modules/server/src/test/scala/scaladex/server/util/PageReadProvider.scala
```diff
@@ -0,0 +1,14 @@
+package scaladex.server.util
+
+import play.api.libs.json.Json
+import play.api.libs.json.Reads
+import scaladex.core.model.search.Page
+import scaladex.core.model.search.Pagination
+
+object PageReadProvider {
```
jyoo980, line 8: This can be reused in future tests that may require a paginated response.

All the test would need to do is bring this `object` into scope, and then declare an implicit instance for whatever
response the `Page` wraps. In this. case, a [JSON reader for `ArtifactResponse`](https://github.com/scalacenter/scaladex/pull/992/files#diff-9523193ad631d9906f11f9ec3dab8b6a641973808d238dd90fde442adc6babb9R27)

modules/server/src/test/scala/scaladex/server/util/PaginatedResponseReader.scala
```diff
@@ -0,0 +1,14 @@
...
+import play.api.libs.json.Json
+import play.api.libs.json.Reads
+import scaladex.core.model.search.Page
+import scaladex.core.model.search.Pagination
+
+trait PaginatedResponseReader {
+
+  implicit def pageItemsJson[A](implicit items: Reads[A]): Reads[Seq[A]] = Reads.seq(items)
+  implicit val paginationJson: Reads[Pagination] = Json.reads[Pagination]
+  implicit def pageJson[A](implicit items: Reads[A]): Reads[Page[A]] = Json.reads[Page[A]]
```
jyoo980, line 12: The unit tests are failing before they're even running because of [this error](https://github.com/scalacenter/scaladex/runs/5958735189?check_suite_focus=true#step:5:503), which reports `items` here as being unused.

I think this is needed, otherwise `pageJson` fails to find an implicit value for `A`. 

I can mark this with a compiler annotation `@nowarn`, but I'd be interested to see if there's a way around this, maybe I'm not understanding implicits here correctly.

jyoo980:
Sorry for the delay on this, busy week for my thesis!

modules/server/src/test/scala/scaladex/server/util/PaginatedResponseReader.scala
julienrf: I recommend not to generate JSON codecs other than the ones you can get by interpreting the schemas defined in `PaginationSchema`. This ensures that the codecs you use for testing are the same as the codecs used in the app.

jyoo980: Thanks for the response. Could you clarify what you mean by 

>  interpreting the schemas defined in `PaginationSchema`

There's only a single schema defined there, which is `Pagination`. Wouldn't we still have to define a codec for the `Page` case class as well? Sorry for all the questions, I appreciate your help :)

julienrf: Hey, you are totally right, we also need to define a schema for `Page` :) Currently, how are pages serialized in HTTP responses?

Interpreting a schema means mixing an implementation trait into a trait that contains schema definitions:

~~~ scala
object PlayJsonSchemas extends PaginationSchema with endpoints4s.playjson.JsonSchemas

// And then
PlayJsonSchemas.paginationSchema.reads // : Reads[Pagination]
~~~

I just realized that the trait `PaginationSchema` extends `JsonEntitiesFromSchemas`, but it should only extend `JsonSchemas` to be able to apply the play-json interpreter as shown above.

Let me know if you have other questions.

jyoo980: Currently, `Page` is serialized like this in `ArtifactEndpoints`

```scala
  implicit val artifactEndpointResponse: JsonSchema[Page[ArtifactResponse]] =
    field[Pagination]("pagination")
      .zip(field[Seq[ArtifactResponse]]("items"))
      .xmap[Page[ArtifactResponse]] { case (pagination, artifacts) => Page(pagination, artifacts) } {
        case Page(pagination, artifacts) => (pagination, artifacts)
      }

```

I think I'm getting close with the `PlayJsonSchemas` object, 

```scala
trait PaginatedResponseReader {

  implicit val readPagination: Reads[Pagination] = PlayJsonCodecs.paginationSchema.reads

}
```

But I'm unsure of how to interpret a `Page[ArtifactResponse]`, since it's a generic case class. There's also the problem of [this line of code](https://github.com/scalacenter/scaladex/pull/992/files#diff-5cf5e83486098319404a7d13272635d19c70ffe281deaad2b5395c45e32a4cf9R38), which no longer compiles given that `PaginationSchema` no longer extends `JsonEntitiesFromSchemas`.

I guess `ArtifactEndpoints` itself could extend `JsonEntitiesFromSchemas`, but I think that leads to the problem of generate JSON codecs other than the ones you can get by interpreting the schemas defined in `PaginationSchema`, which is what we want to avoid in the first place.

build.sbt
```diff
@@ -221,6 +221,7 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       "io.github.cquiroz" %%% "scala-java-time" % "2.3.0",
       "com.typesafe.play" %%% "play-json" % V.playJsonVersion,
       "org.endpoints4s" %%% "algebra" % "1.6.0",
+      "org.endpoints4s" %% "json-schema-playjson" % "1.7.0",
```
julienrf, line 224:
```suggestion
      "org.endpoints4s" %% "json-schema-playjson" % "1.7.0" % Test,
```
?

modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactEndpointSchema.scala
```diff
@@ -0,0 +1,22 @@
...
+      .xmap[ArtifactResponse] { case (groupId, artifactId) => ArtifactResponse(groupId, artifactId) } {
+        case ArtifactResponse(groupId, artifactId) => (groupId, artifactId)
+      }
+
+  implicit val artifactEndpointResponse: JsonSchema[Page[ArtifactResponse]] =
+    field[Pagination]("pagination")
+      .zip(field[Seq[ArtifactResponse]]("items"))
+      .xmap[Page[ArtifactResponse]] { case (pagination, artifacts) => Page(pagination, artifacts) } {
+        case Page(pagination, artifacts) => (pagination, artifacts)
+      }
```
julienrf, line 21:
```suggestion
```

This is now redundant with `pageSchema`

modules/server/src/main/scala/scaladex/server/route/api/ArtifactApi.scala
```diff
@@ -0,0 +1,52 @@
...
+          for (artifacts <- database.getAllArtifacts(maybeLanguage, maybePlatform))
+            yield {
+              val distinctArtifacts =
+                artifacts.map(artifact => ArtifactResponse(artifact.groupId.value, artifact.artifactId)).distinct
+              // TODO: The values below are placeholders, will need to populate them w. real data.
+              // See: https://github.com/scalacenter/scaladex/pull/992#discussion_r841500215
+              Page(
+                pagination = Pagination(
+                  current = 0,
+                  pageCount = 0,
```
julienrf, line 38:
```suggestion
                  pageCount = 1,
```

I don’t know if pages start at 0 or 1, maybe `current` needs to be changed to `1` to?

julienrf (commented):
Thank you! The design looks good to me, I have left minor comments.

modules/server/src/main/scala/scaladex/server/route/api/ArtifactApi.scala
jyoo980: Looks like they start at 1, I'll update this 

https://github.com/scalacenter/scaladex/blob/d59165e8b31a19a9351df4bccdce83bb8ebb7245/modules/server/src/main/scala/scaladex/server/route/package.scala#L36

julienrf (approved):

PR #1002, opened 2022-04-14
Displaying commit activity on the Awesome page, the project page and the search page 

joke1196:


adpi2:
This is a really good start @joke1196

joke1196:
@adpi2 Is there a way I could update the `small index`? As I added a new field `commit_activity` all the JSON files used to populate the test DB cannot be decoded, and this will make the integration tests fail as well.

adpi2:
> @adpi2 Is there a way I could update the small index? As I added a new field commit_activity all the JSON files used to populate the test DB cannot be decoded, and this will make the integration tests fail as well.

I would have expected the decoder to accept the missing `commitActivity` since it can be an empty `Seq`. Is there a easy way to make that way?

Otherwise, if your database is already populated, you can update the small index by running the `data / run subIndex` task.
And if it is not then you can populate it using the main branch then run `data / run subIndex` in your branch.

adpi2:
> The last thing is I cannot reproduce the failling integration test. I ran several times infra / IntegrationTest / test, but the getCommitActivity test always succeeds. Not sure how to solve this issue.

It could be that the Github token of the CI does not have the right to read the commit activity of the Scalafix project. But when I rerun the CI the test passed so it is a bit mysterious.

adpi2:
I deployed this PR in the dev environment and the result can be seen here: https://index-dev.scala-lang.org/

It will take some time to download all the commit activity.

adpi2:
@joke1196 this is a great contribution! I really like the look of the sparkline in the project page. There are a few things that we may want to add before merging this. Here are my suggestions:

### 1. Graph legend

The sparkline is a great idea and the look is beautiful but currently there is no legend so it is impossible to know what it represents and the scale of it.

I would recommend to add a title and legend that look like this:
github-activity-2

### 2. Search and Awesome Scala

In the search page and the awesome Scala page, my feeling is that the the Github activity graph takes too much place. I think it should rather be a single stats, like the number of stars or the number of contributors. Getting the inspiration from the good old [awesome-scala](https://github.com/lauris/awesome-scala) repository, we could simply use the total number of commits per year. This would appear as a regular stat in the project result box:

search-result

It is good to have the detailed activity week by week in the `Project` entity, however, in the `ProjectDocument` entity I would only put the simpler field `commitsPerYear: Int`. Later we could use this field in the scoring fnuction of the search page and awesome page to give better score to the most active Scala projects.

Thanks again for your contribution. I'll make sure this is going to be released sooner rather than later. Please let me know how you stand about those 2 suggestions.

joke1196:
@adpi2 Great, glad that you like the overall look of it. So I was thinking more of it as a lifeline for the project this is why I didn't add any more information on the graph, in the sense that in a glimpse you get a feel if the repo is alive or just flatlining :). I could add some details of course; it would make a lot of sense on the project page. And sure for the search page and the awesome page, it is a good idea, it will declutter the UI a bit.

adpi2:
Oh yeah now I understand your idea of a lifeline. There is something similar in Github in the main page of an organization. For instance: https://github.com/com-lihaoyi.

So I still think we should have some details in the project page. For the other pages the scale of the vertical axis should be more or less constant so that a project with very few commits is almost completely flat. For instance in this [search page](https://index-dev.scala-lang.org/search?q=cats) the line of cats-retry should be flatter. It contains only one commit.

joke1196:
@adpi2 finally got to work on the chart. Tell me if this is ok for you. I went with the classical line chart approach.

adpi2:
Awesome @joke1196! I deployed the branch to the dev environment: https://index-dev.scala-lang.org/.

We should have all the commit_activity populated in about 9 hours.

modules/webclient/src/main/scala/scaladex/client/Sparkline.scala
```diff
@@ -37,7 +37,7 @@ object Sparkline {
             backgroundColor = fillGrad,
             borderColor = grad,
             borderWidth = 1,
-            fill = "shape"
+            fill = "origin"
```
adpi2, line 40: :+1:

modules/core/shared/src/main/scala/scaladex/core/model/GithubInfo.scala
```diff
@@ -61,7 +63,8 @@ case class GithubInfo(
       forks = forks,
       contributorCount = contributorCount,
       scalaPercentage = scalaPercentage,
-      license = license
+      license = license,
+      commitsPerYear = Option(commitActivity.map(_.total).sum)
```
adpi2, line 67: It would be better to have `None` instead of `Some(0)` when the `commitActivity` is not known.

```suggestion
      if (commitActivity.isEmpty) None else Some(commitActivity.map(_.total).sum)
```

modules/infra/src/main/scala/scaladex/infra/elasticsearch/ElasticsearchMapping.scala
```diff
@@ -74,6 +74,7 @@ object ElasticsearchMapping {
     objectField("formerReferences").fields(
       textField("organization").analyzer("standard"),
       textField("repository").analyzer("standard")
-    )
+    ),
+    nestedField("githubInfo.commitActivity")
```
adpi2, line 78: Now it is `commitsPerYear: Option[Int]`
```suggestion
    intField("githubInfo.commitsPerYear")
```

modules/server/src/main/assets/css/partials/_awesome.scss
```diff
@@ -199,10 +198,27 @@
...
+				justify-self:end;
+				grid-column: 3;
+			}
+			.commit-activity-container {
+				max-height:20px;
+				justify-self:end;
+			}
+			canvas.commit-activity {
+				max-width:100%;
+			}
```
adpi2, line 220: Don't need that anymore.
```suggestion
```

modules/server/src/main/assets/css/partials/_awesome.scss
```diff
@@ -315,10 +334,32 @@
...
         @media screen and (min-width: $screen-md-min) {
           text-align: right;
         }
       }
+			.commit-activity-container {
+				max-height:20px;
+			}
+			canvas.commit-activity {
+				max-width:100%;
+			}
```
adpi2, line 362: I don't think we need that either.

```suggestion
```

modules/template/src/main/twirl/scaladex/view/project/project.scala.html
```diff
@@ -68,10 +75,22 @@ <h4>Statistics</h4>
   </div>
 }
 
+@commitActivities = @{project.githubInfo.map(_.commitActivity).getOrElse(Seq.empty).sortBy(_.week)}
+@startingDay = @{commitActivities.headOption.map(_.week.getEpochSecond)}
+@lastDay = @{commitActivities.lastOption.map(_.week.plus(7, ChronoUnit.DAYS).getEpochSecond)}
+@commitActivityBox = {
+	@if(commitActivities.nonEmpty){
+		<div class="box commit-activities">
+			<h4>Activity</h4>
```
adpi2, line 84:
```suggestion
			<h4>Commit Activity</h4>
```

adpi2 (commented):
This is great! I left some minor suggestions and I am looking forward to merging it.

joke1196:
Thank you for the review! I did the corrections as well as a small clean up.

adpi2 (approved):
Thanks @joke1196

Issue #869, opened 2022-01-19
Schedule job to update claimed artifacts


adpi2:
## Motivation

Some artifacts are published without the `scmInfo` field in their pom files which we use to map artifacts to github repositories.

To handle these we have a `scaladex-contrib` project in which library authors can claim that they own some artifacts.

We need to run a scheduled job to take this information into account and update artifacts that are claimed.

## Identified Obstacles

Currently artifacts with no project (github repository) are not indexed. We now need to index them so that the scheduler can find them in the database.

## Implementation guidelines

1. Add a method in `LocalStorage` that can load the `claims.json` file from the `scaladex-contrib` folder.
2. Add methods in `SchedulerDatabase` to load all the claimed artifacts and update their corresponding projects.
3. Schedule a job in the `SchedulerService` to update the claimed artifacts.

As a bonus the scheduler can also fetch the `scaladex-contrib` main branch from Github.  

## Requirements

1. The new methods in `LocalStorage` and `SchedulerDatabase` should be tested.

jyoo980:
I can work on indexing artifacts with no associated projects

jyoo980:
Quick question for this, it looks like a number of methods in `Artifact.scala` (and beyond!) operate on the assumption that `projectRef` is always defined, e.g.,

https://github.com/scalacenter/scaladex/blob/22f1672a0582da10036e8d2ddbfb0388a1edb4cc/modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala#L38-L43

But we know this is no longer going to be the case. Do you have any suggestions on how this should be handled? Maybe for the url-specific ones, we can point to a 404 page (if one exists)

adpi2:
Those methods should probably return `Option`. They are called in the [twirl files](https://github.com/scalacenter/scaladex/tree/main/modules/template/src/main/twirl/scaladex/view) (the front-end generator) to create links between pages. So then we need to deal with the `None` case. One way is to not display the `<a>` element if the url is `None`:

```html
@for(url <- artifact.fullHttpUrl) { <a href="@url"> ... </a>}
```

By the way, I think many of those methods will soon disappear because of the [new project page](https://github.com/scalacenter/scaladex/pull/945) we are currently working on with @mlachkar. I am going to work on it today. It should be merged later this week or maybe next week.

jyoo980:
Sounds good, I'll wait :)

PR #1008, opened 2022-04-15
Adding `/api/artifacts/<group_id>/<artifact_id>` endpoint

### LINKED ISSUE #998: Add endpoint `api/artifacts/<group_id>/<artifact_id>` to obtain metadata for a given artifact
## Summary

Currently, the `api/artifacts` endpoint returns a list of all artifacts with the response below:

```
{
  "items": [
    {
      "groupid": "ch.epfl.scala",
      "artifactid": "scala-debug-adapter_2.13"
    },
    {
       "groupid": "ch.epfl.scala",
       "artifactid": "scalafix_2.13"
    }
  ]
}
```

We now want to add an endpoint that enables clients to get the [semantic version](https://github.com/scalacenter/scaladex/blob/main/modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala#L22) of an artifact given its `groupId` and `artifactId`. 

The endpoint should have the form:

```
api/artifacts/<group_id>/<artifact_id>
```

It should respond with a value of type `Page[ArtifactMetadata]`, where the metadata type has the following information

```
// ArtifactMetadata object for cats-core
{
  "version": "2.7.0",
  "project_reference": "typelevel/cats",
  "release_date": "Nov 28, 2021",
  "language": "Scala 2.13",
  "platform": "JVM"
}
```

## Tasks

- [x] Add a method in `WebDatabase` to select an artifact, given a `groupId` and an `artifactId`
  - [x] This method should be tested in `SqlDatabaseTests` and `ArtifactTableTests`
- [x] Create an endpoint in `ArtifactEndpoints.scala` for `api/artifacts/<group_id>/<artifact_id>`
  - [x] This method should be tested in `ArtifactApiTests.scala`

## Other Details
This work is the result of [this discussion](https://github.com/scalacenter/scaladex/discussions/997).

jyoo980:
## Summary of Work

  * [`WebDatabase#getArtifacts(groupId, artifactId): Future[Seq[Artifact]]`](https://github.com/scalacenter/scaladex/compare/yoo/add-artifact-data-endpoint?expand=1#diff-2ce2689f5215078761c39afa95acdd451e9be2261b8c1da5d532a75ded5936ceR25), with associated tests in `ArtifactTestTable.scala` and `SqlDatabaseTests.scala`.
  * Defined the `/api/artifacts/<group_id>/<artifact_id>` endpoint in `ArtifactEndpoints`, added them to `ArtifactApi.scala`, with associated tests.

Closes #998

### DIFF
--- a/modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactMetadataResponse.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactMetadataResponse.scala
@@ -1,0 +1,9 @@
+package scaladex.core.api.artifact
+
+final case class ArtifactMetadataResponse(
+    version: String,
+    projectReference: Option[String],
+    releaseDate: String,
+    language: String,
+    platform: String
+)
--- a/modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactEndpointSchema.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactEndpointSchema.scala
@@ -10,4 +10,14 @@ trait ArtifactEndpointSchema extends PaginationSchema {
       .xmap[ArtifactResponse] { case (groupId, artifactId) => ArtifactResponse(groupId, artifactId) } {
         case ArtifactResponse(groupId, artifactId) => (groupId, artifactId)
       }
+
+  implicit val artifactMetadataResponseSchema: JsonSchema[ArtifactMetadataResponse] =
+    field[String]("version")
+      .zip(optField[String]("projectReference"))
+      .zip(field[String]("releaseDate"))
+      .zip(field[String]("language"))
+      .zip(field[String]("platform"))
+      .xmap[ArtifactMetadataResponse](ArtifactMetadataResponse.tupled)(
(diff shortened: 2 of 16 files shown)


modules/core/shared/src/main/scala/scaladex/core/api/artifact/ArtifactMetadataResponse.scala
```diff
@@ -0,0 +1,9 @@
+package scaladex.core.api.artifact
+
+final case class ArtifactMetadataResponse(
+    version: String,
+    projectReference: Option[String],
```
jyoo980, line 5: This field in practice is always `Some`, is there a reason why you want it to be `None`?

I think there's work planned [in #869](https://github.com/scalacenter/scaladex/issues/869) that would index artifacts that don't have an associated project – are we just planning for this?

adpi2: You are 100% right. Currently the field can only be `Some` and we plan that it can be `None` in #869.

adpi2 (approved):
Looks very good! Thanks!

PR #1013, opened 2022-04-26
Index artifacts with no associated projects

### LINKED ISSUE #869: Schedule job to update claimed artifacts

jyoo980:
This is part of the general work for #869.

Specifically, this changeset enables the storage of artifacts that have no associated projects.

Helps resolve #869


jyoo980:
Does not compile yet, just pushing up incremental work to get some eyes on it before I might waste time going down a bad path.

modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
```diff
@@ -48,11 +49,11 @@ class SqlDatabase(datasource: HikariDataSource, xa: doobie.Transactor[IO]) exten
...
     val unknownStatus = GithubStatus.Unknown(time)
     for {
-      isNewProject <- insertProjectRef(artifact.projectRef, unknownStatus)
+      maybeIsNewProject <- artifact.projectRef.traverse(ref => insertProjectRef(ref, unknownStatus))
       _ <- run(ArtifactTable.insertIfNotExist(artifact))
-      _ <- run(ReleaseTable.insertIfNotExists.run(artifact.release))
+      _ <- artifact.release.traverse(release => run(ReleaseTable.insertIfNotExists.run(release)))
       _ <- insertDependencies(dependencies)
-    } yield isNewProject
+    } yield maybeIsNewProject.fold(true)(identity)
```
jyoo980, line 56: @adpi2 I hope you don't mind me tagging you for some early feedback.

This is the part that I'm most iffy about. What do we consider a new artifact now, with the ability to store artifacts that don't _immediately_ have an associated project?

The result of this `maybeIsNewProject` is used in [a conditional](https://github.com/scalacenter/scaladex/pull/1013/files#diff-8c8a9fdd0e802d7eeb93efa0e7420ad297fe446a0b2eb77639ae84c1ff43179fR89) later to determine whether to update the GitHub project information.

modules/server/src/main/scala/scaladex/server/service/PublishProcess.scala
```diff
@@ -65,35 +61,50 @@ class PublishProcess(
...
+  }
+
+  private def insertArtifactWithDependencies(
+      artifact: Artifact,
+      dependencies: Seq[ArtifactDependency],
+      maybeProjectRef: Option[Project.Reference],
+      maybeUserState: Option[UserState]
+  ): Future[Unit] =
+    database.insertArtifact(artifact, dependencies, Instant.now).flatMap { isNewProject =>
+      if (isNewProject && maybeProjectRef.isDefined && maybeUserState.isDefined)
```
jyoo980, line 89: I imagine this logic will have to change to account for some other cases, e.g.

What if `isNewProject = false` but `maybeProejctRef.isDefined = true`? In this case, we should update the GitHub information, correct?

modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
adpi2: It looks okay to return false if an artifact has no associated project.

modules/server/src/main/scala/scaladex/server/service/PublishProcess.scala
adpi2: > What if isNewProject = false but maybeProejctRef.isDefined = true? In this case, we should update the GitHub information, correct?

Updating the GithubInfo is expensive because we have a limited rate of requests to the Github API. So here we only want to do it on new projects. For the existing projects we have a scheduled job that runs slowly in the background every day to update the GithubInfo.

PR #1042, opened 2022-06-21
Adding Scala CLI dependency install instructions

### LINKED ISSUE #1016: Add a Scala CLI tab in the install box of an artifact
### Motivation

In the artifact page, for instance https://index.scala-lang.org/typelevel/cats/artifacts/cats-core/2.7.0, the install box should contain a Scala CLI tab.

### Identified Obstacles

_No response_

### Implementation Guideline

- Add a method in `Artifact` that return the install instruction for Scala CLI (as in https://scala-cli.virtuslab.org/docs/reference/directives#examples-3)
- Call this method in the `artifact.scala.html` twirl template to display the infromation in a new tab of the install box

### Expectations

- The new tab should appear between Mill and Ammonite
- It should give correct information that can easily be copy pasted in a Scala CLI script

jyoo980:
The syntax for installing an artifact via Scala CLI now appears between
the Mill and Ammonite install instructions.

Resolves #1016.

### DIFF
--- a/modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
@@ -156,6 +156,9 @@ case class Artifact(
     ).flatten.mkString("\n")
   }
 
+  def scalaCliInstall: String =
+    s"//> using lib $groupId::$artifactId:$version"
+
   def csLaunch: String =
     s"cs launch $groupId:$artifactId:$version"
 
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -156,6 +156,9 @@ case class Artifact(
     ).flatten.mkString("\n")
   }
 
+  def scalaCliInstall: String =
+    s"//> using lib $groupId::$artifactId:$version"
```
julienrf, line 160:
```suggestion
    s"//> using lib $groupId$artifactOperator$artifactId:$version"
```

julienrf (approved):

PR #1060, opened 2022-08-18
Sort by commit activity

adpi2:
Add `Commit Activity` as a sorting criteria of the search page.

It replaces the `Forks` sorting criteria. I also removed the fork stat in the search result.


adpi2:
And the result is here: https://index-dev.scala-lang.org/search?q=*&sort=commit-activity

PR #1061, opened 2022-08-18
Improve usability of commit activity chart

adpi2:
- Group commits by week
- Use hit radius of 10 px

@joke1196, I allowed myself to tweak the activity chart a bit, is it okay?

## Before
activity-before

## After
activity-after


joke1196:
Yes of course! It looks really nice!

PR #1072, opened 2022-09-10
Fix badges page

takezoe:
In Badges page, markdown examples are wrong (`)` is missing after the image url).

Also, "Copy Markdown" buttons for platforms copy the latest version example.


adpi2 (approved):
Thanks a lot!

PR #1077, opened 2022-10-07
resolves #1032

### LINKED ISSUE #1032: Ability to add projects with no published artifacts to Maven Central
### Motivation

Currently only the projects that publishes to Maven Central are going to be indexed into Scaladex. This is okay because Maven Central is predominant in the Scala ecosystem.

However we expect some library authors to experiment with other package repositories like [Github Packages](https://github.com/features/packages), or maybe they don't publish any artifacts at all.

Before we integrate Scaladex with other package repositories, it would be good to be able to manually insert a Github repository, without any known artifacts.

### Identified Obstacles

Scaladex should be able to load the project pages of a project with no artifacts.

### Implementation Guideline

- [x] #1077 In the `AdminService` add a method that takes a `Project.Reference` and that start a task that insert it into the database. As an example of what a task is you can see the `AdminService#runMissingArtifactsTask`.
- [ ]  In the `AdminPages` add a POST endpoint that expects a `Project.Reference` and that call the `AdminService`.
- [ ]  Make sure that we can load the project page of a project with no artifact. Currently it will fail in `ProjectPages` because it expects every project to have a latest version. So `WebDatabase#getLatestVersion` should return an option. Maybe other methods should be adapted in the same way.
- [ ]  Optionnaly, in the `AdminService` method, we should check that the project is valid by calling the `GithubClient.getProjectInfo` on it: the response should be ok, and the percentage of Scala code should be greater than 0.

### Expectations

An admin can insert a project from the admin page: https://index.scala-lang.org/admin. (The admin page is only accessible to Scaladex admins but anyone can try it locally on https://localhost:8080/admin)

mfirry:


mfirry:
@adpi2 would you find the time to have a look here?

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
```diff
@@ -74,6 +74,43 @@ class AdminService(
 
   def allTaskStatuses: Seq[Task.Status] = tasks.map(_.status)
 
+  def addProjectNoArtifact(reference: Project.Reference, user: UserState
+                          ): Unit = {
+
+    def buildSettings(): Settings = Settings(
+      false, None, false, None, List.empty, false,false, Set.empty, Set.empty, None, None)
```
adpi2, line 81: This is `Setttings.default`

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
```diff
@@ -74,6 +74,43 @@ class AdminService(
...
+
+    def buildProject(info: GithubInfo): Project = new Project(
+      organization = Organization(""), // FIXME
+      repository = Repository(""), // FIXME
+      creationDate = None, // FIXME
```
adpi2, line 86: This comes from `reference`

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
```diff
@@ -74,6 +74,43 @@ class AdminService(
...
 
+  def addProjectNoArtifact(reference: Project.Reference, user: UserState
+                          ): Unit = {
+
+    def buildSettings(): Settings = Settings(
+      false, None, false, None, List.empty, false,false, Set.empty, Set.empty, None, None)
+
+    def buildProject(info: GithubInfo): Project = new Project(
+      organization = Organization(""), // FIXME
+      repository = Repository(""), // FIXME
```
adpi2, line 85: This comes from `reference`

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
adpi2, line 86: This field is not relevant so you can leave it with `None`

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
```diff
@@ -74,6 +74,43 @@ class AdminService(
...
+      creationDate = None, // FIXME
+      GithubStatus.Ok(java.time.Instant.now),
+      githubInfo = Some(info),
+      settings = buildSettings()
+    )
+
+    val input = Seq("Organization" -> reference.organization.value, "Repository" -> reference.repository.value)
+
+    val task = TaskRunner.run(Task.missingProjectNoArtifact, user.info.login, input) { () =>
+      githubClientOpt.fold(throw new Exception("Failed because 1 ??")) { githubClient =>
```
adpi2, line 95: Can you find a better error message here and also in line `98` and `100`.

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
```diff
@@ -74,6 +74,43 @@ class AdminService(
...
+        githubClient.getProjectInfo(reference).flatMap {
+          case GithubResponse.Ok((_, info)) =>
```
adpi2, line 97: You should also handle other possible `GithubResponse` cases, possibly with an error message.

modules/server/src/main/scala/scaladex/server/service/AdminService.scala
mfirry: Of course :) I actually wanted to ask you how you handle exceptions in the project. I'll look around other parts of the code and see.

mfirry:
One test is failing. I did not add any though.
`getCommitActivity *** FAILED ***`

adpi2:
Thanks for the contribution @mfirry!

This is the first step of #1032. The next steps are, to add the corresponding endpoint in `ProjectPages` and then to make sure that a project page with no artifacts can be loaded.

PR #1078, opened 2022-10-10
add quotes to scala-cli artifacts

bishabosha:
using directives syntax is to use scala strings for arguments

I didn't check if any tests need updating


julienrf:
I’ve hit this issue too, thank you for submitting a fix!

adpi2 (approved):
Thanks @bishabosha

PR #1079, opened 2022-10-14
resolves #1032 part 2

### LINKED ISSUE #1032: Ability to add projects with no published artifacts to Maven Central

mfirry:
Second task. new `post` added.

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/route/AdminPage.scala
+++ b/modules/server/src/main/scala/scaladex/server/route/AdminPage.scala
@@ -5,6 +5,7 @@ import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
 import scaladex.core.model.Artifact
 import scaladex.core.model.Env
+import scaladex.core.model.Project
 import scaladex.core.model.UserState
 import scaladex.server.TwirlSupport._
 import scaladex.server.service.AdminService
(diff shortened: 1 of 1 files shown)


adpi2:
@mfirry That's great!

May I ask you to also add the corresponding line in [this table](https://github.com/scalacenter/scaladex/blob/7ac05383f9b16e4eebec2839af1676b02b3751b7/modules/template/src/main/twirl/scaladex/view/admin/admin.scala.html#L66-L84), so that we can run the new task from the `Admin` page?

Currently it contains only on line to add a missing artifact manually.
image

adpi2 (approved):
Thanks again for your contribution!

PR #1136, opened 2022-12-10
Switch to endpoints4s::fetch-client

julienrf:
The fetch-client interpreter is more actively maintained than the xhr-client interpreter.


adpi2 (approved):
Thanks

PR #1153, opened 2023-01-30
Reverse sort stats by Scala/SJS/SN/SBT/Mill version on frontpage

### LINKED ISSUE #1151: Show Scala versions in descending order
instead of "2.10 2.11 2.12 2.13 3.x", IMO it should show as "3.x 2.13 2.12 2.11 2.10", on the general design principle of putting the more important things first

(suggested on Discord by @keynmol)

keynmol:
Closes #1151 

Front page:

image

Search results: 
image

VS the current ordering on scaladex: 
image

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/route/FrontPage.scala
+++ b/modules/server/src/main/scala/scaladex/server/route/FrontPage.scala
@@ -56,11 +56,11 @@ class FrontPage(env: Env, database: WebDatabase, searchEngine: SearchEngine)(imp
         "Typelevel" -> "typelevel"
       )
 
-      val scalaVersions = languages.collect { case (v: Scala, c) => (v, c) }
-      val scalaJsVersions = platforms.collect { case (v: ScalaJs, c) => (v, c) }
-      val scalaNativeVersions = platforms.collect { case (v: ScalaNative, c) => (v, c) }
-      val sbtVersions = platforms.collect { case (v: SbtPlugin, c) => (v, c) }
-      val millVersions = platforms.collect { case (v: MillPlugin, c) => (v, c) }
+      val scalaVersions = languages.collect { case (v: Scala, c) => (v, c) }.sortBy(_._1.version).reverse
+      val scalaJsVersions = platforms.collect { case (v: ScalaJs, c) => (v, c) }.sortBy(_._1.version).reverse
+      val scalaNativeVersions = platforms.collect { case (v: ScalaNative, c) => (v, c) }.sortBy(_._1.version).reverse
+      val sbtVersions = platforms.collect { case (v: SbtPlugin, c) => (v, c) }.sortBy(_._1.version).reverse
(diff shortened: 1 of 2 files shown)


julienrf (changes requested):
Thank you @keynmol for helping with this!

I think what you propose works, but I wonder if we should fix the issue upstream, by changing the definition of the `Ordering[SemanticVersion]` to always show higher versions first (which is the natural ordering for versions). That will probably have impacts in a couple of other places, though… Would you be interested in trying that approach?

keynmol:
> always show higher versions first (which is the natural ordering for versions).

Hmm I'm not sure I fully agree whether it's "natural", i.e. would you expect `1.2.0 > 3.2.0`? 

Alternatively, I can define a `LatestFirst` explicit ordering, and move the ordering out of the templates and into the code that calls them.

I've been interpreting ordering as numeric, and from the semver docs: 
image

obviously they `<` is an overloaded operator :)

julienrf:
OK, you are right that what I said would be confusing with `>`… Let’s keep the `Ordering[SemanticVersion]` as it is.

modules/server/src/main/scala/scaladex/server/route/FrontPage.scala
```diff
@@ -56,11 +56,11 @@ class FrontPage(env: Env, database: WebDatabase, searchEngine: SearchEngine)(imp
...
-      val scalaVersions = languages.collect { case (v: Scala, c) => (v, c) }
-      val scalaJsVersions = platforms.collect { case (v: ScalaJs, c) => (v, c) }
-      val scalaNativeVersions = platforms.collect { case (v: ScalaNative, c) => (v, c) }
-      val sbtVersions = platforms.collect { case (v: SbtPlugin, c) => (v, c) }
-      val millVersions = platforms.collect { case (v: MillPlugin, c) => (v, c) }
+      val scalaVersions = languages.collect { case (v: Scala, c) => (v, c) }.sortBy(_._1.version).reverse
+      val scalaJsVersions = platforms.collect { case (v: ScalaJs, c) => (v, c) }.sortBy(_._1.version).reverse
+      val scalaNativeVersions = platforms.collect { case (v: ScalaNative, c) => (v, c) }.sortBy(_._1.version).reverse
+      val sbtVersions = platforms.collect { case (v: SbtPlugin, c) => (v, c) }.sortBy(_._1.version).reverse
+      val millVersions = platforms.collect { case (v: MillPlugin, c) => (v, c) }.sortBy(_._1.version).reverse
```
julienrf, line 63: It seems `languages` and `platforms` are created on lines 38 and 39, from the result of calls to `SearchEngine#countByLanguages()` and `SearchEngine#countByPlatforms()`, and those methods return the result in the order that we want but we reverse them at lines 38 and 39. Should we just drop the `.map(_.sorted)` from those lines?

julienrf (approved):
Thank you @keynmol for your help! I agree with your analysis.

julienrf:
This is now live on https://index.scala-lang.org!

PR #1163, opened 2023-02-14
Improve landing page - topic cloud, SJS/SN/scala highlights

keynmol:
Commence the bikeshedding!

Main changes:

- Denoise the tag cloud, by removing parentheses and moving library count into superscripts
- Highlight current versions of Scala, SJS, SN, SBT/Mill - older versions are relegated into much smaller space, to not create the impression that all are equally prominent

Before: 
image

After (note that local data doesn't have mill plugins, so it's not showing up)
image


modules/server/src/main/scala/scaladex/server/route/FrontPage.scala
```diff
@@ -69,18 +61,146 @@ class FrontPage(env: Env, database: WebDatabase, searchEngine: SearchEngine)(imp
...
+              version = sbt.version,
+              deprecated = sbt.isDeprecated,
+              libraryCount = cnt,
+              search = Url(s"search?platforms=${sbt.label}")
+            )
+        }
+        selectCurrent(allVersions).map {
+          case (currentVersion, otherVersions) =>
+            EcosystemHighlight(
+              ecosystem = "SBT",
```
sjrd, line 163:
```suggestion
              ecosystem = "sbt",
```
"sbt" is a proper noun that is always spelled all-lowercase. Even at the start of sentences.

modules/server/src/main/scala/scaladex/server/route/FrontPage.scala
```diff
@@ -69,18 +61,146 @@ class FrontPage(env: Env, database: WebDatabase, searchEngine: SearchEngine)(imp
...
+              deprecated = sjs.isDeprecated,
+              libraryCount = cnt,
+              search = Url(s"search?platforms=${sjs.label}")
+            )
+        }
+        selectCurrent(allVersions).map {
+          case (currentVersion, otherVersions) =>
+            EcosystemHighlight(
+              "Scala.js",
+              Some("Robust Scala-to-JavaScript compiler for browser and node.js."),
```
sjrd, line 84: I'm not sure if these taglines for Scala.js and Scala Native are helping in this context. Removing them would declutter a bit more, IMO.

sjrd (commented):
Looks nice. Two small suggestions.

keynmol:
> but I feel like the logos take too much space and overall the content is misaligned

I agree about the misalignment, let me try to flex a bit more.

With regards to logos, I could make them smaller - in particular I think if the Scala logos will be same size (or actually smaller) than the scalajs/native ones, it will look cleaner.

Will do another pass, and if it's a no go - will remove them completely.

bishabosha:
(from the screenshot) the projects count for Scala Native looks like it doesn't go to the next line, but does for Scala.js, is this just because the scala.js description is too long?

keynmol:
Logos were too fiddly to get right, they all have different sizes and different empty space around them.

- Removed the logos and descriptions
- greyed out versions that are truly deprecated
- Centered everything 
- Projects and versions are separate paragraphs
image

julienrf:
Thank you again for your work! The screenshot looks good to me!

Do you mind posting screenshots for various sizes of screens (especially tablet and mobile)?

keynmol:
@julienrf I took a few screenshots - seems fine at first sight.

On really narrow things we can add a media query and just flex the entire list into a single column, but imo there's not many screens around with a resolution that low to care :)

Screenshot 2023-02-14 at 16 24 08

Screenshot 2023-02-14 at 16 23 56

Screenshot 2023-02-14 at 16 23 46

keynmol:
@julienrf anything else you would want to change here?

I still think that even this (toned down) version is an improvement, but I'm happy to adjust if you have any more ideas :)

modules/template/src/main/twirl/scaladex/view/topicsCloud.scala.html
```diff
@@ -0,0 +1,10 @@
+@import scaladex.core.model._
+
+@(
+  topics: Seq[TopicCount]
+)
```
adpi2, line 5: As for `ecosystemHighlights`, I would define this component in `frontpage.scala.html`.

modules/server/src/main/assets/css/partials/_main.scss
```diff
@@ -126,6 +126,71 @@ main {
...
     }
+    .ecosystem-heading {
+      width: 100%;
+      text-align: center;
+      font-weight: bold;
+      color: white;
+      margin: 5px;
+    }
+    .ecosystem-badges {
+      display: flex;
```
adpi2, line 137: I would rather use Bootstrap grid system than flexbox. Bootstrap is built internally using flexbox, but it is easier to work with and maintain. Especially in this case where we have a fixed number of boxes on each row.

build.sbt
```diff
@@ -105,6 +105,14 @@ lazy val infra = project
         s"-Dscaladex.elasticsearch.port=$elasticsearchPort"
       )
     },
+    reStart / javaOptions ++= {
```
adpi2, line 108: Do we still need `reStart`? Why not use `bgRun` instead?

I wonder why `reStart / javaOptions` does not contain `Compile / run / javaOptions` by default. I think this should work:
```scala
reStart / javaOptions := (Compile / run / javaOptions).value
```

adpi2 (commented):
Thanks a lot @keynmol for this great proposal and for implementing it. It is very much welcome. Scaladex deserves a better front page and unfortunately we did not find much time to work on it.

I left a few comments about the code style. Here are a few more comments about the design itself:
- I think we can remove the deprecated information, to improve the readability. Because currently it is hard to understand that the style of `2.12` is different because `2.12` is not deprecated.
image
- I would expect the "181 projects" to be clickable as part of the "Scala 2.13" box, because clicking on "Scala 2.13" brings to the list of projects.
- I don't like the "Build tool plugins" title floating alone in the middle of the page. It would be more consistent if we also put titles above the Scala versions and the platform versions. But I would not mind putting no titles at all. Anywhere else on Scaladex, sbt and Mill are considered platforms, in the same way as Scala.js and Scala Native.

Those comments are open to discussion. Also if you don't want to implement the changes now, I would be okay to merge and improve later.

adpi2:
For testing, I deployed this PR to the dev environment: https://index-dev.scala-lang.org/

adpi2:
I am taking this PR over to push it to the finish line.

PR #1171, opened 2023-03-02
update scala-cli dependency using directive

laurin-no:
with scala-cli v0.2.0 the dependency using directive is renamed to `dep`. see release notes: https://github.com/VirtusLab/scala-cli/releases/tag/v0.2.0


adpi2:
Thanks for the update

adpi2 (approved):

PR #1304, opened 2023-11-21
Issue 1268

### LINKED ISSUE #1268: [Bug] image in readme would crash when default branch is not master
### Current behavior

image in readme would crash when default branch is not master

see project logo in [here](https://index.scala-lang.org/scalacenter/bloop)

### Expected Behavior

image in readme should display

### Extra comments

Should tune `Client.scala` rather hard code `master`

### Search terms

`master` or `blob` in `Client.scala`

mfirry:
Fixes #1268

### DIFF
--- a/modules/webclient/src/main/scala/scaladex/client/Client.scala
+++ b/modules/webclient/src/main/scala/scaladex/client/Client.scala
@@ -74,21 +74,21 @@ object Client {
         .map(t => headers + ("Authorization" -> s"bearer $t"))
         .getOrElse(headers)
 
-    val readmeRequest = new Request(
+    val readmeRequest: Request = new Request(
       s"https://api.github.com/repos/$organization/$repository/readme",
       new RequestInit {
         headers = headersWithCreds.toJSDictionary
       }
     )
 
-    val repoRequest = new Request(
+    val repoRequest: Request = new Request(
       s"https://api.github.com/repos/$organization/$repository",
       new RequestInit {
         headers = headersWithCreds.toJSDictionary
       }
     )
 
-    def setReadme() = fetch(readmeRequest).toFuture
+    def setReadme(): Future[Unit] = fetch(readmeRequest).toFuture
       .flatMap { res =>
         if (res.status == 200) {
           res.text().toFuture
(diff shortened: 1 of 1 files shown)


modules/webclient/src/main/scala/scaladex/client/Client.scala
```diff
@@ -36,27 +42,30 @@ object Client {
...
+
+      fetch(readmeRequest).toFuture
+        .flatMap { res =>
+          if (res.status == 200) {
+            res.text().toFuture
+          } else {
+            Future.successful("No README found for this project, please check the repository")
+          }
+        }
+        .flatMap(res => Future { element.innerHTML = res })
```
adpi2, line 61:
```suggestion
        .map(res => element.innerHTML = res)
```

modules/webclient/src/main/scala/scaladex/client/Client.scala
```diff
@@ -84,6 +93,33 @@ object Client {
...
+          headers = headersWithCreds.toJSDictionary
+        }
+      )
+      fetch(repoRequest).toFuture
+        .flatMap { res =>
+          if (res.status == 200) {
+            res.text().toFuture
+          } else {
+            Future.successful("{\"default_branch\": \"master\"}")
+          }
```
adpi2, line 109: `
For the fallback, I would avoid creating a JSON string if we then need to parse it and extract one field.

```suggestion
          def extractDefaultBranch(text: String): String =
            js.JSON.parse(text).asInstanceOf[Repo].default_branch
          if (res.status == 200) res.text().map(extractDefaultBranch).toFuture
          else Future.successful("master")
```

modules/webclient/src/main/scala/scaladex/client/Client.scala
```diff
@@ -36,27 +42,30 @@ object Client {
...
+            res.text().toFuture
+          } else {
+            Future.successful("No README found for this project, please check the repository")
+          }
+        }
+        .flatMap(res => Future { element.innerHTML = res })
+    }
+
+    def setLogo(): Future[Unit] = {
+      def getLogoAndSetHTML(branch: String, organization: String, repository: String): Unit = {
```
adpi2, line 65: It does not just set the logo, it fixes all the URLs of all the images in the readme file.

```suggestion
    def getDefaultBranchAndFixImages(): Future[Unit] = {
      def fixImages(branch: String, organization: String, repository: String): Unit = {
```

adpi2 (commented):
Here are a few suggestions but overall it looks good.

For the commit message I would appreciate if you could squash all commits into a single one.

PR #1305, opened 2023-11-22
Fixes #1268

### LINKED ISSUE #1268: [Bug] image in readme would crash when default branch is not master

mfirry:


modules/webclient/src/main/scala/scaladex/client/Client.scala
```diff
@@ -84,6 +96,31 @@ object Client {
...
+        .flatMap { res =>
+          if (res.status == 200) {
+            res.text().toFuture.map(extractDefaultBranch)
+          } else { Future.successful("master") }
+        }
+        .map { res =>
+          val resJson = js.JSON.parse(res)
+          val branch = resJson.asInstanceOf[Repo].default_branch
+          fixImages(branch, organization, repository)
+        }
```
adpi2, line 116: I deployed it to https://index-dev.scala-lang.org/scalacenter/bloop but it does not work.

It seems there is a problem here because it tries again to parse the default_branch to JSON.

```suggestion
        .map(branch => fixImages(branch, organization, repository))
```

mfirry: Hmmm weird. I tried it locally.

mfirry: Ah no. My bad. Let me see.

PR #1330, opened 2024-01-11
Fix latest-by-scala-version badge summary to prefer releases

rtyley:
I just noticed that the versions shown in the Scaladex badges for [`content-api-firehose-client`](https://index.scala-lang.org/guardian/content-api-firehose-client/badges) don't match - one is showing the stable release number (`1.0.12` 👍), and the other is showing a pre-release (less good):

* The 'Latest version' badge is showing a good version: `1.0.12` 
image

* The 'JVM badge': `1.0.13-PREVIEW.rt-dbremove-travis-file.2024-01-10T1738.6e259255` 
image

image

The JVM badge shows a latest-by-scala-version summary I introduced with https://github.com/scalacenter/scaladex/pull/660, but it looks like I missed out logic which is present in the other badge - preferring to show releases, and only showing pre-releases if they are the _only_ available version!

This PR fixes the behaviour on the latest-by-scala-version badge to match the behaviour of the other badge - it does this with a new `PreferReleases` ordering for `SemanticVersion`s. I did have a look to see if I could refactor the logic used by the first badge to also make use of this ordering, but it would have required a bigger refactoring:

https://github.com/scalacenter/scaladex/blob/9c07f4f13e1053d8f4f3179d6f338f566be8bf22/modules/server/src/main/scala/scaladex/server/route/Badges.scala#L159-L162

### DIFF
--- a/modules/server/src/test/scala/scaladex/server/route/BadgesTests.scala
+++ b/modules/server/src/test/scala/scaladex/server/route/BadgesTests.scala
@@ -9,9 +9,12 @@ import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.model.headers.Location
 import akka.http.scaladsl.server.Route
 import org.scalatest.BeforeAndAfterAll
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
 import scaladex.core.model.Scala._
 import scaladex.core.test.Values._
 import scaladex.core.util.ScalaExtensions._
+import scaladex.server.route.Badges.summaryOfLatestVersions
 
 class BadgesTests extends ControllerBaseSuite with BeforeAndAfterAll {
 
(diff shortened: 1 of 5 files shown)


modules/server/src/test/scala/scaladex/server/route/BadgesTests.scala
```diff
@@ -22,17 +25,6 @@ class BadgesTests extends ControllerBaseSuite with BeforeAndAfterAll {
...
-    Badges.summaryOfLatestVersions(
-      Map(
-        `2.11` -> Seq(`7.0.0`, `7.1.0`),
-        `2.12` -> Seq(`7.0.0`, `7.1.0`, `7.2.0`),
-        `2.13` -> Seq(`7.0.0`, `7.1.0`, `7.2.0`, `7.3.0`),
-        `3` -> Seq(`7.2.0`, `7.3.0`)
-      )
-    ) shouldBe "7.3.0 (Scala 3.x, 2.13), 7.2.0 (Scala 2.12), 7.1.0 (Scala 2.11)"
-  }
-
```
rtyley, line 35: To take advantage of this being a simple unit test, which can be run easily without external infrastructure like docker, I've moved it further down the file (along with a couple of other new unit tests) into `BadgesUnitTests` - this doesn't extend `ControllerBaseSuite` and so can be quickly run in IntelliJ, etc, without having to start docker.

adpi2 (approved):
Thanks for the contribution!

rtyley:
Thanks! I can confirm that the badges on this sample project now show the correct release version:

https://index.scala-lang.org/rtyley/sample-project-using-gha-scala-library-release-workflow/badges

image


...even though there are 'higher' number preview releases:

https://index.scala-lang.org/rtyley/sample-project-using-gha-scala-library-release-workflow/artifacts/core?pre-releases=true#

image

PR #1374, opened 2024-04-02
Add Scastie Box Into Artifact Page

ayushkoli772:
With this commit I have fixed the Scastie URL according to Try library, inside `Artifact.scala`

I will completely fix this issue with subsequent commits

@adpi2 Please review the changes to `scastieUrl` so I can move forward with adding it to `artifact.scala.html`


modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -221,29 +221,65 @@ case class Artifact(
...
+  def isSbtPlugin: Boolean = platform match {
+    case SbtPlugin(_) => true
+    case _            => false
+  }
 
-    val targetParam = binaryVersion.platform match {
-      case ScalaJs(_) => Some("t" -> "JS")
-      case _          => None
-    }
+  def getOrganization: String = projectRef.organization.toString
```
adpi2, line 253: Small suggestion: I don't think we need those getters, instead you can inline `projectRef.organization.toString`.

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -221,29 +221,65 @@ case class Artifact(
...
+        case Scala(v) => Some("sv" -> v.toString)
+        case _        => None
+      }
+
+      val params: List[(String, String)] = List(
+        "g" -> groupId.value,
+        "a" -> artifactName.value,
+        "v" -> version.toString,
+        "o" -> getOrganization, // get organization value
+        "r" -> getRepository // get repository value
```
adpi2, line 278: I am curious about the Scastie endpoint. How did you find out that we need those params?

adpi2 (commented):
So far it looks good. I tried to build the same URL manually and it worked.

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
ayushkoli772: I found it here- https://github.com/scalacenter/scastie/#try-library

In that it said that `o -` scaladex organization and `r -` scaladex repository fields are `required` fields

ayushkoli772: I will replace the getter setters with inline `projectRef.organization.toString` in next commit

I had another doubt, the html code of Scastie box is given and also it is mentioned that we have to add this in `artifact.scala.html` 

But it is not mentioned where actually Scastie box was previously situated inside `artifact.scala.html`

Can you please guide me in this?

adpi2: I don't remember how this Scastie box was.

I would try to add it here to see how it looks. Maybe it is better to nest it in another white box or maybe not.

image

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -221,29 +221,65 @@ case class Artifact(
...
-    }
+  def getRepository: String = projectRef.repository.toString
 
-    val params: List[(String, String)] = List(
-      "g" -> groupId.value,
-      "a" -> artifactName.value,
-      "v" -> version.toString
-    ) ++ targetParam ++ scalaVersionParam
+  def scastieURL: Option[String] =
+    if (isSbtPlugin) {
```
adpi2, line 258: Also if it is Scala Native it should return `None`.

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
ayushkoli772: I will add it as you suggested and we can also experiment with it later to find where it fits best

Thank you for your guidance, I will commit with above changes implemented as soon as possible!

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
ayushkoli772: Done, it will be visible in next commit, here's how I changed it to also return None if Scala Native- 

`if (isSbtPlugin || binaryVersion.platform == ScalaNative(_)) {
      None
    }` 

I hope this fulfills the requirements

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
ayushkoli772: Hello @adpi2 

This how it looks after adding Scastie box-

image

And this is url you get when clicked on `Try online with scastie box`- https://scastie.scala-lang.org/try?g=com.typesafe.akka&a=akka-actor&v=2.6.19&o=akka&r=akka&t=JVM&sv=3.x

Does this meets your expectations?

ayushkoli772:
Hello @adpi2, with this commit I fixed the Scastie URL and also added Scastie box

This how it looks after adding Scastie box-

image

And this is url you get when clicked on Try online with scastie box- https://scastie.scala-lang.org/try?g=org.typelevel&a=cats-core&v=2.8.0&o=typelevel&r=cats&t=JVM&sv=3.x

Also it adds the library in library list like this-

image

Are there any other changes you would like to suggest?

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -221,28 +221,59 @@ case class Artifact(
...
+
+      val targetParam = binaryVersion.platform match {
+        case ScalaJs(_) => Some("t" -> "JS")
+        case _          => Some("t" -> "JVM")
+      }
+
+      val scalaVersionParam = binaryVersion.language match {
+        case Scala(v) => Some("sv" -> v.toString)
+        case _        => None
+      }
```
adpi2, line 266: I tried the Scastie URL with a missing `sv` param and it did not work. The `sv` param is required.

So if we don't have a Scala version, we cannot build a Scastie URL, and the method should return `None`.

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -221,28 +221,59 @@ case class Artifact(
...
+  // }
+
+  // def isSbtPlugin: Boolean = platform match {
+  //   case SbtPlugin(_) => true
+  //   case _            => false
+  // }
+
+  def scastieURL: Option[String] = binaryVersion.platform match {
+    case SbtPlugin(_) | ScalaNative(_) => None
+    case _ =>
```
adpi2, line 255: Using pattern matching is okay, it is clear and it does what we need.

But I would like to propose an alternative, using a for-comprehension on options:

```scala
val targetParamOpt = binaryVersion.platform match {
  case ScalaJs(_) => Some("t" -> "JS")
  case _          => Some("t" -> "JVM")
}

val scalaVersionParamOpt = binaryVersion.language match {
  case Scala(v) => Some("sv" -> v.toString)
  case _        => None
}

// combine the optinal target parameter and optional scala version parameter to build the URL
for {
  targetParam <- targetParamOpt
  scalaVersionParam <- scalaVersionParamOpt
}  yield {
  val params: List[(String, String)] = List(
    "g" -> groupId.value,
    "a" -> artifactName.value,
    "v" -> version.toString,
    "o" -> projectRef.organization.toString,
    "r" -> projectRef.repository.toString,
    targetParam,
    scalaVersionParam  
  )
  params.map { case (k, v) => s"$k=$v" }.mkString(tryBaseUrl + "?", "&", "")
}
```

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
```diff
@@ -221,28 +221,59 @@ case class Artifact(
...
+
+  //   params
+  //     .map { case (k, v) => s"$k=$v" }
+  //     .mkString(tryBaseUrl + "?", "&", "")
+  // }
+
+  // def isSbtPlugin: Boolean = platform match {
+  //   case SbtPlugin(_) => true
+  //   case _            => false
+  // }
```
adpi2, line 251: Please remove the unnecessary comments.

modules/template/src/main/twirl/scaladex/view/project/artifact.scala.html
```diff
@@ -97,6 +98,17 @@ <h4>Documentation</h4>
...
+  @if(artifact.scastieURL.nonEmpty) {
+    <div class="box">
+      <a href="@artifact.scastieURL" target="_blank" rel="nofollow noopener">
+        <h6>Try online with Scastie</h6>
+        <img style="height: 100px;" src="/assets/img/scastie.svg" alt="Try on Scastie">
+      </a>
+    </div>
+  }
+}
+
```
adpi2, line 111: Looks good to me! :+1:

adpi2:
Also could you add some unit tests on this `scastieURL` method?

modules/core/shared/src/main/scala/scaladex/core/model/Artifact.scala
ayushkoli772: @adpi2 What do you think about this?

I have used for-comprehension and added the check whether `SbtPlugin` or `ScalaNative` inside targetParam

```
def scastieURL: Option[String] = {
      val tryBaseUrl = "https://scastie.scala-lang.org/try"

      val targetParam = binaryVersion.platform match {
        case ScalaJs(_) => Some("t" -> "JS")
        case SbtPlugin(_) | ScalaNative(_) => return None
        case _ => Some("t" -> "JVM")
      }

      val scalaVersionParam = binaryVersion.language match {
        case Scala(v) => Some("sv" -> v.toString)
        case _        => return None
      }

      for {
        target <- targetParam
        scalaVersion <- scalaVersionParam
      } yield {
        val params: List[(String, String)] = List(
          "g" -> groupId.value,
          "a" -> artifactName.value,
          "v" -> version.toString,
          "o" -> projectRef.organization.toString,
          "r" -> projectRef.repository.toString,
          target,
          scalaVersion
        )
        params.map { case (k, v) => s"$k=$v" }.mkString(tryBaseUrl + "?", "&", "")
      }
  }
```

adpi2: About this:
```scala
val targetParam = binaryVersion.platform match {
    case ScalaJs(_) => Some("t" -> "JS")
    case SbtPlugin(_) | ScalaNative(_) => return None
    case _ => Some("t" -> "JVM")
  }
```
I have two comments:
1. In Scala we try not to use any return. Here you don't need any because the for-comprehension will return None if any of `targetParam` or `scalaVersionParam` is None.
2. The platform param should be "JVM" only if the platform is `JVM`. The default case, which handles `SbtPlugin`, `ScalaNative` and `MillPlugin`, should be `None`.

ayushkoli772: I accidentally used `return`. I apologize for that

Here is updated code for `scastieURL` method-

```
def scastieURL: Option[String] = {
      val tryBaseUrl = "https://scastie.scala-lang.org/try"

      val targetParam = binaryVersion.platform match {
        case ScalaJs(_) => Some("t" -> "JS")
        case JVM => Some("t" -> "JVM")
        case _ => None
      }

      val scalaVersionParam = binaryVersion.language match {
        case Scala(v) => Some("sv" -> v.toString)
        case _        => None
      }

      for {
        target <- targetParam
        scalaVersion <- scalaVersionParam
      } yield {
        val params: List[(String, String)] = List(
          "g" -> groupId.value,
          "a" -> artifactName.value,
          "v" -> version.toString,
          "o" -> projectRef.organization.toString,
          "r" -> projectRef.repository.toString,
          target,
          scalaVersion
        )
        params.map { case (k, v) => s"$k=$v" }.mkString(tryBaseUrl + "?", "&", "")
      }
  }
```

ayushkoli772:
I have added unit tests for `scastieURL` method

This is output after `compile` and `test`-

image

I also wanted to add another unit test, in which it should return `None` if `ScalaVersion` is not given. But the problem is Im unable to pass `ScalaVersion` as `None` while creating an artifact instance so I tried passing `JVM` like below-

```
it("should return None if artifact doesn't have ScalaVersion") {
      val artifact = createArtifact(
        groupId = "org.typelevel",
        artifactId = "cats-effect_3",
        version = "2.6.1",
        binaryVersion = BinaryVersion(Jvm, JVM)
      )
      artifact.scastieURL should be(None)
    }
```

But I think the problem is, `createArtifact` method doesnt takes the language only from what we passed in `BinaryVersion()` while creating instance but instead from `artifactId` like below-

```
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
    Artifact(
      groupId = GroupId(groupId),
      artifactId = artifactId,
      version = SemanticVersion.parse(version).get,
      artifactName = artifactIdResult.name,
      platform = artifactIdResult.binaryVersion.platform,
      language = artifactIdResult.binaryVersion.language,
      projectRef = projectRef.getOrElse(Project.Reference.from("", "")),
      description = None,
      releaseDate = Instant.now(),
      resolver = resolver,
      licenses = Set(),
      isNonStandardLib = false
    )
```

@adpi2 Can you please guide me in this? Thank you

ayushkoli772:
I also wanted to add another unit test, in which it should return `None` if `ScalaVersion` is not given. But the problem is Im unable to pass `ScalaVersion` as `None` while creating an artifact instance so I tried passing `JVM` like below-

```
it("should return None if artifact doesn't have ScalaVersion") {
      val artifact = createArtifact(
        groupId = "org.typelevel",
        artifactId = "cats-effect_3",
        version = "2.6.1",
        binaryVersion = BinaryVersion(Jvm, JVM)
      )
      artifact.scastieURL should be(None)
    }
```

But I think the problem is, `createArtifact` method doesnt takes the language only from what we passed in `BinaryVersion()` while creating instance but instead from `artifactId` like below-

```
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
    Artifact(
      groupId = GroupId(groupId),
      artifactId = artifactId,
      version = SemanticVersion.parse(version).get,
      artifactName = artifactIdResult.name,
      platform = artifactIdResult.binaryVersion.platform,
      language = artifactIdResult.binaryVersion.language,
      projectRef = projectRef.getOrElse(Project.Reference.from("", "")),
      description = None,
      releaseDate = Instant.now(),
      resolver = resolver,
      licenses = Set(),
      isNonStandardLib = false
    )
```

@adpi2 Can you please guide me in this? Thank you

ayushkoli772:
I have added unit tests for `scastieURL` method

This is output after `compile` and `test`-

image


I wanted to add another unit test, in which it should return `None` if `ScalaVersion` is not given. But the problem is Im unable to pass `ScalaVersion` as `None` while creating an artifact instance so I tried passing `JVM` like below-

```
it("should return None if artifact doesn't have ScalaVersion") {
      val artifact = createArtifact(
        groupId = "org.typelevel",
        artifactId = "cats-effect_3",
        version = "2.6.1",
        binaryVersion = BinaryVersion(Jvm, JVM)
      )
      artifact.scastieURL should be(None)
    }
```

But I think the problem is, `createArtifact` method doesnt takes the language only from what we passed in `BinaryVersion()` while creating instance but instead from `artifactId` like below-

```
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
    Artifact(
      groupId = GroupId(groupId),
      artifactId = artifactId,
      version = SemanticVersion.parse(version).get,
      artifactName = artifactIdResult.name,
      platform = artifactIdResult.binaryVersion.platform,
      language = artifactIdResult.binaryVersion.language,
      projectRef = projectRef.getOrElse(Project.Reference.from("", "")),
      description = None,
      releaseDate = Instant.now(),
      resolver = resolver,
      licenses = Set(),
      isNonStandardLib = false
    )
```

@adpi2 Can you please guide me in this? Thank you

ayushkoli772:
I have added unit tests for `scastieURL` method

This is output after `compile` and `test`-

image


I wanted to add another unit test, in which it should return `None` if `ScalaVersion` is not given. But the problem is Im unable to pass `ScalaVersion` as `None` while creating an artifact instance so I tried passing `JVM` like below-

```
it("should return None if artifact doesn't have ScalaVersion") {
      val artifact = createArtifact(
        groupId = "org.typelevel",
        artifactId = "cats-effect_3",
        version = "2.6.1",
        binaryVersion = BinaryVersion(Jvm, JVM)
      )
      artifact.scastieURL should be(None)
    }
```

But I think the problem is, `createArtifact` method doesnt takes the language only from what we passed in `BinaryVersion()` while creating instance but instead from `artifactId` like below-

```
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
    Artifact(
      groupId = GroupId(groupId),
      artifactId = artifactId,
      version = SemanticVersion.parse(version).get,
      artifactName = artifactIdResult.name,
      platform = artifactIdResult.binaryVersion.platform,
      language = artifactIdResult.binaryVersion.language,
      projectRef = projectRef.getOrElse(Project.Reference.from("", "")),
      description = None,
      releaseDate = Instant.now(),
      resolver = resolver,
      licenses = Set(),
      isNonStandardLib = false
    )
```

@adpi2 Can you please guide me in this? Thank you

ayushkoli772:
I have added unit tests for `scastieURL` method

This is output after `compile` and `test`-

image


I wanted to add another unit test, in which it should return `None` if `ScalaVersion` is not given. But the problem is Im unable to pass `ScalaVersion` as `None` while creating an artifact instance so I tried passing `JVM` like below-

```
it("should return None if artifact doesn't have ScalaVersion") {
      val artifact = createArtifact(
        groupId = "org.typelevel",
        artifactId = "cats-effect_3",
        version = "2.6.1",
        binaryVersion = BinaryVersion(Jvm, JVM)
      )
      artifact.scastieURL should be(None)
    }
```

But I think the problem is, `createArtifact` method doesnt takes the language only from what we passed in `BinaryVersion()` while creating instance but instead from `artifactId` like below-

```
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
    Artifact(
      groupId = GroupId(groupId),
      artifactId = artifactId,
      version = SemanticVersion.parse(version).get,
      artifactName = artifactIdResult.name,
      platform = artifactIdResult.binaryVersion.platform,
      language = artifactIdResult.binaryVersion.language,
      projectRef = projectRef.getOrElse(Project.Reference.from("", "")),
      description = None,
      releaseDate = Instant.now(),
      resolver = resolver,
      licenses = Set(),
      isNonStandardLib = false
    )
```

@adpi2 Can you please guide me in this? Thank you

ayushkoli772:
I have added unit tests for `scastieURL` method

This is output after `compile` and `test`-

image


I wanted to add another unit test, in which it should return `None` if `ScalaVersion` is not given. But the problem is Im unable to pass `ScalaVersion` as `None` while creating an artifact instance so I tried passing `JVM` like below-

```
it("should return None if artifact doesn't have ScalaVersion") {
      val artifact = createArtifact(
        groupId = "org.typelevel",
        artifactId = "cats-effect_3",
        version = "2.6.1",
        binaryVersion = BinaryVersion(Jvm, JVM)
      )
      artifact.scastieURL should be(None)
    }
```

But I think the problem is, `createArtifact` method doesnt takes the language only from what we passed in `BinaryVersion()` while creating instance but instead from `artifactId` like below-

```
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
    Artifact(
      groupId = GroupId(groupId),
      artifactId = artifactId,
      version = SemanticVersion.parse(version).get,
      artifactName = artifactIdResult.name,
      platform = artifactIdResult.binaryVersion.platform,
      language = artifactIdResult.binaryVersion.language,
      projectRef = projectRef.getOrElse(Project.Reference.from("", "")),
      description = None,
      releaseDate = Instant.now(),
      resolver = resolver,
      licenses = Set(),
      isNonStandardLib = false
    )
```

@adpi2 Can you please guide me in this? Thank you

ayushkoli772:
With this commit, I have added unit tests for `scastieURL` method

This is terminal output after `compile` and `test`-

Screenshot from 2024-04-14 07-47-21

@adpi2 Please review. Thank you

modules/core/shared/src/test/scala/scaladex/core/model/ArtifactTests.scala
```diff
@@ -141,13 +141,70 @@ class ArtifactTests extends AnyFunSpec with Matchers {
...
+    it("should return a valid URL for a ScalaJs artifact") {
+      createArtifact(
+        groupId = "org.scala-js",
+        artifactId = "scalajs-library_2.13",
+        version = "1.16.0",
+        binaryVersion = BinaryVersion(ScalaJs.`1.x`, Scala.`3`),
+        projectRef = Some(Project.Reference.from("Scala.js", "scalajs-library"))
+      ).scastieURL should be(
+        Some(
+          "https://scastie.scala-lang.org/try?g=org.scala-js&a=scalajs-library&v=1.16.0&o=scala.js&r=scalajs-library&t=JVM&sv=2.13"
```
adpi2, line 194: This `t=JVM` looks wrong. If it is a Scala.Js library the target param should be `JS`. The problem is not in the implementation but in the `createArtifact` method:

Let's look at this line of code:
```scala
val artifactIdResult =
      artifactName.map(name => ArtifactId(name, binaryVersion)).orElse(Artifact.ArtifactId.parse(artifactId)).get
```

It can translate to: if there is an `artifactName` we combine it with `binaryVersion` to create the `ArtifactId` otherwise we parse the `String` in the `artifactId` argument. Whenever `artifactName` is `None` the `binaryVersion` is just ignored. In this test the `artifactName` is `None` so the given `binaryVersion` (`BinaryVersion(ScalaJs.1.x, Scala.3)`) is never used. Instead the `createArtifact` get the `binaryVersion` from parsing the `artifactId`. The `artifactId` is `scalajs-library_2.13` which means the binary version is `BinaryVersion(JVM, Scala.2.13)`. For a Scala.js artifact, the `artifactId` should be something like `scalajs-library_sjs1_2.13`.

Remark however that the real `scalajs-library_sjs1_2.13` artifact does not exist because `scalajs-library` is not a Scala.js library, but a JVM library. It's quite surprising. Instead you can use the `scalajs_dom_sjs1_2.13` artifact ([link](https://index.scala-lang.org/scala-js/scala-js-dom/artifacts/scalajs-dom/2.8.0?binary-version=_sjs1_2.13))

modules/core/shared/src/test/scala/scaladex/core/model/ArtifactTests.scala
```diff
@@ -141,13 +141,70 @@ class ArtifactTests extends AnyFunSpec with Matchers {
...
+        binaryVersion = BinaryVersion(SbtPlugin.`0.13`, Scala.`2.10`)
+      ).scastieURL should be(None)
+    }
+
+    it("should return None if artifact is ScalaNative") {
+      createArtifact(
+        groupId = "org.typelevel",
+        artifactId = "cats-core",
+        version = "2.6.1",
+        binaryVersion = BinaryVersion(ScalaNative.`0.4`, Scala.`2.13`)
```
adpi2, line 160: Here as well, this is inconsistent. If the binary version is `BinaryVersion(ScalaNative.`0.4`, Scala.`2.13`)` then the `artifactId` should be `cats-core_native0.4_2.13`

modules/core/shared/src/test/scala/scaladex/core/model/ArtifactTests.scala
```diff
@@ -141,13 +141,70 @@ class ArtifactTests extends AnyFunSpec with Matchers {
...
+      ).scastieURL should be(None)
+    }
+
+    it("should return None if artifact doesn't have ScalaVersion") {
+      createArtifact(
+        groupId = "org.typelevel",
+        artifactId = "",
```
adpi2, line 167: Here it would be better to use a real Java artifact. For instance this one: https://index.scala-lang.org/lightbend/config/artifacts/config/1.4.3.

```suggestion
        groupId = "com.typesafe",
        artifactId = "config",
```

adpi2 (changes requested):

modules/core/shared/src/test/scala/scaladex/core/model/ArtifactTests.scala
ayushkoli772: I thought `scalajs-library` is a Scala.js library and built URL according to that. Thank you for clarification

ayushkoli772:
Done. Please review. 

Thank you for your time

adpi2 (approved):

PR #1382, opened 2024-04-18
Update README To Remove Gitter Channel

ayushkoli772:
Gitter is not used anymore so replace it with Github discussions

@adpi2 Please review. Thank you


adpi2 (approved):
Looks great! Thank you!

Issue #1024, opened 2022-04-29
Add a `fullScalaVersion` field in `Artifact`


adpi2:
### Motivation

Sometimes we want to know which version of the compiler are used by the library authors. We can find this information in the dependencies of an artifact. But it would more convenient to have it directly from the `Artifact` entity.

### Implementation Guideline

- Add a `fullScalaVersion: Option[SemanticVersion]` in the `Artifact` class.
- Add a `full_scala_version` in the `artifacts` table by adding a SQL migration file [here](https://github.com/scalacenter/scaladex/tree/main/modules/infra/src/main/resources/migrations) and adapt the SQL requests in `ArtifactsTable`.
- In `ArtifactConverter` find the full Scala version in the artifact dependencies. Get the version from the `scala3-library` dependency or the `scala-library`. Beware that Scala 3 libraries depend on both `scala3-library` and `scala-library`. 
- Display the full Scala version to the user in the `artifacts.scala.html` template.

### Expectations

The full Scala version of newly published artifacts should appear in the artifact page.

WojciechMazur:
In the context of the Scala 3 LTS vs Scala 3 Next and SIP-51 (drop binary compatibility of Scala 2.13 standard library) it might be useful to present information about major Scala version used to build the project. Some libraries do require Scala 3 Next features and are not released using Scala LTS. Similarly, once some of Scala 2.13 libraries might use backward incompatible features once the SIP-51 introduces them. Presenting information about using Scala major version used for publishing would inform users before-hand and would allow to decide if we want to use/upgrade to newer version library without upgrading used version of Scala. 
We might use `scala3-library_3`and `scala-library_2.13` versions from POM file to gather required information. Then, we can present them in some more readable way. Based on that we might even introduce filters for searching. 

Why introduce special handling when we already have a list of dependencies? 
Let's take a look at first library showing in the Scaladex homepage to me: https://index.scala-lang.org/playframework/playframework/artifacts/play/2.9.3 - becouse of multiple dependencies the version of Scala is hard to find without using browser search tool. It could help if Scala version would be presented at the top. Even better if it would be presented under the `group id`, `version`
<img width="594" alt="image" src="https://github.com/scalacenter/scaladex/assets/19353690/605db2c6-01ce-440b-8c4c-ebe3a615caea">


[...]

PR #1406, opened 2024-05-26
scaladex: 1024- Add a fullScalaVersion field in Artifact (#1)

kannupriyakalra:
Issue- https://github.com/scalacenter/scaladex/issues/1024

- added fullScalaVersion in Artifact class
- added extractScalaVersion method
- added "full_scala_version" field in ArtifactTable
- added data migration file V20__add_full_scala_version.sql
- displayed the full Scala version to the user in the artifacts.scala.html template.
- updated tests

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/service/ArtifactConverter.scala
+++ b/modules/server/src/main/scala/scaladex/server/service/ArtifactConverter.scala
@@ -62,12 +62,7 @@ class ArtifactConverter(paths: DataPaths) extends LazyLogging {
       (dep.artifactId == "scala-library" || dep.artifactId == "scala3-library_3")
     }
 
-    val sortedScalaDependencies = scalaDependencies.sortBy(_.artifactId)
-    val fullScalaVersion = scalaDependencies match {
-      case _ :: _ :: _ => Some(sortedScalaDependencies.last.version)
-      case head :: Nil => Some(head.version)
-      case _           => None
-    }
+    val fullScalaVersion = scalaDependencies.sortBy(_.artifactId).lastOption.map(_.version)
 
     fullScalaVersion.flatMap(SemanticVersion.parse)
   }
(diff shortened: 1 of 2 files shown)


modules/server/src/main/scala/scaladex/server/service/ArtifactConverter.scala
```diff
@@ -63,6 +56,22 @@ class ArtifactConverter(paths: DataPaths) extends LazyLogging {
...
+      dep.groupId == "org.scala-lang" &&
+      (dep.artifactId == "scala-library" || dep.artifactId == "scala3-library_3")
+    }
+
+    val sortedScalaDependencies = scalaDependencies.sortBy(_.artifactId)
+    val fullScalaVersion = scalaDependencies match {
+      case _ :: _ :: _ => Some(sortedScalaDependencies.last.version)
+      case head :: Nil => Some(head.version)
+      case _           => None
+    }
```
adpi2, line 70: I think this is the same as:
```suggestion
    val fullScalaVersion = scalaDependencies.sortBy(_.artifactId).lastOption.map(_.version)
```

modules/template/src/main/twirl/scaladex/view/project/artifact.scala.html
```diff
@@ -62,6 +62,7 @@ <h3>
               @if(artifact.resolver.isEmpty) {
                 @info("Files") { <a href="@artifact.mavenReference.repoUrl">View all</a> }
               }
+              @artifact.fullScalaVersion.map{ version => @info("Scala version") { @version.toString }}
```
adpi2, line 65: I think we need to be more specific:

```suggestion
              @artifact.fullScalaVersion.map{ version => @info("Full Scala version") { @version.toString }}
```

adpi2 (approved):
Some suggestions. Apart from that it is looking quite good.

adpi2:
Thanks @kannupriyakalra. I deployed this to https://index.scala-lang.org/. Let's wait a bit for new artifacts to come and we should see the `fullScalaVersion`.

mzuehlke:
First artifact appeared: 

https://index.scala-lang.org/nationalarchives/da-metadata-schema/artifacts/da-metadata-schema/0.0.19?binary-version=_3

PR #1438, opened 2024-08-14
Pre-compute the latest version of each artifact

### LINKED ISSUE #1429: [Bug] Scaladex site is returning HTTP 503  responses
### Current behavior

Responses on https://index.scala-lang.org/ are currently eventually timing out with an HTTP 503 (Service Unavailable) error:

```
The server was not able to produce a timely response to your request.
Please try again in a short while!
```

image

### Expected Behavior

HTTP 200!

### Extra comments

_No response_

### Search terms

_No response_

adpi2:
Fixes #1429 

When loading the project page, we need to compute the latest version of each of its artifacts. To do so, we used to sending an SQL JOIN request of the `artifacts` table on itself. This request was too frequent and too expensive, and it caused the recent downtimes of Scaladex.

Here we pre-compute the `is_latest_version` of each artifact: each time a new artifact is added, we get all the artifacts with the same ID, we compute the max version, and flag it with `is_latest_version`. Now the SQL request to get the latest version of each artifact of a project is a simple SELECT.


modules/infra/src/main/scala/scaladex/infra/sql/ArtifactTable.scala
```diff
@@ -114,26 +118,16 @@ object ArtifactTable {
...
+  def unsetOthersLatestVersion: Update[Artifact.MavenReference] =
+    updateRequest0(
       table,
-      Seq("group_id", "artifact_id", "MAX(release_date) as release_date"),
-      where = Seq("release_date IS NOT NULL") ++ isReleaseFilters ++ Seq("organization=?", "repository=?"),
-      groupBy = Seq("group_id", "artifact_id")
+      set = Seq("is_latest_version=false"),
+      where = Seq("group_id=?", "artifact_id=?", "version<>?")
     )
-  }
```
rtyley, line 138: Just checking - it looks like the old code would check `"is_semantic='true'", "is_prerelease='false'"` if `stableOnly` was true - ie there could be more than one 'latest' for a set of artifacts, depending on whether you only wanted to see stable artifacts or not. Is that logic removed now? I think personally I'd prefer that the 'latest' be restricted to the latest _stable_ version?

I'm not sure where this value is surfaced but I would guess it's here, near the top of the [project page](https://index.scala-lang.org/scanamo/scanamo) by the repo name: 

scaladex

...would that `2.0.0` now become the very long pre-release version number (`3.0.0-PREVIEW.add-transact-to-expose-TransactionCanceledException.2024-08-10T0658.6eeb6522`)?

adpi2: There is only one set of latest artifacts at a time, but it can be stable or maybe not stable depending on the project settings. The `stableOnly ` setting is used when we compute the `is_latest_version`, and if the project setting changes we re-compute it.

Also I already deployed this PR, so what you see on the website is the current PR. You can try to log in, go to the project settings, disable the `Stable Only` filter and you should see the new latest version becomes the very long pre-release version.

rtyley: Thanks @adpi2 that makes sense! The only thing is that, even though I'm logged in on Scaladex, I don't think I've ever been able to access my projects settings - where is that functionality?

Here are a couple of projects that I think I should be able to administer:

image
image

If I click on `My projects` from the dropdown, I only see this:

image

I don't actually _want_ to be able to disable the `Stable only` flag, I'm just curious how I would do it!?

adpi2: Ah yes, that's because it takes a very long time to get the list of projects, and organizations of a user using the Github API, and so we do it asynchronously. Usually it takes about 30 seconds. But yesterday I had the impression that it was taking forever, or that it was failing. Maybe come back to your project in a few minutes.

Also if you log out it will reset the timer, so after logging in again you'll have to wait again.

adpi2: @rtyley Are you able to access you project settings? If not could you open a new issue.

rtyley: > Are you able to access you project settings?

Only on repos which I directly own, not organisation-held ones! I've opened https://github.com/scalacenter/scaladex/issues/1445 to track that.

adpi2: Thanks!

Issue #1407, opened 2024-05-27
Add background task to reindex all artifacts


adpi2:
### Motivation

Each time we update the `Artifact` model we need to re-download all POM files to get the new values and populate the new fields in the database.

### Identified Obstacles

- We need a new endpoint to trigger the task. (The difference between a scheduled job and a task is: a scheduled job runs automatically, a task is triggered by an admin once in a while).
- We should make sure that downloading all POM files will not fill the disk completely. (Probably it won't because POM files are quite small, smaller than JAR files).
- What if we cannot download some artifact: should we remove it from Scaladex?

### Implementation Guideline

- In `Task`, add a new task `update-maven-artifacts`
- In `AdminPage`, add a new endpoint to trigger this task. 
- In `AdminService` create a method `updateMavenArtifacts` that will download each POM file and reindex it. Maybe create a new class File `MavenArtifactUpdater` to implement this method. It should
  - Query the database to get all Maven artifact references.
  - Use `CoursierResolver` to download the POM file
  - Use `PublishProcess` to publish the POM file.

### Expectations

We should be able to trigger the task in the admin page and it should update all artifacts one by one without failing. If there is a failure it should appear in the admin page.

Issue #979, opened 2022-03-29
Extract versionScheme, scaladocUrl, developers from pom files 


adpi2:
### Motivation

The artifact page contains useful information about the pom file of an artifact.

Currently we have some useful information:
- organization and name
- project (Github repository)
- language (Java, Scala 2.12, Scala 2.13...)
- platform (JVM, Scala.js, Scala Native, sbt)
- licence
- (resolver)
- dependencies (and inverse dependencies)

See for instance https://index.scala-lang.org/typelevel/cats/artifacts/cats-core/2.9.0

But we can get more from the pom file:
- the scaladoc url:
```xml
<properties>
  <info.apiURL>http://endpoints4s.github.io/api</info.apiURL>
<properties>
```
- the version scheme
```xml
<properties>
  <info.versionScheme>early-semver</info.versionScheme>
</properties>
```
- the list of developers:
```xml
<developers>
  <developer>
    <id>julienrf</id>
    <name>Julien Richard-Foy</name>
    <url>http://julien.richard-foy.fr</url>
  </developer>
</developers>
```

As a good example of pom file you can see:
https://repo1.maven.org/maven2/org/endpoints4s/algebra_2.13/1.7.0/algebra_2.13-1.7.0.pom

### Identified Obstacles

Manage missing information with optional values.

### Implementation Guideline

1. Add new fields in `ArtifactModel`: `scaladocUrl` and `versionScheme` should be optional
2. Fill those fileds in the `PomConvert.apply` method
3. Add corresponding types and fields in the `Artifact` case class: for `scaladocUrl` we can use the already existing `Url` case class, for `versionScheme` we can use a `String`, and for `developers` we can introduce a new case class `Developer`.
4. Create a migration file in [this folder](https://github.com/scalacenter/scaladex/tree/main/modules/infra/src/main/resources/migrations) to add new columns in the `artifacts` table
5. Update the `ArtifactsTable` object to declare those field and read/write them. For the `developers` field we need a serializer. I recommend to define it using circe in the `Codecs` class.
6. Add some of information in the artifacts of the `Values` class to test that we can read and write them to the database without trouble. Those tests already exist in `SqlDatabaseTests`.
7. Add a scheduled job to find the missing information from the existing artifacts.

### Expectations

1. We should extract the information from poms that we receive from Sonatype.
2. Reading and writing those field into the database should be tested.

muskan3006:
Hi I would start looking into this one, is that okay?

[...]

PR #1446, opened 2024-08-20
1407 :  Background task to reindex all artifacts &  979 : UI Changes

skingle:
- https://github.com/scalacenter/scaladex/issues/1407
- https://github.com/scalacenter/scaladex/issues/979

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -83,7 +83,9 @@ lazy val infra = project
       "io.circe" %% "circe-core",
       "io.circe" %% "circe-generic",
       "io.circe" %% "circe-parser"
-    ).map(_ % V.circe),
+    ).map(_ % V.circe) ++ Seq(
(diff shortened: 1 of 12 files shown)


adpi2:
Great work @skingle!

adpi2:
Deployed to https://index-dev.scala-lang.org/

adpi2:
The task is still running after 18 hours, which seems normal. I checked a few projects but I don't see any new fields appearing in the artifact page. I'll run it locally to see if it's working as expected. Maybe there is something we can do to speed it up.

adpi2:
I can see in [this artifact](https://index-dev.scala-lang.org/com-lihaoyi/mill/artifacts/mill-bsp/0.11.10) that the re-indexing is working.

It takes about 150 ms to re-index a pom file. So it is going to take about 80 hours total.

PR #1483, opened 2024-10-17
Fix wrong ordering of platforms

### LINKED ISSUE #1450: [Bug] Wrong ordering of platform on Scala Native 0.5
### Current behavior

https://index.scala-lang.org/scalameta/munit-scalacheck/artifacts/munit-scalacheck?stable-only=true

image

### Expected Behavior

`Scala Native 0.5` should be on top of `Scala Native 0.4`

### Extra comments

_No response_

### Search terms

_No response_

ayushkoli772:
Changes to fix #1450 - 
Screenshot from 2024-10-18 01-15-39

@adpi2 Please review

### DIFF
--- a/modules/template/src/main/twirl/scaladex/view/project/artifactsFilters.scala.html
+++ b/modules/template/src/main/twirl/scaladex/view/project/artifactsFilters.scala.html
@@ -15,7 +15,7 @@
   <select class="selectpicker" name="binary-versions" 
     title= "@params.binaryVersionsSummary.getOrElse("Scala Versions")" multiple data-style="btn-secondary" data-actions-box="true" data-selected-text-format="static"
     onchange="this.form.submit()">
-    @for((platform, binaryVersions) <- allBinaryVersions.sorted.groupBy(_.platform)) {
+    @for((platform, binaryVersions) <- allBinaryVersions.sorted.groupBy(_.platform).toSeq.sortBy(_._1).reverse) {
       <optgroup label="@platform">
         @for(binaryVersion <- binaryVersions.reverse) {
             <option value="@binaryVersion.label"


modules/template/src/main/twirl/scaladex/view/project/artifactsFilters.scala.html
```diff
@@ -15,7 +15,7 @@
   <select class="selectpicker" name="binary-versions" 
     title= "@params.binaryVersionsSummary.getOrElse("Scala Versions")" multiple data-style="btn-secondary" data-actions-box="true" data-selected-text-format="static"
     onchange="this.form.submit()">
-    @for((platform, binaryVersions) <- allBinaryVersions.sorted.groupBy(_.platform)) {
+    @for((platform, binaryVersions) <- allBinaryVersions.sorted.groupBy(_.platform).toSeq.sortBy(_._1).reverse) {
```
adpi2, line 18: I don't think we need to sort it twice. The first `.sorted` call should be enough, although it may need to be reversed. Otherwise we can remove it and only keep the `sortBy(_._1).reverse`. But I think I would prefer to keep the first one.

adpi2 (commented):
Hello Ayush, I am glad to see you there again :)

ayushkoli772:
> Hello Ayush, I am glad to see you there again :)

Hello Adrien! Im glad to see you too :)

modules/template/src/main/twirl/scaladex/view/project/artifactsFilters.scala.html
ayushkoli772: Correct me if im wrong. 

First `.sorted` will sort the binary versions and then the `sortBy(_._1).reverse` will sort the grouped platforms. So output of this approach will be binary versions are in sorted order and the platforms will also be sorted that way Scala Native 0.5 will always appear before Scala Native 0.4.

Alternative to this approach will be sorting the grouped platforms based on the max version in each platform group and then sort the binary versions to show the latest first. It can be done by something like this- `allBinaryVersions.groupBy(_.platform).toSeq
      .sortBy { binaryVersions.map(_.version).max }`
Output of this approach will be binary versions are in sorted order and the platforms will be sorted based on which platform has latest/max binary version. So Scala Native 0.5 will be over Scala Native 0.4.

What do you think?

adpi2: > First .sorted will sort the binary versions and then the sortBy(_._1).reverse will sort the grouped platforms

The binary versions is composed of the platform and the language. So sorting the binary version should also sort the platforms and we should not need to sort them again. But then the `.groupBy`  creates a `Map` which does not preserve the order. So I think the solution here is to remove the first `sorted`:

```suggestion
    @for((platform, binaryVersions) <- allBinaryVersions.groupBy(_.platform).toSeq.sortBy(_._1).reverse) {
```

> Alternative to this approach will be sorting the grouped platforms based on the max version in each platform group and then sort the binary versions to show the latest first.

There already is an `Ordering[Platform]` defined as : https://github.com/scalacenter/scaladex/blob/a306d3cd9ea23a206b1ece15f18309f6b8483bfa/modules/core/shared/src/main/scala/scaladex/core/model/Platform.scala#L81-L87

We should better use it than redefining a custom one here.

adpi2:
Superseded by #1484

PR #1492, opened 2024-11-05
Separate developers by |

Philippus:
but don't leave a trailing |

I think this should fix the trailing | behind the last developer.

Scherm­afbeelding 2024-11-05 om 17 13 10


adpi2 (approved):
LGTM

PR #1494, opened 2024-11-09
Override title because of typo in case object

Philippus:
Ideally we would change the name of the case object and add a migration I think? But not completely sure how that works, any pointers would be appreciated.

### DIFF
--- a/modules/core/shared/src/main/scala/scaladex/core/model/Category.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/model/Category.scala
@@ -73,7 +73,9 @@ object Category {
   case object Compilers extends Category
   case object CodeGeneration extends Category
   case object DependencyInjection extends Category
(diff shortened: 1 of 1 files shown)


adpi2:
Thanks @Philippus 

> Ideally we would change the name of the case object and add a migration I think? But not completely sure how that works, any pointers would be appreciated. But not completely sure how that works, any pointers would be appreciated.

We use flyway to automate the migrations. You need to do add a SQL file in [this folder](https://github.com/scalacenter/scaladex/tree/main/modules/infra/src/main/resources/migrations) and flyway will take care of running it once. The name of the file should start with `V25__`, and it should contain an SQL query:
```sql
UPDATE project_settings SET category='functional-programming-and-category-theory' WHERE category='functionnal-programming-and-category-theory';
```

I am going to merge this PR but it would be great if you could push the proper rename and migration.

PR #1500, opened 2024-11-18
Rename FunctionnalProgrammingAndCategoryTheory and add migration

Philippus:
Follow-up of https://github.com/scalacenter/scaladex/pull/1494. It looks like migrations V25 and V26 are hidden in another directory.

### DIFF
--- a/modules/core/shared/src/main/scala/scaladex/core/model/Category.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/model/Category.scala
@@ -73,9 +73,7 @@ object Category {
   case object Compilers extends Category
   case object CodeGeneration extends Category
   case object DependencyInjection extends Category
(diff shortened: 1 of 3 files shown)


Philippus:
@adpi2 Any idea how to fix the failing test?

adpi2:
> It looks like migrations V25 and V26 are hidden in another directory.

Indeed, we can also write Scala migration files. Flyway can find them in the classpath and execute them.

> @adpi2 Any idea how to fix the failing test?

We have a sub-module `small-index` which contains a JSON snapshot of projects and artifacts for the integration tests. Some of those projects contain `"category" : "functionnal-programming-and-category-theory"` which needs to be renamed. I allowed myself to push the update to your branch.

adpi2 (approved):
Thanks for the quick fix

Philippus:
Ah yes, I didn't check the sub-modules.

PR #1519, opened 2025-01-03
Fix project default artifact

### LINKED ISSUE #1508: [Bug] Only Scala 2 version of a library presented, although Scala 3 is also available
### Current behavior

I maintain a [CSV library](https://github.com/fingo/spata/). It used to be correctly presented in Scaladex, as available for [Scala 3](https://central.sonatype.com/artifact/info.fingo/spata_3) and Scala 2.13, until I published a new version for [Scala 2 only](https://central.sonatype.com/artifact/info.fingo/spata_2.13). See it in [Scaladex](https://index.scala-lang.org/fingo/spata).

### Expected Behavior

Support for both Scala versions, 2.13 and 3.x, should be presented.

### Extra comments

The Scala 3 release is marked as the recent one in Github, but it doesn't influence Scaladex.
Scala 2 and Scala 3 versions are kept on different branches, which may be the source of the problem.

### Search terms

_No response_

adpi2:
Fix #1508 :
- Use latest date to get the default artifact name
- Use latest version the get the latest artifact with that name

### DIFF
--- a/modules/core/shared/src/main/scala/scaladex/core/model/ProjectHeader.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/model/ProjectHeader.scala
@@ -44,36 +44,49 @@ final case class ProjectHeader(
     s"/$ref/artifacts/${defaultArtifact.name}$queryParams"
   }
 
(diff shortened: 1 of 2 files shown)


adpi2:
image

PR #1651, opened 2025-12-01
fix(chart): use milliseconds for commit activity timestamps (fixes #1…

### LINKED ISSUE #1: API
JSON APIs for accessing Scaladex data
- [ ] Joris's needs

In the future: (to keep in mind)
- [ ] Felix will potentially need for linking dottydoc
- [ ] Adriaan for community build

arnavsharma990:
### Description
The "Commit Activity" chart on project pages was displaying dates starting from 1970 (Unix Epoch). This was caused by a unit mismatch: the server was providing a timestamp in **Seconds**, but the frontend client (and Chart.js) expected **Milliseconds**.

### Changes
* **`project.scala.html`**: Updated the template to explicitly convert the start date to **Milliseconds** using `.toEpochMilli` before rendering it to the HTML attribute.
* **`Sparkline.scala`**: Updated the Scala.js frontend to parse the `data-commit-activity-starting-day` attribute using `Instant.ofEpochMilli`.

### Verification
* Ran `./cs launch sbt -- compile` to ensure type safety for the new time conversions.
* Verified that the logic now consistently uses Unix Milliseconds on both server and client.

### DIFF
--- a/modules/server/src/test/scala/scaladex/server/service/ArtifactServiceTests.scala
+++ b/modules/server/src/test/scala/scaladex/server/service/ArtifactServiceTests.scala
@@ -2,6 +2,7 @@ package scaladex.server.service
 
 import scaladex.core.model.Version
 import scaladex.core.test.InMemoryDatabase
+
 import org.scalatest.funspec.AsyncFunSpec
 import org.scalatest.matchers.should.Matchers
 


kannupriyakalra:
@arnavsharma990 

It looks like the PR introduces (or surfaces) a regression in `OldSearchApiTests`. The failure comes from an `UnsupportedOperationException: empty.max` in `ArtifactService.computeLatestVersion`, specifically when calling `max` on an empty collection:

```
java.lang.UnsupportedOperationException: empty.max
  at scala.collection.IterableOnceOps.max
  ...
  at scaladex.server.service.ArtifactService.computeLatestVersion
```

This suggests that `versions` (or whatever sequence is being reduced) can be empty for some artifacts. Previously the test suite passed, so the new logic likely changed how versions are collected or filtered.

Could you:

1. Add a guard for the empty case in `computeLatestVersion` (e.g., `headOption`, `maxOption`, or an explicit fallback)?
2. Verify why `OldSearchApiTests` ends up with no versions, is this expected after your changes or an unintended side effect?
3. Consider adding a small test case that exercises this empty-versions scenario to prevent regressions.

Once that’s addressed, the rest of the suite passes, so this should unblock the PR.

arnavsharma990:
CI needed a tiny formatting tweak (import spacing). Fixed it now.

Issue #1609, opened 2025-09-26
[Bug] Incorrect list of supported Scala binary versions


WojciechMazur:
### Current behavior

https://index.scala-lang.org/eed3si9n/verify as a whole is not showing support for Scala 3

However we can find in artifacts that it actually has versions published for Scala 3 

<img width="756" height="766" alt="Image" src="https://github.com/user-attachments/assets/52045fce-4be0-46ba-9c99-d1612005f81f" /> 

This data is probably based on highest semantic version, not on the last/any version. 

### Expected Behavior

Show versions supported by library so it can be found during filtered search

arnavsharma990:
I have a fix ready for this!

I've updated the logic in ProjectHeader.scala to aggregate supported languages/platforms from the entire artifact history (instead of just the latest one).

Quick question: Since latestScalaVersions (and the related methods) now return all versions, the name is currently misleading.

Shall I rename these methods to allScalaVersions (and update the templates) as part of this PR? Or would you prefer I keep the existing names (latest...) to verify the fix first and minimize the diff?

PR #1652, opened 2025-12-01
fix(metadata): aggregate supported platforms from all artifacts and r…

arnavsharma990:
### Description
Fixes Issue #1609 where projects were listing incorrect supported Scala versions. Previously, the project header and search index only considered the **latest** released artifact. This meant that if a project supported Scala 3 in version `1.0.0` but released a maintenance patch `1.0.1` only for Scala 2.13, the Scala 3 support would "disappear" from the UI and search.

### Changes
* **Aggregation Logic:** Updated `ProjectHeader.scala` to aggregate supported languages and platforms from **all** artifacts in the project history, not just the latest.
* **Refactoring:** Renamed methods in `ProjectHeader` (e.g., `latestScalaVersions` -> `allScalaVersions`) to accurately reflect that they now return the complete history of supported versions.
* **Template Updates:** Updated `headproject.scala.html` to use the new method names.
* **Search Index:** Updated `ProjectDocument.scala` to index these aggregated fields so search results are accurate.

### Verification
* Ran `./cs launch sbt -- compile` to ensure all template references and model calls were updated to the new names.
* Verified the logic collects distinct versions from the entire artifact list.


arnavsharma990:
hey @WojciechMazur can you review it please when you have time

SethTisue:
afaik Scaladex is now maintainerless, since Adrien left the Scala Center and there isn't currently funding to replace him

I looked at the Sovereign Tech projects just now and I don't see one that we can plausibly shoehorn minor Scaladex maintenance into, either.

So maybe we'll just need to merge this and cross our fingers?

Issue #1576, opened 2025-04-30
[Bug] Information about ScalaFX shows very old version, incorrect artifact


jpsacha:
### Current behavior

Scaladex shows ScalaFX version as 8.0.0-M2 using Scala 2.10. This relates to a very old version of ScalaFX. 
The ReadMe.md pulled from the project has correct version from Maven Central info: 24.0.0-R35.
The this in turn impacts Scala Community Build not recognizing ScalaFX as a Scala 3 project.

Looks that the problem is with Scaladex picking up wrong artifact. Rather than using "scalafx" it is using "scalafx-demos" that is not published at ScalaFX artifact. The "artifacts" tab listing many/most items that are not part of ScalaFX project

### Expected Behavior

Point to the "scalafx" and correctly list last release version, currently 24.0.0-R35, and supported Scala versions: 2.12, 2.13, 3.3.

aaryan-iiitu:
I’ve opened a PR that fixes this by making artifact name matching and alphabetical ordering case insensitive  and added a regression test.  Happy to make any changes if needed

jpsacha:
This is still a problem. Scaladex still shows ScalaFX version as 8.0.0-M2 and wrong artifacts when for newer versions: instead "scalafx" it shows "scalafx-demos" or "scalafx-extras". They are released under the same organization but are not the correct artifacts.

tgodzik:
Should be fixed now

jpsacha:
@tgodzik Thanks. It works now.

PR #1662, opened 2026-01-05
Fix artifact selection when repository name casing differs

aaryan-iiitu:
### Context

While looking into issue #1576, I noticed that Scaladex can end up selecting an outdated artifact when the repository name and the artifact name differ only by casing (for example, `ScalaFX` vs `scalafx`). This happens because artifact selection currently relies on case sensitive string comparisons and case sensitive alphabetical ordering.

### What this PR changes

- Compare repository names and artifact names in a case-insensitive way
- Make alphabetical ordering of artifact names case insensitive as well
- Add a regression test to ensure this behavior doesn’t break again

### Why this approach

The existing selection rules and priority order are kept exactly the same. This change only normalizes string comparisons so that casing differences don’t affect which artifact is selected.

### Verification

- Added a test that reproduces the issue
- Ran the full `coreJVM/test` suite locally; all tests pass


jpsacha:
Getting this fixed will be very helpful

Issue #1645, opened 2025-11-19
[Bug] Maven central hosted project doesn't appear on scaladex


mghmay:
I made a small project here: https://github.com/mghmay/play-json-shaper

I made sure to add scm information in the pom and the artifact name follows naming conventions: https://central.sonatype.com/artifact/io.github.mghmay/play-json-shaper_2.13

I have a scaladex account linked to my github account, nothing appears there.

### Expected behavior

I would like my maven central hosted library to appear on scaladex.

### Current Behavior

It doesn't appear.

Algebrazebra:
I have the same issue for my project ([slick-duckdb](https://github.com/Algebrazebra/slick-duckdb)) not showing up in Scaladex despite being listed on [Maven Central](https://central.sonatype.com/artifact/io.github.algebrazebra/slick-duckdb_2.13/overview) with the following SCM in the POM:

```xml
<scm>
        <url>https://github.com/algebrazebra/slick-duckdb</url>
        <connection>scm:git:git@github.com:algebrazebra/slick-duckdb.git</connection>
</scm>
```

In the meantime I have opened a [PR to manually add my project](https://github.com/scalacenter/scaladex-contrib/pull/68), waiting for review/approval.

tgodzik:
Fixed the following:
https://index.scala-lang.org/mghmay/play-json-shaper 
https://index.scala-lang.org/spamegg1/tarski
https://index.scala-lang.org/algebrazebra/slick-duckdb

It seems everything is due to sonatype changes, there is a separate issue in https://github.com/scalacenter/scaladex/issues/1778

You can comment on the issue above or directly to me to add your repo.

PR #1688, opened 2026-03-30
Add a parser test for #1645

Gedochao:
It looks to me like the assumption from #1650 is wrong, as the parser seems to handle the example from #1645 correctly.
(I know #1650 fixes nothing, as it currently just renames APIs)
Adding a test as a sanity check.


tgodzik (approved):
LGTM

PR #1720, opened 2026-07-29
improvement: Try to fix an issue with performance

tgodzik:


lbialy-iterable (approved):
lgtm with added paging and delays, should help with the traffic overload

lbialy (approved):
lgtm with added paging and delays, should help with the traffic overload

PR #1730, opened 2026-08-12
feat: make Elasticsearch REST client pool and timeouts configurable

tbekas:
Raise the ES client connection pool above the default (10/route) and set bounded connect/socket/lease timeouts so concurrent requests neither starve for a connection nor hang indefinitely. All values are externalized to reference.conf under scaladex.elasticsearch.


tgodzik (approved):
Let's try it out!

PR #1731, opened 2026-08-12
feat: add Gatling load-test harness

tbekas:
Adds a `loadtest` sbt module with Gatling simulations for the read-heavy web-UI endpoints (realistic browsing mix + ramp-to-failure stress, isolated for Elasticsearch and Postgres) to correlate with the OpenTelemetry metrics.

Feeders are generated from small-index by the `data` module (`data/run generateFeeders`, reusing FilesystemStorage; only Ok-status projects, to avoid redirect/404 noise) and committed. Simulations run on demand via `loadtest/Gatling/test`; the module is aggregated so it compiles with the build but never runs on `sbt test`.


build.sbt
```diff
@@ -240,4 +255,6 @@ lazy val V = new {
   val circe = "0.14.12"
   val json4s = "4.1.0"
   val coursier = "2.1.24"
+  val otel = "1.64.0"
```
tgodzik: Seems unused

tbekas: Good catch 👍 Removing it

tgodzik (approved):

PR #1734, opened 2026-08-14
fix: encode redirect URIs to avoid crash on artifact names with special chars

tbekas:
## Problem

Legacy artifact redirect routes built the redirect target by interpolating the already-decoded path segments into a string, which was then re-parsed as a `Uri`. When an artifact name contains a character that needs percent-encoding (e.g. a space), parsing fails and the request returns a **500** instead of redirecting.

Example that failed in production:

```
GET /linkedin/talks/Feathr%20Feature%20Store%20Talk.pdf

org.apache.pekko.http.scaladsl.model.IllegalUriException: Illegal URI reference:
  Invalid input ' ', expected pchar, '/', '?', '#' or 'EOI'
  /linkedin/talks/artifacts/Feathr Feature Store Talk.pdf
```

The `Segment` matcher decodes `%20` → space, then `redirect(s"/$projectRef/artifacts/$artifactName")` re-parses the raw string (with the space) as a `Uri` and throws.

## Fix

Build the redirect target with pekko's `Uri.Path` / `Uri.Query` builders, which percent-encode each segment/param automatically. The settings redirect is hardened the same way for consistency.

Added a regression test covering an artifact name with a space.

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/route/ProjectPages.scala
+++ b/modules/server/src/main/scala/scaladex/server/route/ProjectPages.scala
@@ -204,11 +204,12 @@ class ProjectPages(
               _ <- artifactService.updateLatestVersions(projectRef, form.preferStableVersion)
               _ <- searchSynchronizer.syncProject(projectRef)
             yield ()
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM

PR #1760, opened 2026-08-17
Cache hot project/artifact reads with configurable Scaffeine caches

tbekas:
## What

Adds short-lived in-memory (Scaffeine) caches on the hottest database read paths in `SqlDatabase`, to cut Postgres/connection-pool pressure from bursts of repeated reads (e.g. crawlers enumerating every version of a project's artifacts).

## Details

- **Cached reads:** `getProject`, `getProjectArtifacts` (by name and by version), `getDirectDependencies`, `getReverseDependencies`, and `getProjectLatestArtifacts`. `getProject` runs on every project/artifact page, so caching it collapses a flood of per-version requests for one project into a single query per TTL window.
- **Invalidation:** the `getProject` cache is invalidated on every project-metadata write (`insertProjectRef`, `updateProjectSettings`, `updateGithubStatus`, `updateProjectCreationDate`, and the composite `updateGithubInfoAndStatus` / `moveProject`), so settings edits, GitHub updates and moves are reflected immediately. The artifact/dependency caches are TTL-only.
- **Configurable:** a new `CacheConfig` (`ttl`, `max-size`) is loaded from a `scaladex.caching` config block and threaded explicitly through `ServerConfig` / `IndexConfig` into `SqlDatabase`. `max-size` bounds memory (and bounds 404-flooding, which caches `None`).
- A small `buildCache` helper removes duplication across the cache definitions.

## Config

```hocon
scaladex {
  caching {
    ttl = 10 minutes
    max-size = 10000
  }
}
```

## Testing

`SqlDatabaseTests` pass (including the invalidation-sensitive cases: settings updates, GitHub status, creation date, moved project). All affected modules compile.

### DIFF
--- a/modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
+++ b/modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
@@ -9,6 +9,7 @@ import scala.concurrent.duration.*
 
 import scaladex.core.model.*
 import scaladex.core.service.SchedulerDatabase
+import scaladex.infra.config.CacheConfig
 import scaladex.infra.sql.*
 
 import cats.effect.IO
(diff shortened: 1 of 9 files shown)


modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
```diff
@@ -18,8 +19,46 @@ import com.typesafe.scalalogging.LazyLogging
...
+    }
+
+  private val projectArtifactsByVersionCache
+      : AsyncLoadingCache[(Project.Reference, Artifact.Name, Version), Seq[Artifact]] =
+    buildCache {
+      case (ref, name, version) =>
+        run(ArtifactTable.selectArtifactByProjectAndNameAndVersion.to[Seq](ref, name, version))
+    }
+
+  private val projectLatestArtifactsCache: AsyncLoadingCache[Project.Reference, Seq[Artifact]] =
```
tgodzik, line 50: I wonder if this one is too useful, but I guess this is only moved.

tgodzik (approved):
Looks good, let's try it!

PR #1762, opened 2026-08-17
Make the database connection pool size configurable

tbekas:
## What

Makes the Postgres (HikariCP) connection pool size configurable, and sets an explicit default of **25** (previously it relied on HikariCP's implicit default of 10).

## Details

- Add `scaladex.database.pool-size` (default `25`) to `PostgreSQLConfig`.
- Apply it in `DoobieUtils.getHikariDataSource` via `config.setMaximumPoolSize(conf.poolSize)`.

Override per environment via `scaladex.database.pool-size` in `application.conf` (or a `-D` system property).

## Note

The server opens two datasources (web + scheduler), so the process uses up to `2 × pool-size` connections — make sure Postgres `max_connections` accommodates that.

### DIFF
--- a/modules/infra/src/main/resources/reference.conf
+++ b/modules/infra/src/main/resources/reference.conf
@@ -16,6 +16,7 @@ scaladex {
     name = scaladex
     name = ${?POSTGRESQL_DATABASE}
     port = 5432
+    pool-size = 25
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
Let's try it, though I think we are already using the max of the processor time.

PR #1764, opened 2026-08-18
fix: return empty search results on malformed query instead of crashing

tbekas:
A search query with Lucene reserved characters (e.g. a leading `:`) fails to parse in Elasticsearch. `findPage` called `.result` on the failed response, throwing and returning a 500:

```
ERROR scaladex.server.Server$ - Unhandled exception in http://index.scala-lang.org/GET?q=%3Ahttps%3A%2F%2Ftinyurl.com%2Ftxy3b83h
java.util.NoSuchElementException: Request Failure ElasticError(search_phase_execution_exception,all shards failed,...
    at com.sksamuel.elastic4s.RequestFailure.result(responses.scala:71)
    at scaladex.infra.ElasticsearchEngine.findPage$$anonfun$1(ElasticsearchEngine.scala:203)
```

Now `findPage` handles a failed response by logging and returning an empty page. Added a regression test.


tgodzik (approved):
LGTM

PR #1765, opened 2026-08-18
perf: cache getProjectArtifactRefs reads

tbekas:
`getProjectArtifactRefs` hit Postgres on every call, so crawler bursts on large projects (e.g. `zio-aws`) piled heavy queries onto the connection pool — enough to exhaust it (`/api/project` timing out on `getConnection`).

Cache all three overloads (by project, by name, by version) using the existing scaffeine `buildCache` helper, and invalidate the affected entries on artifact writes:

- `insertArtifact` / `insertArtifacts` — invalidate the artifact's project/name/version keys.
- `updateArtifacts` (project moves) — invalidate all ref caches.


tgodzik (approved):
LGTM

PR #1766, opened 2026-08-18
feat: express Elasticsearch timeouts as durations

tbekas:
The Elasticsearch connect/socket/connection-request timeouts were integer milliseconds. Change them to `FiniteDuration` so they can be written as e.g. `connect-timeout = 5 seconds` in the config; the values are converted back to millis when configuring the REST client.

Note: the config keys were renamed (dropped the `-ms` suffix), so any environment overriding them needs updating.


tgodzik (approved):
LGTM

PR #1768, opened 2026-08-18
fix: don't 500 on malformed search queries (centralize ES failure handling)

tbekas:
Same class of bug as #1764, which fixed `findPage` — but the facet aggregations (`countByLanguages`/`Platforms`/`Topics`), `autocomplete`, and `scroll` still called `.result` on the ES response directly, so a malformed query 500'd there:

```
ERROR scaladex.server.Server$ - Unhandled exception in .../GET?q=(venv) PS D:\SAAS-ERP2> git config --get-regexp "credential|remote"
java.util.NoSuchElementException: Request Failure ElasticError(search_phase_execution_exception, ... parse_exception, Cannot parse ...)
    at com.sksamuel.elastic4s.RequestFailure.result(responses.scala:71)
    at scaladex.infra.ElasticsearchEngine.aggregation$$anonfun$1(ElasticsearchEngine.scala:332)
```

Centralize the handling in an `executeOrEmpty` helper used by all read paths (and guard `scroll`'s recursion): on a failed request, log a warning and return empty results instead of throwing. Added a regression test for the aggregation path.

### DIFF
--- a/modules/infra/src/main/scala/scaladex/infra/ElasticsearchEngine.scala
+++ b/modules/infra/src/main/scala/scaladex/infra/ElasticsearchEngine.scala
@@ -127,7 +127,7 @@ class ElasticsearchEngine(esClient: ElasticClient, index: String)(using Executio
   override def count(): Future[Int] =
     val query = must(matchAllQuery())
     val request = search(index).query(query).size(0)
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM

PR #1770, opened 2026-08-19
perf: bound the reverse-dependencies query on the artifact page

tbekas:
For artifacts with many reverse dependencies (e.g. `scala-library`), the reverse-dependency query took a long time to compute and returned a large number of rows — while the artifact page only ever displays the first 100 and discards the rest.

This bounds the query:

- `getReverseDependencies` now takes `limit`/`offset` (matching `getGroupIds`/`getArtifactRefs`), hardcoded to the first 100 in `ProjectPages`.
- The page query dedups (`DISTINCT ON`) and applies `LIMIT`/`OFFSET` before joining `artifacts`, so the join runs only for the kept rows.
- A separate `COUNT(*)` (index-backed, no join) provides the header total.
- Results are re-sorted in-app by `(scope, groupIdAndName)`.

### DIFF
--- a/modules/infra/src/test/scala/scaladex/infra/SqlDatabaseTests.scala
+++ b/modules/infra/src/test/scala/scaladex/infra/SqlDatabaseTests.scala
@@ -135,10 +135,25 @@ class SqlDatabaseTests extends AsyncFunSpec with BaseDatabaseSuite with Matchers
     for
       _ <- database.insertArtifacts(Seq(Cats.`core_3:2.6.1`, Cats.`kernel_3:2.6.1`))
       _ <- database.insertDependencies(Cats.dependencies)
-      reverseDependencies <- database.getReverseDependencies(Cats.`kernel_3:2.6.1`)
+      reverseDependencies <- database.getReverseDependencies(Cats.`kernel_3:2.6.1`, limit = 100, offset = 0)
     yield
       val sources = reverseDependencies.map(_.source)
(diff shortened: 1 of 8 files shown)


modules/infra/src/test/scala/scaladex/infra/SqlDatabaseTests.scala
```diff
@@ -135,12 +135,27 @@ class SqlDatabaseTests extends AsyncFunSpec with BaseDatabaseSuite with Matchers
...
+      ArtifactDependency(Cats.`core_3:4`.reference, Cats.`laws_3:2.6.1`.reference, Scope("compile"))
+    )
+    for
+      _ <- database.insertArtifacts(Seq(Cats.`core_3:2.6.1`, Cats.`core_3:4`, Cats.`laws_3:2.6.1`))
+      _ <- database.insertDependencies(deps)
+      reverseDependencies <- database.getReverseDependencies(Cats.`laws_3:2.6.1`, limit = 100, offset = 0)
+      count <- database.countReverseDependencies(Cats.`laws_3:2.6.1`)
+    yield
+      reverseDependencies should have size 1
+      count shouldBe 2L
```
tbekas, line 156: This is replication of the current behavior. The returned list is of size 1, but the returned count is 2.

In subsequent PRs I will make it so that both list and count will be consistent.

tgodzik (approved):
LGTM!

PR #1772, opened 2026-08-20
perf: DB statement timeout, request timeout and connection-pool tuning

tbekas:
Hardens database connection management against pool starvation, where slow queries under heavy (crawler) traffic held connections long enough to exhaust the pool.

- **Configurable DB statement timeout** (`scaladex.database.statement-timeout`, default `10 seconds`): applied to the web connection pool via Hikari `connectionInitSql`, so a runaway query is cancelled by Postgres and its connection returned instead of being held for the full request. Set to `0` (disabled) on the scheduler pool, since batch jobs may run legitimately long queries.
- **Configurable HTTP request timeout** (`scaladex.server.request-timeout`, default `20 seconds`): applied with `withRequestTimeout`.
- **Scheduler pool is timeout-free and smaller** (size 10): batch jobs and Flyway migrations run on it without the statement timeout, so a long-running migration (e.g. creating an index) is never cancelled.
- **Connect EC sized to the JDBC pool**: the doobie connect execution context now matches the connection-pool size instead of a hardcoded 32. Per the [doobie docs](https://typelevel.org/doobie/docs/14-Managing-Connections.html): *"The maximum thread limit for connectEC should be the same as your underlying JDBC connection pool, since any more threads are guaranteed to be blocked."*
- **Flyway extracted from `SqlDatabase`**: `SqlDatabase` is now purely a query layer; migrations/clean go through `DoobieUtils.flyway(...)`, with `clean` disabled in the prod environment via Flyway's `cleanDisabled`.

### DIFF
--- a/modules/data/src/main/scala/scaladex/data/Main.scala
+++ b/modules/data/src/main/scala/scaladex/data/Main.scala
@@ -54,14 +54,15 @@ object Main extends LazyLogging:
       DoobieUtils
         .transactor(datasource)
         .use { xa =>
-          val database = new SqlDatabase(datasource, xa, config.caching)
+          val database = new SqlDatabase(xa, config.caching)
           IO.fromFuture(IO(f(database)))
         }
         .unsafeRunSync()
     end usingDatabase
 
     def init(): Unit =
-      usingDatabase(database => Init.run(database, localStorage))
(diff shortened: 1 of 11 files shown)


tgodzik (approved):
LGTM

PR #1773, opened 2026-08-20
perf: add database indexes for artifact query hotspots

tbekas:
Adds three indexes (migration V28):

- **`artifact_latest_version_index`** — partial index on `artifacts (organization, repository) WHERE is_latest_version`. Speeds up the project/artifact header query, which previously scanned every artifact of large projects (e.g. `zio-aws`) just to filter out non-latest rows.
- **`artifact-dependencies-index`** — on `artifact_dependencies (target_group_id, target_artifact_id, target_version)`, for reverse-dependency lookups. Created manually on prod, reconciled here so fresh databases match.
- **`artifact_index_4`** — on `artifacts (organization, repository, group_id, artifact_id, release_date)`, for project + Maven-coordinate lookups. Created manually on prod, reconciled here so fresh databases match.

Plain `CREATE INDEX` (no `CONCURRENTLY`): migrations run before the server binds and before jobs start, so there are no writers to block.

### DIFF
--- a/modules/infra/src/main/resources/migrations/V28__add_is_latest_version_index.sql
+++ b/modules/infra/src/main/resources/migrations/V28__add_is_latest_version_index.sql
@@ -1,3 +1,0 @@
-CREATE INDEX IF NOT EXISTS artifact_latest_version_index
-  ON artifacts (organization, repository)
-  WHERE is_latest_version;
--- a/modules/infra/src/main/resources/migrations/V28__add_is_latest_version_index.sql
+++ b/modules/infra/src/main/resources/migrations/V28__add_is_latest_version_index.sql
@@ -1,0 +1,3 @@
+CREATE INDEX IF NOT EXISTS artifact_latest_version_index
+  ON artifacts (organization, repository)
+  WHERE is_latest_version;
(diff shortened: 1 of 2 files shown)


modules/infra/src/main/resources/migrations/V28__add_is_latest_version_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX IF NOT EXISTS artifact_latest_version_index
```
tgodzik: I don't think we can create indexes with the default user.

tbekas: If that is the case, we will need to address this issue - either by elevating the user's permissions or by running migrations as another user who has the necessary permissions. 

I checked production, and the database user used by the server owns all tables, which indicates that they should be able to create an index.

tbekas: I changed the scope of this PR to include also indexes that are already present in the production db. Once it's possible to create indexes again I think we should keep them in sync with migrations.

tgodzik (approved):
LGTM!

PR #1774, opened 2026-08-24
improvement: Log when we cannot find directories on maven

tgodzik:


modules/infra/src/main/scala/scaladex/infra/MavenCentralClientImpl.scala
```diff
@@ -99,6 +94,18 @@ class MavenCentralClientImpl()(using system: ActorSystem)
...
         Future.successful(None)
 
+  private def listDirectories(uri: String, status: model.StatusCode, page: String): Seq[String] =
+    if status != StatusCodes.OK then
+      logger.warn(s"Cannot list $uri: $status")
+      Seq.empty
+    else
+      val directories = JsoupUtils.listDirectories(uri, page)
+      if directories.isEmpty then logger.warn(s"No directories parsed from $uri (HTTP $status): ${preview(page)}")
+      directories
```
tbekas: Tiny suggestion - we can skip materializing response body into a String when response is not ok.
```suggestion
  private def listDirectories(uri: String, response: HttpResponse) =
    if response.status != StatusCodes.OK
    then
      logger.warn(s"Cannot list $uri: ${response.status}")
      response.discardEntityBytes()
      Future.successful(Seq.empty)
    else
      Unmarshaller
        .stringUnmarshaller(response.entity)
        .map(page =>
          val directories = JsoupUtils.listDirectories(uri, page)
          if directories.isEmpty then
            logger.warn(s"No directories parsed from $uri (HTTP ${response.status}): ${preview(page)}")
          directories
        )
  end listDirectories
```

modules/infra/src/main/scala/scaladex/infra/MavenCentralClientImpl.scala
```diff
@@ -51,23 +51,18 @@ class MavenCentralClientImpl()(using system: ActorSystem)
...
 
     for
-      responseFuture <- queueRequest(request)
-      page <- responseFuture.entity.dataBytes.runFold(ByteString(""))(_ ++ _).map(_.utf8String)
-    yield
-      val artifactIds = JsoupUtils.listDirectories(uri, page)
-      artifactIds.map(Artifact.ArtifactId.apply)
+      response <- queueRequest(request)
+      page <- response.entity.dataBytes.runFold(ByteString(""))(_ ++ _).map(_.utf8String)
+    yield listDirectories(uri, response.status, page).map(Artifact.ArtifactId.apply)
```
tbekas:
```suggestion
      dirs <- listDirectories(uri, response)
    yield dirs.map(Artifact.ArtifactId.apply)
```

modules/infra/src/main/scala/scaladex/infra/MavenCentralClientImpl.scala
```diff
@@ -51,23 +51,18 @@ class MavenCentralClientImpl()(using system: ActorSystem)
...
-
     val future = for
-      responseFuture <- queueRequest(request)
-      page <- Unmarshaller.stringUnmarshaller(responseFuture.entity)
-      versions = JsoupUtils.listDirectories(uri, page)
-      versionParsed = versions.map(Version.apply)
-    yield versionParsed
+      response <- queueRequest(request)
+      page <- Unmarshaller.stringUnmarshaller(response.entity)
+    yield listDirectories(uri, response.status, page).map(Version.apply)
```
tbekas:
```suggestion
      dirs <- listDirectories(uri, response)
    yield dirs.map(Version.apply)
```

tbekas (changes requested):
A small change requested

tbekas (approved):
LGTM

PR #1775, opened 2026-08-25
perf: shed excess web DB queries instead of queuing unboundedly

tbekas:
Under load, requests piled up in doobie's unbounded connect-EC queue waiting for a web-pool connection (handler times up to ~763s). `statement_timeout` only caps a *running* query and `withRequestTimeout` doesn't cancel the handler, so neither bounds that wait.

This bounds concurrent queries on the **web** `SqlDatabase` with a `Semaphore`: excess `run` calls fail fast with `DatabaseOverloadedException`, mapped to **503**, so the connect-EC queue can't grow. Configurable via `scaladex.database.max-concurrent-queries` (default 50, kept a bit above `pool-size`); the scheduler database stays unbounded.

### DIFF
--- a/modules/infra/src/main/resources/reference.conf
+++ b/modules/infra/src/main/resources/reference.conf
@@ -23,6 +23,7 @@ scaladex {
     port = 5432
     pool-size = 25
     statement-timeout = 10 seconds
+    max-concurrent-queries = 50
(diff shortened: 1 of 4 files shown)


tbekas:
This change may not be needed if we have cancellation in place #1776

tgodzik (approved):
LGTM

PR #1776, opened 2026-08-25
improvement: Switch to IO instead of Future

tgodzik:
Mostly automated, but doesn't look too bad.

### DIFF
--- a/modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
+++ b/modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
@@ -3,8 +3,6 @@ package scaladex.infra
 import java.time.Instant
 import java.util.UUID
 
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
 import scala.concurrent.duration.*
 
 import scaladex.core.model.*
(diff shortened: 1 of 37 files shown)


bishabosha (commented):
so many conversions from future to IO, shouldn't those underlying operations be converted also?

tgodzik:
> so many conversions from future to IO, shouldn't those underlying operations be converted also?

yeah, probably. I still need to go through all of it an check.

tgodzik:
I looked through the toIO and it's mostly us using APIs that are based on futures. I don't think we need to do it all together in one PR, the biggest problem we had is that we didn't have any cancelling on the database and it seems to easily get overwhelmed with requests.

modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
```diff
@@ -13,78 +11,64 @@ import scaladex.infra.config.CacheConfig
...
 import doobie.implicits.*
 
 class SqlDatabase(xa: doobie.Transactor[IO], cacheConfig: CacheConfig) extends SchedulerDatabase with LazyLogging:
 
-  private def buildCache[K, V](loader: K => Future[V]): AsyncLoadingCache[K, V] =
+  // Sync values only: a miss runs `load` on the caller's IO so Doobie can cancel it.
+  // refreshAfterWrite is omitted because Caffeine's refresh is Future-based and uncancellable.
+  private def buildCache[K, V](): Cache[K, V] =
     Scaffeine()
-      .refreshAfterWrite(cacheConfig.refreshAfter)
```
tgodzik, line 25: This is an issue with this approach since Scaffeine works on Future, so we don't get the referash

modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
```diff
@@ -349,9 +353,9 @@ class SqlDatabase(xa: doobie.Transactor[IO], cacheConfig: CacheConfig) extends S
...
       invalidateProject(status.destination)
 
-  def updateGithubStatus(ref: Project.Reference, githubStatus: GithubStatus): Future[Unit] =
+  def updateGithubStatus(ref: Project.Reference, githubStatus: GithubStatus): IO[Unit] =
     run(ProjectTable.updateGithubStatus.run(githubStatus, ref)).map(_ => invalidateProject(ref))
 
-  private def run[A](v: doobie.ConnectionIO[A]): Future[A] =
-    v.transact(xa).unsafeToFuture()
+  private def run[A](v: doobie.ConnectionIO[A]): IO[A] =
+    v.transact(xa)
```
tgodzik, line 360: This is the biggest change, we no longer switch to Future, so we should be able to cancell.

modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
tbekas: It was quite neat feature of caffine, but I think it's more important to have proper cancellation propagation, so let's leave it like that.

tbekas:
I know that this is already a very big refactor but I think we should go all the way with IO - that means replacing pekko-http with http4s too.

modules/infra/src/main/scala/scaladex/infra/SqlDatabase.scala
tbekas: The problem I see with this change is that concurrent requests for the same key will do concurrent db queries. In current impl if there was many queries for given cache key, they would await for a single Future to complete.

PR #1780, opened 2026-08-27
fix: make InMemoryDatabase thread-safe for concurrent test inserts

tbekas:
Tests insert artifacts concurrently (e.g. `OldSearchApiTests` via `Future.traverse`), but `InMemoryDatabase` used non-thread-safe `mutable.Map`, causing intermittent lost writes and an `empty.max` failure. Backs the maps with `TrieMap` and guards the dependencies buffer.


tgodzik (approved):
LGTM

PR #1781, opened 2026-08-27
fix: count deduplicated artifact reverse dependencies

tbekas:
The "N Dependents" label used `COUNT(*)`, but the list below dedups by `(source_group_id, source_artifact_id)`. The label now counts distinct dependents so it matches the list.


tgodzik (approved):
LGTM

PR #1782, opened 2026-08-28
fix: throttle and retry HTTP requests to Maven Central

tbekas:
Adds a client-side rate limiter and bounded 429/503 backoff (honoring `Retry-After`) so the Maven Central crawl stops getting rate limited. Throttle + retry live in `CommonAkkaHttpClient`, driven by a shared `HttpClientConfig` (`http-client` blocks in `reference.conf`); both are optional and default-off, so incidental GitHub clients opt out.

### DIFF
--- a/modules/infra/src/main/resources/reference.conf
+++ b/modules/infra/src/main/resources/reference.conf
@@ -44,7 +44,7 @@ scaladex {
   maven-central {
     http-client {
       throttle {
-        requests = 5
+        requests = 1
         per = 1 second
       }
       retry {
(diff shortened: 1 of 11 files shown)


modules/infra/src/main/resources/reference.conf
```diff
@@ -29,6 +29,30 @@ scaladex {
...
+    http-client {
+      throttle {
+        requests = 5000
+        per = 1 hour
+      }
+      retry {
+        max-retries = 5
+        initial-delay = 1 second
+        max-delay = 60 seconds
+      }
```
tbekas, line 41: This is adding retry to the GitHub client too. My logic was - since we have the retry feature in the abstract http-client, we can use it for GitHub client too, especially since GitHub had some stability issues recently.

tgodzik (approved):
LGTM

PR #1785, opened 2026-09-02
discover new Maven Central namespaces and admin review

halotukozak:
Scaladex had no way to find group IDs newly published to Maven Central: the periodic jobs only re-scan group IDs already in the artifacts table, and the old Sonatype post-deploy webhook no longer fires for Central Portal publishers (new namespace registration moved off it from 2024-02-01; e.g. com.halotukozak never appeared).

Adds a discovery stage that reads the Maven Central nexus incremental index (`.index/*.N.gz`), keeps a cursor, filters for Scala binary versions, and auto-indexes each new namespace through the existing `MavenCentralService.syncOne` (no filtering, matching the old publish-webhook behaviour).

- Starts from chunk 0 (the chain goes back to 2012-06-15) rather than only "from now on", so the gap left by the OSSRH→Central Portal migration gets backfilled too.
- No per-run cap on chunks or groups synced — a run pulls everything currently available. Pacing is entirely the shared, cross-client throttle on `CommonAkkaHttpClient` (same instance now used by both Maven Central clients), so a large backlog drains at the throttled request rate instead of trickling in over many scheduled runs.
- `discovered_group_id` tracks dedup + sync-retry bookkeeping only — there's no admin review queue. It never actually gated indexing (a group auto-syncs in the same run it's discovered, before any human could act on it), so it was pure maintenance with no real control; removed per review feedback rather than trying to make it cheaper to operate.
- New job `discover-maven-namespaces`, 12h frequency.

### DIFF
--- a/doc/dev/maven-central-discovery.md
+++ b/doc/dev/maven-central-discovery.md
@@ -1,0 +1,560 @@
+# Discovering new Scala organisations on Maven Central
+
+Research notes — 2026-09-02. Context: artifacts under new namespaces (e.g.
+`com.halotukozak`) never appear in Scaladex even though they are live on Maven
+Central, are valid Scala 3 / Scala.js / Native artifacts, and would be useful for
+the community build regression net (scala/scala3#26958).
+
+This document describes **how discovery works today**, **why it misses new
+namespaces**, **what Maven Central actually offers as an API**, and a
+**recommended pipeline** to close the gap.
+
+---
+
+## 1. How Scaladex ingests artifacts today
+
+Every artifact that enters the index goes through exactly one function:
+
+`PublishProcess.publishPom` → `ArtifactConverter.convert` → `artifactService.insertArtifact`
+(`modules/server/src/main/scala/scaladex/server/service/PublishProcess.scala`).
+
+There are only four callers:
+
+| Caller | Trigger | Can discover a brand-new groupId? |
+|---|---|---|
+| `PublishApi` `PUT /publish` | external HTTP push (sbt plugin / CI / Sonatype) | **yes** |
+| `MavenCentralService.findMissing` | job `missing-maven-artifacts`, every 24h | no — iterates `database.getGroupIds()` |
+| `MavenCentralService.findNonStandard` | job `non-standard-artifacts`, every 2h | no — fixed list from `non-standard.json` |
+| `AdminService.findMissingArtifacts` / `syncOne` | manual admin action, groupId typed by hand | no (human supplies the groupId) |
+
+`database.getGroupIds()` is `SELECT DISTINCT group_id FROM artifacts`
+(`SqlDatabase.getGroupIds` → `ArtifactTable.selectGroupIds`). So the periodic
+Maven Central sync is a **closed loop**: it can only re-scan namespaces that are
+*already* in the index. It backfills missing versions/artifacts of known groups;
+it can never see a new group.
+
+### The Maven Central client is a directory scraper
+
+`MavenCentralClientImpl` (`modules/infra/.../MavenCentralClientImpl.scala`) has
+three operations, all against `https://repo1.maven.org/maven2`:
+
+- `getAllArtifactIds(groupId)` — GET `…/<group/as/path>/`, parse the Apache
+  autoindex HTML with `JsoupUtils.listDirectories`.
+- `getAllVersions(groupId, artifactId)` — same, one level deeper.
+- `getPomFile(ref)` — GET the `.pom`.
+
+There is **no** call to any search/enumeration API. Given a groupId it works
+fine — e.g. `https://repo1.maven.org/maven2/com/halotukozak/` returns
+`alpaca_3/`, `commons_3/`, `mcodec_sjs1_3/`, … right now. The missing piece is
+purely *knowing the groupId exists*.
+
+### Where new groups used to come from
+
+Historically Scaladex "receives poms automatically from Maven Central (Sonatype)"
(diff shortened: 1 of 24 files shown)


doc/dev/maven-central-discovery.md
```diff
@@ -0,0 +1,560 @@
...
+- `getPomFile(ref)` — GET the `.pom`.
+
+There is **no** call to any search/enumeration API. Given a groupId it works
+fine — e.g. `https://repo1.maven.org/maven2/com/halotukozak/` returns
+`alpaca_3/`, `commons_3/`, `mcodec_sjs1_3/`, … right now. The missing piece is
+purely *knowing the groupId exists*.
+
+### Where new groups used to come from
+
+Historically Scaladex "receives poms automatically from Maven Central (Sonatype)"
```
tgodzik: This is useful as a docs for this PR, but I think we can remove it

modules/template/src/main/twirl/scaladex/view/admin/admin.scala.html
```diff
@@ -130,8 +131,86 @@ <h5>@Task.republishArtifacts.name</h5>
...
+              </div>
+            </td>
+            <td><button type="submit" class="btn btn-primary">Submit</button></td>
+          </form>
+        </tr>
+      </tbody>
+    </table>
+    <h2>Discovered projects (review queue)</h2>
+    <p>
+      Group IDs found on Maven Central by the <code>discover-maven-namespaces</code> job and auto-indexed.
```
tgodzik: This might be quite large amount, no? Should we show it at all?

modules/server/src/main/scala/scaladex/server/route/AdminPage.scala
```diff
@@ -72,6 +73,23 @@ class AdminPage(env: Env, adminService: AdminService):
...
+            post {
+              path("tasks" / Task.rewindDiscoveryCursor.name) {
+                formField("chunks-back") { raw =>
+                  adminService.rewindDiscoveryCursor(raw.toIntOption.getOrElse(8), user)
+                  redirect(Uri("/admin"), StatusCodes.SeeOther)
+                }
+              }
+            } ~
+            post {
+              path("discovered" / Segment / "review") { rawGroupId =>
```
tgodzik: I wonder if we want a manual decision here though, I think it might be too much maintenance.

modules/server/src/main/scala/scaladex/server/route/AdminPage.scala
```diff
@@ -72,6 +73,23 @@ class AdminPage(env: Env, adminService: AdminService):
                 adminService.republishArtifacts(user)
                 redirect(Uri("/admin"), StatusCodes.SeeOther)
               }
+            } ~
+            post {
+              path("tasks" / Task.rewindDiscoveryCursor.name) {
+                formField("chunks-back") { raw =>
+                  adminService.rewindDiscoveryCursor(raw.toIntOption.getOrElse(8), user)
```
tgodzik, line 78: Where do we start cursorvise, do you think we could set the initial cursor to ~1 year ago ?

tgodzik (changes requested):

modules/server/src/main/scala/scaladex/server/route/AdminPage.scala
halotukozak: hmmm, I worked a bit on this and IMO we should set cursor at 0 here. We will have to skip (not re-sync!) all the already synced artefacts, but won't miss any non-synced ones. It will be an expensive operation at the beginning, but it has to be done only once. We won't reach maven 429 thanks to the throttling.

modules/server/src/main/scala/scaladex/server/route/AdminPage.scala
halotukozak: hmmm, initially I came up with an idea to make a soft allowance. Every library is shown by default unless an admin reviews and rejects it. But you are right, no one would do it. It's easier to just accept all the slop libraries

tgodzik: We would have remember to check those. Alternatively we could set up some notification system, but I don't think it's worth it.

tgodzik (approved):
LGTM Let's try it on the dev instance first,

tbekas:
It's a substantial change, I would say let's test it on dev first for a week or so to see if it works as expected.

tgodzik:
Looks like there is some issues, trying to fix in https://github.com/scalacenter/scaladex/pull/1829

brianf:
Can we take a step back and have a conversation about what you are trying to do? This is all new from the central pov. There may be some other ways to achieve what you need without driving up the load or bumping against rate limits or tos violations for scraping. Can we hop on a zoom to hash it out?

tgodzik:
> Can we take a step back and have a conversation about what you are trying to do? This is all new from the central pov. There may be some other ways to achieve what you need without driving up the load or bumping against rate limits or tos violations for scraping. Can we hop on a zoom to hash it out?

I was afraid this was lost at some point, I think the contact person we discussed it with at Sonatype doesn't work there anymore. We used to get a push request every time new artefact was published, but that is no longer being invoked I think, which is why I am trying to work around it. And it seems we can use the indexes at https://repo1.maven.org/maven2/.index to make it more reliable and less resource intensive. We've been also adding handling for 429 responses to make sure we don't send too many requests.

I should be available next week and tomorrow for sure in CET timezone (can join late if needed). I can be reached by my work email `tgodzik @ virtuslab.com`

PR #1786, opened 2026-09-02
extend REST API v1 with search, dependencies and project settings

halotukozak:
Let's make scaladex more friendly for our text-based friends.

Authenticated endpoints (HTTP Basic `token:<github-token>`):
- GET   /api/v1/users/me
- GET   /api/v1/projects/{org}/{repo}/settings
- PATCH /api/v1/projects/{org}/{repo}/settings

Public endpoints:
- GET /api/v1/search
- GET /api/v1/projects/{org}/{repo}/dependencies
- GET /api/v1/projects/{org}/{repo}/dependents

Auth uses endpoints4s BasicAuthentication with a Scaffeine-cached token -> UserState lookup.
Page size/number are clamped, and dependency lists are collapsed to one entry per project so pagination metadata stays consistent. Settings writes are shared with the web form through the new ProjectSettingsService.

Vibe-codec, then reviewed and adjusted.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,7 +74,7 @@ lazy val infra = project
       "com.github.pjfanning" %% "pekko-http-circe" % "3.9.2",
       ("io.get-coursier" %% "coursier" % V.coursier).cross(CrossVersion.for3Use2_13),
       ("io.get-coursier" %% "coursier-sbt-maven-repository" % V.coursier).cross(CrossVersion.for3Use2_13),
-      "com.github.blemale" %% "scaffeine" % "5.3.0",
+      "com.github.blemale" %% "scaffeine" % V.scaffeine,
       "org.tpolecat" %% "doobie-core" % V.doobie,
       "org.tpolecat" %% "doobie-h2" % V.doobie,
       "org.tpolecat" %% "doobie-postgres" % V.doobie,
@@ -164,7 +164,7 @@ lazy val server = project
       "org.apache.pekko" %% "pekko-http-testkit" % V.pekkoHttp % Test,
       "org.apache.pekko" %% "pekko-http-cors" % V.pekkoHttp,
       "com.softwaremill.pekko-http-session" %% "core" % "0.7.1",
-      "com.github.blemale" %% "scaffeine" % "5.3.0",
+      "com.github.blemale" %% "scaffeine" % V.scaffeine,
       "org.apache.pekko" %% "pekko-http" % V.pekkoHttp,
       "org.endpoints4s" %% "pekko-http-server" % "2.0.1",
       "org.webjars" % "bootstrap-sass" % "3.4.1",
@@ -257,4 +257,5 @@ lazy val V = new {
   val json4s = "4.1.1"
   val coursier = "2.1.24"
   val gatling = "3.15.1"
+  val scaffeine = "5.3.0"
 }
--- a/build.sbt
+++ b/build.sbt
@@ -164,6 +164,7 @@ lazy val server = project
       "org.apache.pekko" %% "pekko-http-testkit" % V.pekkoHttp % Test,
       "org.apache.pekko" %% "pekko-http-cors" % V.pekkoHttp,
       "com.softwaremill.pekko-http-session" %% "core" % "0.7.1",
+      "com.github.blemale" %% "scaffeine" % "5.3.0",
       "org.apache.pekko" %% "pekko-http" % V.pekkoHttp,
       "org.endpoints4s" %% "pekko-http-server" % "2.0.1",
       "org.webjars" % "bootstrap-sass" % "3.4.1",
--- a/modules/core/shared/src/main/scala/scaladex/core/api/Endpoints.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/api/Endpoints.scala
@@ -1,10 +1,18 @@
 package scaladex.core.api
 
 import scaladex.core.model.*
+import scaladex.core.model.search.Page
+import scaladex.core.model.search.PageParams
 
 import endpoints4s.Validated
+import endpoints4s.algebra.BasicAuthentication
+import endpoints4s.algebra.BasicAuthentication.Credentials
 
-trait Endpoints extends JsonSchemas with endpoints4s.algebra.Endpoints with endpoints4s.algebra.JsonEntitiesFromSchemas:
+trait Endpoints
+    extends JsonSchemas
+    with endpoints4s.algebra.Endpoints
+    with endpoints4s.algebra.JsonEntitiesFromSchemas
+    with BasicAuthentication:
 
   val v0 = None
   val v1: Some[String] = Some("v1")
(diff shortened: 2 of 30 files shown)


modules/core/shared/src/main/scala/scaladex/core/service/ProjectService.scala
```diff
@@ -58,6 +61,59 @@ class ProjectService(database: WebDatabase, searchEngine: SearchEngine)(using Ex
...
+      case None => Future.successful(None)
+      case Some(project) =>
+        val resolvedVersion = version match
+          case some: Some[Version] => Future.successful(some)
+          case None => getHeader(project).map(_.map(_.latestVersion))
+        resolvedVersion.flatMap {
+          case None => Future.successful(Some(Page.empty[ProjectDependency]))
+          case Some(v) =>
+            for dependencies <- database.getProjectDependencies(ref, v)
+            yield Some(paginate(ProjectDependency.collapseByProject(dependencies, _.target), page))
```
tbekas: Let's do pagination all the way to the db layer - so with `OFFSET` and `LIMIT` and separate query for `COUNT`. Currently we fetch all rows and the `paginate` method simply throws away any results outside of the page (waste of resources). 

Alternatively, you can create an issue for this and we'll pick it up later on.

modules/core/shared/src/main/scala/scaladex/core/api/Endpoints.scala
```diff
@@ -145,6 +176,53 @@ trait Endpoints extends JsonSchemas with endpoints4s.algebra.Endpoints with endp
...
+  val getProjectDependentsV1: Endpoint[(Project.Reference, PageParams), Option[Page[ProjectDependency]]] =
+    endpoint(
+      get(api(v1) / projectDependentsPath /? pageParams),
+      ok(jsonResponse[Page[ProjectDependency]]).orNotFound()
+    )
+
+  /** `GET /api/v1/users/me` — information about the user authenticated by the
+    * `Authorization: Basic token:<github-token>` header. Returns 401 if no credentials are provided, 403 if the token
+    * is invalid.
+    */
```
tbekas: Let's strip pretty much all code comments from this PR - unless they explain something non obvious.

modules/core/shared/src/main/scala/scaladex/core/model/search/PageParams.scala
```diff
@@ -1,3 +1,13 @@
...
 final case class PageParams(page: Int, size: Int)
+
+object PageParams:
+  val MaxSize: Int = 100
+
+  /** Builds [[PageParams]] with `page` forced to at least 1 and `size` clamped to `[1, MaxSize]`, so that callers
+    * cannot trigger negative/huge SQL `LIMIT`/`OFFSET` or nonsensical pagination metadata.
+    */
+  def bounded(page: Int, size: Int): PageParams =
+    PageParams(page.max(1), size.max(1).min(MaxSize))
```
tbekas: This is the first time pagination is used in public API (with the exception of `OldSearchApi` that use some pagination too). I think we should be careful about how we introduce it. 

The Scaladex service is currently under constant moderate-high load and we have to use caching aggressively to facilitate this load. Caching and paging together means we have to go for some tradeoffs, if we simply allow arbitrary page size in range 1-100 we risk at least 2 harms:
1. cache pollution - cache can be easily polluted with calls using 100 different values of page size;
2. db amplification - pretty much all requests except for page size=100 will be unique, so they will translate increased db load.

I would suggest to quantise the allow list for page size to 20/50/100. Any requests with parameters other than these values should result with http 400 (with appropriate error message).

@tgodzik wdyt?

modules/server/src/main/scala/scaladex/server/GithubAuthImpl.scala
```diff
@@ -70,10 +70,15 @@ private class GithubAuthImpl(clientId: String, clientSecret: String, redirectUri
     githubClient.getUserState().map {
       case GithubResponse.Ok(userState) => Some(userState)
       case GithubResponse.MovedPermanently(userState) => Some(userState)
-      case GithubResponse.Failed(errorCode, errorMessage) =>
-        logger.warn(s"Failed to get user state: $errorCode, $errorMessage")
+      // 401 means the token itself is invalid; any other failure (rate-limit, 5xx, network) is transient and must not be
+      // reported as "unauthenticated", otherwise callers would see a misleading 403.
+      case GithubResponse.Failed(401, errorMessage) =>
```
tbekas: Instead of `401` literal i think it would be better to use enum/const `StatusCodes.Unauthorized.intValue` from `org.apache.pekko.http.scaladsl.model.StatusCodes`

build.sbt
```diff
@@ -164,6 +164,7 @@ lazy val server = project
       "org.apache.pekko" %% "pekko-http-testkit" % V.pekkoHttp % Test,
       "org.apache.pekko" %% "pekko-http-cors" % V.pekkoHttp,
       "com.softwaremill.pekko-http-session" %% "core" % "0.7.1",
+      "com.github.blemale" %% "scaffeine" % "5.3.0",
```
tbekas: This is the second time scaffeine is defined as a dependency. Please move string `"5.3.0"` to a new constant `V.scaffeine`.

tbekas (changes requested):

modules/core/shared/src/main/scala/scaladex/core/model/search/PageParams.scala
halotukozak: I've set 20/50/100 allowlist

modules/server/src/main/scala/scaladex/server/route/ProjectPages.scala
```diff
@@ -292,7 +290,7 @@ class ProjectPages(
         header <- projectService.getHeader(project)
         directDependencies <-
           header
-            .map(h => database.getProjectDependencies(ref, h.latestVersion))
+            .map(h => database.getProjectDependencies(ref, h.latestVersion, limit = 1000, offset = 0))
```
tbekas: I think it's better to set limit to 100 instead of 1k and then add a separate call for the count. Then pass the count down to the template and use the count value instead of `dependencies.size` - it's also more friendly to the cache (because of the overlap with API paging).

tbekas (changes requested):
Good changes overall 👍  But one last thing needs attention.

modules/template/src/main/twirl/scaladex/view/project/project.scala.html
```diff
@@ -119,6 +120,9 @@ <h4>@Formats.plural(dependencies.size, "Dependency")</h4>
           </div>
         </li>
       }
+      @if(count > 100) {
+        <p>and @{count - 100} more</p>
+      }
```
tbekas, line 125: Good addition 👍

tbekas (approved):
LGTM

PR #1787, opened 2026-09-03
feat: add a configurable circuit breaker to the HTTP client

tbekas:
Adds an optional, config-driven circuit breaker to `CommonAkkaHttpClient` (alongside the existing retry) so the client stops hammering an upstream that is down, rate limiting us, or rejecting our credentials. It trips on a failed request or any failed response except an absent resource (404/410). Enabled for GitHub via the `http-client.circuit-breaker` block, where a rate-limit 403 also counts as a failure.


tgodzik (approved):
LGTM

PR #1790, opened 2026-09-04
fix: prevent empty.max crash on latest-by-scala-version badge

tbekas:
## Problem

The `latest-by-scala-version.svg` badge route crashes with `empty.max` when the `platform` query param parses to a valid platform that the artifact doesn't actually publish for.

The `platforms.isEmpty` guard only checks that the artifact has *some* platforms — it never validates that the user-supplied `platform` is one of them. So `header.artifacts(name, platform)` returns empty and `summaryOfLatestVersions` calls `.max` on an empty collection.

```
2026-09-04 01:30:27,583 ERROR scaladex.server.Server$:197  - Unhandled exception in http://index.scala-lang.org/localytics/sbt-s3/sbt-s3/latest-by-scala-version.svg?platform=sbt1.0
java.lang.UnsupportedOperationException: empty.max
        at scala.collection.IterableOnceOps.max(IterableOnce.scala:1150)
        at scala.collection.IterableOnceOps.max$(IterableOnce.scala:1147)
        at scala.collection.AbstractIterable.max(Iterable.scala:936)
        at scaladex.server.route.Badges$.scaladex$server$route$Badges$$$summaryOfLatestVersions(Badges.scala:128)
        at scaladex.server.route.Badges.$anonfun$10(Badges.scala:101)
```

## Fix

Only accept the parsed `platform` param if it's among the artifact's available platforms (`.filter(platforms.contains)`), otherwise fall back to `targetType` selection or `platforms.max` as before. This guarantees `header.artifacts(...)` is non-empty.

Added a regression test covering a valid-but-unpublished platform param.

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/route/Badges.scala
+++ b/modules/server/src/main/scala/scaladex/server/route/Badges.scala
@@ -95,6 +95,7 @@ class Badges(projectService: ProjectService)(using ExecutionContext):
             else
               val platform = platformParam
                 .flatMap(Platform.parse)
+                .filter(platforms.contains)
                 .orElse(targetTypeParam.flatMap(selectPlatformFromTargetType(_, platforms)))
                 .getOrElse(platforms.max)
               val artifacts = header.artifacts(artifactName, platform)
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM

PR #1791, opened 2026-09-04
fix: percent-encode organization segment in search pagination URI

tbekas:
Organization names containing spaces (e.g. `/Big%20Data%20-%20small%20machine`) triggered an unhandled `IllegalUriException` and a 500. The decoded path segment was interpolated raw into the pagination URI string, which the template then parsed as a `Uri`. Now the segment is built through `Uri.Path` so it's percent-encoded.

```
2026-09-04 01:55:24,872 ERROR scaladex.server.Server$:197  - Unhandled exception in http://index.scala-lang.org/Big%20Data%20-%20small%20machine
org.apache.pekko.http.scaladsl.model.IllegalUriException: Illegal URI reference: Invalid input ' ', expected pchar, '/', '?', '#' or 'EOI' (line 1, column 17): organization/Big Data - small machine
                ^
        at org.apache.pekko.http.scaladsl.model.IllegalUriException$.apply(ErrorInfo.scala:88)
        at org.apache.pekko.http.scaladsl.model.Uri$.apply(Uri.scala:264)
        at org.apache.pekko.http.scaladsl.model.Uri$.apply(Uri.scala:234)
        at scaladex.server.route.SearchPages.search$$anonfun$1$$anonfun$2$$anonfun$1$$anonfun$1$$anonfun$1(SearchPages.scala:72)
```

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/route/SearchPages.scala
+++ b/modules/server/src/main/scala/scaladex/server/route/SearchPages.scala
@@ -8,6 +8,7 @@ import scaladex.core.service.SearchEngine
 import scaladex.server.TwirlSupport.given
 import scaladex.view.search.html.searchresult
 
+import org.apache.pekko.http.scaladsl.model.Uri
 import org.apache.pekko.http.scaladsl.model.Uri.*
 import org.apache.pekko.http.scaladsl.server.*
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM

PR #1792, opened 2026-09-04
fix: return error badge instead of crashing for unknown artifact name

tbekas:
## Problem

Requesting `latest.svg` for an explicit artifact name that doesn't exist crashes the request handler. `ProjectHeader.getDefaultArtifact` filters artifacts by the requested name, gets an empty set, and throws `empty.maxBy`.

```
2026-09-04 02:14:15,658 ERROR scaladex.server.Server$:197  - Unhandled exception in http://index.scala-lang.org/polyvariant/better-tostring/better-tostring/latest.svg
java.lang.UnsupportedOperationException: empty.maxBy
        at scala.collection.IterableOnceOps.maxBy(IterableOnce.scala:1183)
        at scala.collection.AbstractIterable.maxBy(Iterable.scala:936)
        at scaladex.core.model.ProjectHeader.getDefaultArtifact(ProjectHeader.scala:70)
        at scaladex.core.model.ProjectHeader.getDefaultArtifact0(ProjectHeader.scala:50)
        at scaladex.server.route.Badges.getDefaultArtifact$$anonfun$1$$anonfun$1(Badges.scala:121)
```

## Fix

- `getDefaultArtifact0` returns `Option[Artifact]`, so the `latest.svg` route renders the existing "no published artifacts" badge instead of 500ing.
- `ProjectHeader` artifact resolution is made total (no `maxBy`/`get`): the non-emptiness invariant is discharged once in `apply`, and the default artifact is carried on the header.
- Added a regression test for `latest.svg` with an unknown artifact name.

### DIFF
--- a/modules/core/shared/src/main/scala/scaladex/core/model/ProjectHeader.scala
+++ b/modules/core/shared/src/main/scala/scaladex/core/model/ProjectHeader.scala
@@ -9,18 +9,72 @@ object ProjectHeader:
       defaultArtifactName: Option[Artifact.Name],
       preferStableVersion: Boolean
   ): Option[ProjectHeader] =
-    Option.when(artifacts.nonEmpty) {
-      new ProjectHeader(ref, artifacts, defaultArtifactName, preferStableVersion)
+    // A header only exists if it has a default artifact, which is the case iff artifacts is non-empty.
+    resolveDefaultArtifact(artifacts, defaultArtifactName, preferStableVersion, None, None, None)
+      .map(defaultArtifact =>
+        new ProjectHeader(ref, artifacts, defaultArtifactName, preferStableVersion, defaultArtifact)
+      )
+
+  private def resolveDefaultArtifactName(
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/scaladex/core/model/ProjectHeader.scala
```diff
@@ -9,18 +9,72 @@ object ProjectHeader:
       defaultArtifactName: Option[Artifact.Name],
       preferStableVersion: Boolean
   ): Option[ProjectHeader] =
-    Option.when(artifacts.nonEmpty) {
-      new ProjectHeader(ref, artifacts, defaultArtifactName, preferStableVersion)
+    // A header only exists if it has a default artifact, which is the case iff artifacts is non-empty.
```
tgodzik:
```suggestion
    // A header only exists if it has a default artifact, which is the case if artifacts is non-empty.
```

tgodzik (approved):

PR #1794, opened 2026-09-04
Badges page: let users pick which artifact's badges to view

### LINKED ISSUE #1525: [Bug] Can't view artifact-specific badge, it only shows badges for a single
### Current behavior

Open this project:
https://index.scala-lang.org/jetbrains/sbt-idea-plugin

Notice that it has multiple artifacts, including sbt-idea-plugin:
https://index.scala-lang.org/jetbrains/sbt-idea-plugin/artifacts

Open it:
https://index.scala-lang.org/jetbrains/sbt-idea-plugin/artifacts/sbt-idea-plugin
or even some specific version of it:
https://index.scala-lang.org/jetbrains/sbt-idea-plugin/artifacts/sbt-idea-plugin/4.0.3

This is how it looks like:
image

Now click on "Badges".
Notice that it only shows the badges for the first artifact from the list of artifacts ("sbt-declarative-core"):
image

### Expected Behavior

I should be able to somehow see badges for every separate artifact.
It could be e.g. a simple dropdown selector or something

### Extra comments

_No response_

### Search terms

_No response_

halotukozak:
Fixes #1525.
Projects with multiple artifacts always showed badges for a single default artifact with no way to see the others. Add an artifact selector, matching the one on the Versions page.

implemented by Claude, verified

### DIFF
--- a/modules/server/src/main/assets/css/partials/_project.scss
+++ b/modules/server/src/main/assets/css/partials/_project.scss
@@ -566,10 +566,14 @@
   }
   
   .project-badges {
+    .badges-header {
+      margin-bottom: 20px;
+    }
+
     .badge-section {
       overflow: hidden;
     }
-  
+
     img {
       margin-bottom: 24px;
     }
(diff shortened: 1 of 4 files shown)


tgodzik (changes requested):
Looks good! You only need to rebase and fix conflicts.

tgodzik (approved):

PR #1795, opened 2026-09-07
Don't re-fetch the readme client-side

### LINKED ISSUE #1684: [Bug] GitHub rate-limiting blocks README from displaying
### Current behavior

currently (Tuesday 10 March 2026 10:00 CET) when i load any project in ScalaDex, the readme appears and then suddenly disappears. At the same time there are some request errors showing in the console that hint at rate-limiting when trying to access the readme via Github REST API. (unsure if connected or coincidence)

Image

### Expected Behavior

I expect to see a readme if it is possible

### Extra comments

the fact that there was a readme and then it disappears in UI suggests a javascript bug rather than a fundamental flaw with architecture.

If rate limiting is truly preventing displaying the UI then is there a better way?
e.g. the error already explains that the request that fetches the README isnt authenticated.

Also as last measure perhaps dont use REST API and try to scrape the content via HTML view?

### Search terms

Rate limiting, README

halotukozak:
Fixes #1684

This PR makes READMEs eventually consistent, but an user won't  encounter GitHub rate limits

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -132,7 +132,8 @@ lazy val webclient = project
     scalacOptions += "-Wunused:imports",
     libraryDependencies ++= Seq(
       "com.lihaoyi" %%% "scalatags" % "0.13.1",
-      "org.endpoints4s" %%% "fetch-client" % "4.0.1"
+      "org.endpoints4s" %%% "fetch-client" % "4.0.1",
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM!

PR #1797, opened 2026-09-07
fix O(n) non-standard-lib lookup

halotukozak:
lil performance optimisation

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/service/ArtifactConverter.scala
+++ b/modules/server/src/main/scala/scaladex/server/service/ArtifactConverter.scala
@@ -15,7 +15,8 @@ import com.typesafe.scalalogging.LazyLogging
 private case class ArtifactMeta(artifactId: Artifact.ArtifactId, isNonStandard: Boolean)
 
 class ArtifactConverter(paths: DataPaths) extends LazyLogging:
-  private val nonStandardLibs = NonStandardLib.load(paths)
+  private val nonStandardLookupByCoordinate: Map[(String, String), BinaryVersionLookup] =
+    NonStandardLib.load(paths).map(lib => (lib.groupId, lib.artifactId) -> lib.lookup).toMap
 
   def convert(
       pom: ArtifactModel,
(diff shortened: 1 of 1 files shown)


modules/server/src/main/scala/scaladex/server/service/GithubUpdater.scala
```diff
@@ -26,14 +30,15 @@ class GithubUpdater(database: WebDatabase, github: GithubClient)(using Execution
           .map(_._1)
 
       logger.info(s"Updating github info of ${projectToUpdate.size} projects")
-      projectToUpdate.mapSync(update).map { statuses =>
+      Source(projectToUpdate).mapAsync(GithubUpdater.parallelism)(update).runWith(Sink.seq).map { statuses =>
```
tbekas: If the goal of this change is to speed up this job, it's not going to happen. We're limited not by local processing but by github API limits. We actually throttle local processing to not exceed GH limits.

modules/infra/src/main/scala/scaladex/infra/GithubClientImpl.scala
```diff
@@ -83,13 +83,19 @@ class GithubClientImpl(token: Secret, config: HttpClientConfig = HttpClientConfi
...
-      contributors <- getContributors(repo.ref)
-      openIssues <- getOpenIssues(repo.ref)
-      scalaPercentage <- getPercentageOfLanguage(repo.ref, language = "Scala")
-      commitActivity <- getCommitActivity(repo.ref)
+      readme <- readmeF
+      communityProfile <- communityProfileF
+      contributors <- contributorsF
+      openIssues <- openIssuesF
+      scalaPercentage <- scalaPercentageF
+      commitActivity <- commitActivityF
```
tbekas: Unnecessary.

tbekas (changes requested):

halotukozak:
@tbekas you're right. I've missed the throttle queue and thought it will bring us closed to the limit

tbekas (approved):
LGTM

PR #1798, opened 2026-09-07
Host scaladocs locally, fetched from Maven Central

halotukozak:
Scaladex currently links out to the artifact's own Scaladoc (javadoc.io by default, or a custom pattern from project settings). This PR adds first-class scaladoc hosting: we fetch the artifact's already-published `-javadoc.jar` classifier directly from Maven Central, unpack it once, and serve it ourselves through a thin Scaladex-branded shell.

### DIFF
--- a/modules/server/src/main/scala/scaladex/server/service/ScaladocService.scala
+++ b/modules/server/src/main/scala/scaladex/server/service/ScaladocService.scala
@@ -1,0 +1,103 @@
+package scaladex.server.service
+
+import java.io.ByteArrayInputStream
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.StandardCopyOption
+import java.util.zip.ZipInputStream
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.jdk.CollectionConverters.*
+import scala.util.Using
+import scala.util.control.NonFatal
+
+import scaladex.core.model.Artifact
+import scaladex.core.service.MavenCentralClient
+
+import com.typesafe.scalalogging.LazyLogging
+
+/** Serves a project's scaladoc by lazily downloading and unpacking its `-javadoc.jar` from Maven Central on first
+  * request, then serving the unpacked files from disk on every subsequent request. There is deliberately no eviction:
+  * this is meant to run on a single server with disk to spare, and we'd rather learn real usage before adding one.
+  */
+class ScaladocService(cacheDir: Path, mavenCentralClient: MavenCentralClient)(using ExecutionContext)
+    extends LazyLogging:
+  // Only dedupes concurrent requests for the same not-yet-cached artifact; cleared once resolved either way, so a
+  // transient failure (Maven Central hiccup) doesn't get baked in - the next request just tries again.
+  private val inFlight = TrieMap.empty[Artifact.Reference, Future[Boolean]]
+
+  def localDir(ref: Artifact.Reference): Path =
+    cacheDir
+      .resolve(ref.groupId.value.replace('.', '/'))
+      .resolve(ref.artifactId.value)
+      .resolve(ref.version.value)
+
+  /** @return
+    *   true if the scaladoc for this artifact is available on disk (either already was, or was just downloaded and
+    *   unpacked), false if Maven Central has no javadoc jar for it.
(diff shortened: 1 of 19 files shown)


modules/server/src/main/scala/scaladex/server/service/ScaladocService.scala
```diff
@@ -0,0 +1,104 @@
...
+import scala.util.Using
+import scala.util.control.NonFatal
+
+import scaladex.core.model.Artifact
+import scaladex.core.service.MavenCentralClient
+
+import com.typesafe.scalalogging.LazyLogging
+
+/** Serves a project's scaladoc by lazily downloading and unpacking its `-javadoc.jar` from Maven Central on first
+  * request, then serving the unpacked files from disk on every subsequent request. There is deliberately no eviction:
```
tbekas: >There is deliberately no eviction

The service is used constantly by many crawlers and it's just a matter of time before we're going to download and unpack all possible javadocs for all artifacts in all versions.

I think we just need an eviction policy from the first iteration.

Edit: I just did a small research. Consider backing up the `inFlight` cache with scaffine cache, then there's `removalListener` that could remove files from disk. We would get cache like behavior (with limits and eviction policy) for free.

AI suggested that something like this could work
```
Scaffeine()
  .maximumWeight(maxBytes)
  .weigher((_: Artifact.Reference, e: Entry) => e.sizeOnDisk)
  .removalListener((ref: Artifact.Reference, _, cause) =>
    if cause != RemovalCause.REPLACED then deleteRecursively(localDir(ref)) // first move atomically -> then remove
  )
  .build[Artifact.Reference, Entry]()
```

modules/server/src/main/scala/scaladex/server/service/ScaladocService.scala
```diff
@@ -0,0 +1,104 @@
...
+        catch
+          case NonFatal(exception) =>
+            logger.warn(s"Failed to unpack javadoc jar of $ref", exception)
+            false
+    }
+
+  /** Unpacks into a sibling temp directory, then atomically moves it into place - so a concurrent request never sees a
+    * partially-unpacked directory.
+    */
+  private def unpack(bytes: Array[Byte], destination: Path): Unit =
```
tbekas, line 85: I think we should guard against zip-bomb vulnerability. Small zip file that unpack into very large files.

modules/server/src/main/scala/scaladex/server/service/ScaladocService.scala
```diff
@@ -0,0 +1,104 @@
...
+          }
+      }
+      Files.createDirectories(destination.getParent)
+      Files.move(staging, destination, StandardCopyOption.ATOMIC_MOVE)
+    finally deleteRecursively(staging)
+    end try
+  end unpack
+
+  // no-op once `unpack` has already moved `staging` away; only cleans up on failure
+  private def deleteRecursively(dir: Path): Unit =
```
tbekas: ~~There's no concurrency control. We could be downloading and deleting files at the same time in the same directory. I would suggest first move directory atomically, then delete files from moved directory recursively.~~

Edit: I just realized that there's no need for concurrency control, since the directory name is going to be likely unique.

tbekas (changes requested):
Overall it's a nice change 👍  I left some comments.

tbekas (approved):
Thanks for addressing the comments. LGTM

halotukozak:
> Thanks for addressing the comments. LGTM

note that numbers in scaladoc config are made up,  so they may require adjustment

scripts/PublishMissingArtifacts.scala
```diff
@@ -1,7 +1,3 @@
-//> using scala 3.3.8
-//> using dep org.jsoup:jsoup:1.22.2
-//> using dep org.scala-lang.modules:scala-xml_3:2.2.0
```
tgodzik, line 3: Why remove these? IT's only a separate script and will not work without the deps.

tgodzik (changes requested):
Looks good aside from one unexpected change

scripts/PublishMissingArtifacts.scala
halotukozak: no idea, it wasn't even AI xd

tgodzik (approved):
LGTM!

PR #1799, opened 2026-09-07
Fix artifact install-snippet tabs to match the site's tab styling

halotukozak:
## Summary

On an artifact page (e.g. `/cquiroz/scala-java-time/artifacts/scala-java-time/2.5.0-M2`), the install-snippet tabs (sbt/mill/gradle/...) used a different, inconsistent visual style compared to the rest of the site's tabs (Project/Artifacts/Versions in the header nav):

- `nav-justified` stretched the (usually 2-3) tabs to fill the full width, unlike anywhere else tabs are used.
- The active-state background was applied to the `<li>` instead of the `<a>` it contains, fighting the site's global flat `nav-tabs` styling.

This PR drops `nav-justified` and restyles `.install .nav-tabs` to match the site's flat tab look: transparent/neutral text for inactive tabs, brand-colored text for the active one, with a white background on the active tab so it still visually merges into the content panel directly below it (unlike the header nav, these tabs sit right above a bordered `.box`).

### DIFF
--- a/modules/server/src/main/assets/css/partials/_project.scss
+++ b/modules/server/src/main/assets/css/partials/_project.scss
@@ -8,6 +8,13 @@
     border-top-left-radius: 0;
     margin-bottom: 10px;
     margin-top: -40px;
+
+    // On the single-artifact page, .content-project is a bare wrapper around
+    // separate boxes rather than the box itself - the first one needs the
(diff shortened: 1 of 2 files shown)


halotukozak:
before: 
image

tbekas (approved):
LGTM

PR #1800, opened 2026-09-08
Add per-item tolerance to Maven and GitHub jobs

tbekas:
Wrap each item's future in the Maven and GitHub jobs with `recover(Resilience.tolerateHttpClientErrors(...))`: a non-fatal per-item failure is recovered to a fallback value so the batch continues, while an open circuit breaker (and fatal errors) propagates through `mapSync` and aborts the whole job. `tolerateHttpClientErrorsWith` is the variant that derives the fallback from the error.

Adds `PublishResult.Failed` to represent a tolerated publish failure in job summaries.

Note: HTTP clients still swallow the circuit-breaker exception via `fallbackTo`/`recoverWith`, so the abort-on-circuit-breaker branch is latent until a follow-up PR stops masking it.

### DIFF
--- a/modules/infra/src/main/scala/scaladex/infra/Resilience.scala
+++ b/modules/infra/src/main/scala/scaladex/infra/Resilience.scala
@@ -2,17 +2,27 @@ package scaladex.infra
 
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM

PR #1811, opened 2026-09-10
Drop bootstrap

halotukozak:
I've been trying to bump the Bootstrap version and then I realised that it would be much easier (and beneficial) to drop it entirely.


tgodzik (approved):
Looks good, I can't say much about css, but testing seems to not show any problems.

Why do you think it would be beneficial to drop bootstrap at this point?

halotukozak:
> Looks good, I can't say much about css, but testing seems to not show any problems.
> 
> Why do you think it would be beneficial to drop bootstrap at this point?

Migration to Bootstrap 4, and then to 5 is not trivial. I tried, and it requires many code changes, css adjustments etc. We already have many custom elements. Since we have to make so many changes to migrate, there's no point in sticking with Bootstrap. Vanilla JS is much faster, customizable and easy to maintain than the big, old library. In the age of AI, we are able to have nice grid and pretty buttons without the tech debt.

tgodzik:
Cool, let's merge it!

PR #1815, opened 2026-09-12
Stop hiding non-business errors in the GitHub client

tbekas:
Per-item tolerance is already in place at the job level (`Resilience.tolerateHttpClientErrors` in `GithubUpdater`), so the GitHub client shouldn't collapse every failure into a default value.

- Adds `GithubResponse.NotFound(code)` for "understood, but no data" status codes (404/410/204/202).
- Replaces the five `.fallbackTo(default)` calls with a `getOrDefault` helper: legit empty responses (no README, empty repo, issues disabled, stats warming up) map to the endpoint default; real errors (5xx, exhausted rate-limit, network, open circuit breaker) now fail the `Future` and reach the per-project tolerance instead of silently storing partial `GithubInfo`.
- `getRepository`'s 404 flows as `NotFound`, so `GithubUpdater` drops its magic `if code == 404` check.


tgodzik (approved):
LGTM

PR #1818, opened 2026-09-13
Share a rate-limited GitHub HTTP client, split into two lanes

tbekas:
## Problem

In dev the `user-sessions` job regularly failed with a GitHub **secondary** rate limit:

```
2026-09-12 01:21:39,819 ERROR s.server.service.JobScheduler:56  - Job user-sessions failed
java.lang.Exception: 403: {
  "message" : "You have exceeded a secondary rate limit. Please wait a few minutes before you try again. ...
  request ID 846A:66109:3388061:31F9005:6AA48D03."
}
        at scaladex.infra.GithubClientImpl.getRaw$$anonfun$1(GithubClientImpl.scala:382)
        ...
```

The root cause: `user-sessions` (and login/publish) created a **fresh `GithubClientImpl` per user/request**, each with the default config — no throttle, retry or circuit breaker. So there was no shared limiter across those requests.

Different user tokens don't help here: this is a **secondary** rate limit (anti-abuse/anti-scraping), not the per-token primary quota. It aggregates over what the requests have in common (source IP / OAuth app), so many GitHub calls fanned out across many user accounts from one host trip it regardless of token.

## Why two lanes

A single shared limiter would fix the rate limit but couple **latency-sensitive** interactive traffic (login, admin actions, publish) with the **bulk** scheduled jobs (`github-info`, `user-sessions`) — a batch run could starve a user trying to log in. So GitHub traffic is split into two lanes, each with its **own** throttle/queue/circuit breaker, sized to stay within GitHub's limits when summed:

- **scheduled** — cron jobs; low steady rate, small burst.
- **interactive** — login/admin/publish; low sustained rate but a larger burst so a single login (which fans out into many GraphQL calls) stays snappy. No circuit breaker (routine 401/404 responses would spuriously trip it).

## Changes

- `CommonAkkaHttpClient` is now a concrete, composable, host-agnostic (`superPool`) engine owning one throttle + circuit breaker; `GithubClientImpl` / `MavenCentralClientImpl` use it via composition instead of inheritance.
- GitHub token is passed per call, so one shared client serves all users.
- `user-sessions` tolerates per-item failures instead of aborting the whole job (an open circuit breaker still aborts).
- Optional `max-burst` added to the throttle config.


tgodzik (approved):
LGTM

PR #1821, opened 2026-09-17
Fix community profile decoding when files are null

tbekas:
GitHub's `community/profile` endpoint returns `null` for absent files (e.g. a repo without a `CONTRIBUTING` file). The `CommunityProfile` decoder drilled `downField` into these null nodes; circe treats `downField` on a non-object as a missing required field, failing the entire decode. As a result the whole community profile was discarded (and logged) for such repos:

```
ERROR scaladex.infra.Resilience$ - Recovered from an HTTP client error
DecodingFailure at .files.contributing.html_url: Missing required field
```

Fix: decode each file node as an optional object, so a `null` node yields `None` instead of failing. Adds unit tests for the null and present cases.


tgodzik (approved):
LGTM

PR #1824, opened 2026-09-17
Fetch Maven Central versions from maven-metadata.xml

tbekas:
`getAllVersions` scraped the artifact's HTML directory index to list versions. That per-artifact directory browsing is the bulk of our traffic to `repo1.maven.org` and is exactly the catalog-enumeration pattern Sonatype's abuse detection now blocks with `403 Forbidden`.

This fetches the artifact's `maven-metadata.xml` instead. The goal isn't fewer requests (it's still one per artifact) — it's a **different kind** of request:

- **Supported, spec'd interface** — `maven-metadata.xml` is part of the Maven repository format; HTML index browsing is an unsupported convenience Sonatype actively discourages.
- **Indistinguishable from normal build traffic** — every `mvn`/`sbt`/`gradle` fetches it constantly, so it doesn't read as scraping.
- **CDN-cacheable** — a static file served from Maven Central's edge, vs. a dynamically generated directory index.

`getAllArtifactIds` still lists the group directory: the search API can't reproduce cross-versioned sbt-plugin names (e.g. `sbt-scalafix_2.12_1.0`), but that's one request per group, not per artifact.

**Trade-off:** `maven-metadata.xml` can omit old versions that still exist on disk (e.g. `sbt-scalafix_2.12_1.0` `0.9.23`), so those are no longer indexed. Test updated to match.


tgodzik (approved):
LGTM

